PR #3, opened 2017-03-14
Remove auth token on pickle

### LINKED ISSUE #2: serialization
If gcsfs is to be used in the context of dask, it should not pass around the authorization token when pickled. Instead, we should require the user to have distributed a token file to each machine beforehand or otherwise set up permissions on each node, as we do for s3fs.

martindurant:
Assumed use: that token= is passed a file which is in the same
location for each worker.

Note that the class attribute `tokens` will be loaded for each
process on startup (unless they are threads, in which case we
don't care).

Fixes #2

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -150,10 +150,7 @@ class GCSFileSystem(object):
             raise ValueError('access must be one of {}', self.scopes)
         if project is None:
             warnings.warn('GCS project not set - cannot list or create buckets')
-        if token is not None:
-            if 'type' in token or isinstance(token, str):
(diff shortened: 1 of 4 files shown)


martindurant:
Question @mrocklin : pickle will not preserve the class attributes when pickling/unpickling an instance. Does cloudpickle, and is that which would be serializing in dask/distributed? The attributes `tokens` keeps a cache of recently used auths, so should not be passed around.

mrocklin:
I don't know what cloudpickle's behavior is here.

Generally when Dask serializes data it first tries pickle (this is faster) and then tries cloudpickle if something goes wrong.

PR #6, opened 2017-04-05
fix: multipage listings

vincentschut:
bucket lists give max. 1000 items per request (see https://cloud.google.com/storage/docs/json_api/v1/buckets/list), and uses a multipage approach to list >1000 items. This PR fixes gcsfs listing only the first 1000 items (and thus also denying any possible actions on items that are in the >1000 range).

Not sure how you'd like this to be in the unit tests. Either we'd need to create >1000 objects, or extend the ls/_list_bucket idiom to take kwargs, so we can adjust the max items per page returned. Then we could set that to e.g. 2 and test with only 3 objects.


martindurant:
Thank you for your contribution! It certainly looks good. 
Indeed, I would test by adding a parameter for page size, and setting this to be small in the test. I have not yet documented how to ass tests using VCR, are you comfortable doing this?

vincentschut:
No, I'm not. Could you give a short introduction (or point me to one)?

vincentschut:
Is it allowed to test '_list_bucket' instead of 'ls' for this? Then we don't have to pollute ls with a maxResults parameter (or with args, kwargs) just for testing purposes...

martindurant:
VCR records requests to the remote server, so that they can be replayed during tests - so long as the requests match exactly the original. It is set to strip out sensitive information before writing the request and responses into yaml files in the tests/recordings/ directory; the file-name matches the test name, so all tests must have unique names, across all test files.

The process is as follows: 
- set environment variables so that the tests run against your GCS credentials, and recording occurrs

    GCSFS_RECORD_MODE=once GCSFS_TEST_PROJECT='...' \\
    GCSFS_GOOGLE_TOKEN=/my_credentials.json py.test -vv -x -s gcsfs

- run this again, the RECORD_MODE=once should alert you if your tests are making different requests the second time around

- finally, test as TravisCI will, using only the recordings

    GCSFS_RECORD_MODE=none py.test -vv -x -s gcsfs

To reset recording and start again, delete the yaml file corresponding to the test.

vincentschut:
Cool! I'll give it a go.

Btw, are you aware of the fact that there are 2 'test_ls' functions in test_core.py? Shouldn't one of them be renamed?

martindurant:
I was *not* aware, but you are quite right :)

vincentschut:
followed your test instructions, but they don't work:

_vcr.errors.CannotOverwriteExistingCassetteException: No match for the request (<Request (GET) https://www.googleapis.com/storage/v1/b/gcsfs-testing/o/?maxResults=1000>) was found. Can't overwrite existing cassette ('/home/vincent/PycharmProjects/gcsfs_myfork/gcsfs/tests/recordings/test_simple_upload.yaml') in your current record mode ('once')._

That is of course because the request for all functions that use ls has changed internally, because _list_bucket now has a max_results parameter. Any hints how to proceed? I can of course just keep removing and regenerating yaml files until it succeeds. But maybe there is a smarter way?

martindurant:
Yes, I'm afraid you will have to delete all yaml files which used the ls call internally (which is probably most of them), since their recorded responses will no longer match. You can always delete *all* of the recordings, and then it will take longer to run the tests to create them again.
An easy workaround would be to not pass the maxResults parameter in the call in the case that it happens to be 1000 (or perhaps None).

vincentschut:
all tests pass now on my side, but I can't manage to push this without getting merge conflicts for all recordings. Any tips? I'm really quite a beginner with git...

martindurant:
What I would do is:
- switch to master
- pull
- switch to your branch
- merge master (reports conflicts)
- delete all conflicting yaml files (perhaps all)
- run tests again with the two pytest lines above.

It is a real shame that there is no moto for GCS...

martindurant:
Thank you very much, that did it.
It is a pity that the commits often end up large like this, but I don't think there's a better way.

vincentschut:
my pleasure!

Issue #12, opened 2017-09-21
Connection broken: error 104, ECONNRESET


jean-philippe-martin:
I used gcsfs to read about 800 files and it failed for 3 of them with this error:
```
Connection broken: error("(104, \'ECONNRESET\')",)
```

The call stack looks like this:
```
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 536, in open
    return GCSFile(self, path, mode, block_size)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 622, in __init__
    self.details = gcsfs.info(path)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 464, in info
    files = self.ls(path, True)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 379, in ls
    files = self._list_bucket(bucket)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 320, in _list_bucket
    out = self._call('get', 'b/{}/o/', bucket, maxResults=max_results, pageToken=next_page_token)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 299, in _call
    json=json)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 72, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 58, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 508, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 658, in send
    r.content
  File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 823, in content
    self._content = bytes().join(self.iter_content(CONTENT_CHUNK_SIZE)) or bytes()
  File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 748, in generate
    raise ChunkedEncodingError(e)
requests.exceptions.ChunkedEncodingError: ('Connection broken: error("(104, \'ECONNRESET\')",)', error("(104, 'ECONNRESET')",))
```

From the outside it looks like sometimes GCS has some sort of transient errors and GCSFS ought to be able to survive those and retry.

martindurant:
797/800 isn't bad :)
[...]

PR #15, opened 2017-09-25
Add retries to _call

jean-philippe-martin:
Also includes a helper program to open a file repeatedly (necessary,
along with patience, for the 503s to show up).

The retries print lots of log messages; I'm doing this so we can see
what's going on when it does in case there's a bug there. I'm open to
reducing the amount of logging if that's an issue.

Fixes issue #12

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -153,7 +153,7 @@ class GCSFileSystem(object):
         (see description of authentication methods, above)
     """
     scopes = {'read_only', 'read_write', 'full_control'}
-    retries = 10
+    retries = 4  # number of retries on http failure
     base = "https://www.googleapis.com/storage/v1/"
     _singleton = [None]
     default_block_size = 5 * 2**20
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -24,9 +24,10 @@ PY2 = sys.version_info.major == 2
 
 logger = logging.getLogger(__name__)
 
-not_secret = {"client_id": "586241054156-is96mugvl2prnj0ib5gsg1l3q9m9jp7p."
+# client created 23-Sept-2017
+not_secret = {"client_id": "586241054156-7a3vrghs70ffkkfkmnnatjnbjg03cq9a."
                            "apps.googleusercontent.com",
-              "client_secret": "_F-W4r2HzuuoPvi6ROeaUB6o"}
+              "client_secret": "RsSQZJKYE00oRv2ibYlj28Sx"}
 tfile = os.path.join(os.path.expanduser("~"), '.gcs_tokens')
 ACLs = {"authenticatedread", "bucketownerfullcontrol", "bucketownerread",
         "private", "projectprivate", "publicread"}
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -13,6 +13,7 @@ import os
 import pickle
 import re
 import requests
+import requests.exceptions
 import sys
 import time
 import warnings
@@ -23,10 +24,9 @@ PY2 = sys.version_info.major == 2
 
 logger = logging.getLogger(__name__)
 
-# client created 23-Sept-2017
-not_secret = {"client_id": "586241054156-7a3vrghs70ffkkfkmnnatjnbjg03cq9a."
+not_secret = {"client_id": "586241054156-is96mugvl2prnj0ib5gsg1l3q9m9jp7p."
                            "apps.googleusercontent.com",
-              "client_secret": "RsSQZJKYE00oRv2ibYlj28Sx"}
+              "client_secret": "_F-W4r2HzuuoPvi6ROeaUB6o"}
 tfile = os.path.join(os.path.expanduser("~"), '.gcs_tokens')
 ACLs = {"authenticatedread", "bucketownerfullcontrol", "bucketownerread",
         "private", "projectprivate", "publicread"}
@@ -34,6 +34,7 @@ bACLs = {"authenticatedRead", "private", "projectPrivate", "publicRead",
          "publicReadWrite"}
 DEFAULT_PROJECT = os.environ.get('GCSFS_DEFAULT_PROJECT', '')
 DEBUG = False
+RETRY_COUNT = 4
 
 if PY2:
     FileNotFoundError = IOError
@@ -109,6 +110,37 @@ def validate_response(r, path):
             raise ValueError("Bad Request: %s\n%s" % (path, msg))
         else:
             raise RuntimeError(m)
+
+def _isHtmlError(dictionary):
+    """True if it's a dictionary with keys 'errors', 'code', and 'message'."""
+    goal_keys=['errors', 'code', 'message']
+    return isinstance(dictionary, dict) and all(x in dictionary for x in goal_keys)
+
+def _asHtmlError(error):
+    """
+    Looks for an error that is dictionary with keys
+    "errors", "code", and "message".
+    If the input is one, or is an Error that contains one,
+    then return the dict. Otherwise, None.
+    """
+    if _isHtmlError(error):
+        warnings.warn('is not htmlerror: ' + error.__repr__())
+        return error
+    try:
+        inner = error.args[0]
+        if _isHtmlError(inner):
+            warnings.warn('args[0] is htmlerror: ' + inner.__repr__())
+            return inner
+        if isinstance(inner, dict):
+            inner = inner.get('error', None)
+            if _isHtmlError(inner):
+                warnings.warn('args[0]["error"] is htmlerror: ' + inner.__repr__())
+                return inner
+            warnings.warn('args[0]["error"] is not htmlerror: ' + inner.__repr__())
+        return None
+    except AttributeError:
+        warnings.warn('AttributeError on: ' + error.__repr__())
+        return None
 
 
 class GCSFileSystem(object):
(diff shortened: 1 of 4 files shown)


gcsfs/core.py
```diff
@@ -296,13 +329,47 @@ def _call(self, method, path, *args, **kwargs):
...
+                break
+            except RuntimeError as e:
+                warnings.warn("_call caught RuntimeError: " + e.__repr__())
+                with open('/tmp/error', 'wb') as f:
+                    pickle.dump(e, f)
+                print('error saved to /tmp/error')
+                if retry + 1 == RETRY_COUNT:
+                    warnings.warn('GCS call failure, out of retries: ' + str(e))
+                    raise e
+                html_error = _asHtmlError(e)
```
martindurant: It seems like a lot of code to get the HTTP code. Can you use `r.status_code`, which is always an integer?

gcsfs/core.py
```diff
@@ -296,13 +329,47 @@ def _call(self, method, path, *args, **kwargs):
...
+                         json=json)
+                validate_response(r, path)
+                break
+            except RuntimeError as e:
+                warnings.warn("_call caught RuntimeError: " + e.__repr__())
+                with open('/tmp/error', 'wb') as f:
+                    pickle.dump(e, f)
+                print('error saved to /tmp/error')
+                if retry + 1 == RETRY_COUNT:
+                    warnings.warn('GCS call failure, out of retries: ' + str(e))
```
martindurant: I would not use these warnings. DEBUG-level logging would be ok, or probably I would do without in the first instance.

gcsfs/core.py
```diff
@@ -296,13 +329,47 @@ def _call(self, method, path, *args, **kwargs):
...
+                    if html_error['code'] in [500, 503, 504, '500', '503', '504']:
+                        # 5xx error, let's retry
+                        warnings.warn('Retrying after GCS call failure: ' + str(e))
+                        continue
+                    else:
+                        warnings.warn('GCS call failure, HtmlError but "' + str(html_error.get('code', '')) + '" so not retrying: ' + str(e))
+                else:
+                    warnings.warn('GCS call failure, not an HtmlError so not retrying: ' + str(e))
+                raise e
+            except requests.exceptions.ChunkedEncodingError as e:
```
martindurant: Are there other requests exceptions that should be caught? Perhaps all of them?

gcsfs/core.py
```diff
@@ -296,13 +329,47 @@ def _call(self, method, path, *args, **kwargs):
...
+        for retry in range(RETRY_COUNT):
+            try:
+                time.sleep(2**retry-1)
+                r = meth(self.base + path, headers=self.header, params=kwargs,
+                         json=json)
+                validate_response(r, path)
+                break
+            except RuntimeError as e:
+                warnings.warn("_call caught RuntimeError: " + e.__repr__())
+                with open('/tmp/error', 'wb') as f:
```
martindurant: We definitely don't want to do this, even less so to an absolute path that might not exist or be writable. I would say log in the appropriate places (sparingly) and then the user can always attach an arbitrary log handler to do whatever they like with the messages.

gcsfs/core.py
```diff
@@ -296,13 +329,47 @@ def _call(self, method, path, *args, **kwargs):
...
-        r = meth(self.base + path, headers=self.header, params=kwargs,
-                 json=json)
+        for retry in range(RETRY_COUNT):
+            try:
+                time.sleep(2**retry-1)
+                r = meth(self.base + path, headers=self.header, params=kwargs,
+                         json=json)
+                validate_response(r, path)
+                break
+            except RuntimeError as e:
```
martindurant: Is there anything fundamentally different about the two fail blocks, or can they be combined?

gcsfs/core.py
```diff
@@ -296,13 +329,47 @@ def _call(self, method, path, *args, **kwargs):
...
+                html_error = _asHtmlError(e)
+                if html_error :
+                    if html_error['code'] in [500, 503, 504, '500', '503', '504']:
+                        # 5xx error, let's retry
+                        warnings.warn('Retrying after GCS call failure: ' + str(e))
+                        continue
+                    else:
+                        warnings.warn('GCS call failure, HtmlError but "' + str(html_error.get('code', '')) + '" so not retrying: ' + str(e))
+                else:
+                    warnings.warn('GCS call failure, not an HtmlError so not retrying: ' + str(e))
```
martindurant: What kind of situation causes this? Not-authorised, perhaps?

gcsfs/core.py
jean-philippe-martin: Oh yes sorry, that's debug code I forgot to remove!

gcsfs/core.py
jean-philippe-martin: In the case of the RuntimeError we have to extract the HTTP error code.  There isn't much code beyond that (especially once we remove the logging and debug code I forgot to remove).

gcsfs/core.py
jean-philippe-martin: OK, I removed them.

gcsfs/core.py
jean-philippe-martin: All I know is that we can't exclude the possibility. I suspect a bug in our code or in the Google code could throw a RuntimeError, or perhaps bad user input of some kind.

gcsfs/core.py
jean-philippe-martin: Retrying on a non-retriable error would be counter-productive. I went through the [list of requests.exception errors](http://docs.python-requests.org/en/master/_modules/requests/exceptions/) and added a few that sounded like a retry may work. I don't think we should retry on all, though, as most of the rest sounds pretty definitive.

jean-philippe-martin:
OK, the new code's better, shorter, and retries under more conditions. I also found and fixed a bug.

One thing we may consider at this point is to upgrade validate_response so it does the parsing ahead of time, into a new error type we'd introduce (perhaps "HtmlError").

gcsfs/tests/test_manyopens.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+"""
+Test helper to open the same file many times.
+"""
+from __future__ import print_function
+import sys
+import gcsfs
+
+def run():
```
martindurant: This does not look like a pytest test - you intention is that it should only be run on real GCS connections? It would not work with VCR recording anyway, which would presumably not return any errors (unless there happened to be errors when they were recorded).

gcsfs/core.py
```diff
@@ -296,13 +328,38 @@ def _call(self, method, path, *args, **kwargs):
...
+            except RuntimeError as e:
+                if retry + 1 == RETRY_COUNT:
+                    raise e
+                html_error = _asHtmlError(e)
+                if html_error :
+                    if html_error['code'] in [500, 503, 504, '500', '503', '504']:
+                        # 5xx error, let's retry
+                        continue
+                raise e
+            except (requests.exceptions.ChunkedEncodingError,
```
martindurant: perhaps call this retriable_exceptions and move to utils module.

gcsfs/core.py
martindurant: So runtime error is produced by requests?
If it's the one raised in validate_response (which I think it must be), then r already is defined and will always have the status_code attribute available - to avoid having to parse the contents.
If parsing is the only way to do this (hope not!), then it would be better if the parsing code were in the util module. It could also do with tests of its own.

gcsfs/tests/test_manyopens.py
jean-philippe-martin: This is not a pytest. It's for manual testing. I don't know where to put it, open to suggestions.

gcsfs/core.py
jean-philippe-martin: > So runtime error is produced by requests?
If it's the one raised in validate_response (which I think it must be), then r already is defined and will always have the status_code attribute available - to avoid having to parse the contents.

I *think* you're trying to tell me to use the parsed data that is visible to validate_response instead of trying to parse that data again after validate_response unhelpfully serializes it to a string. If so then I agree and I think that the best way to do that is to raise an appropriate exception that contains the info we need.

jean-philippe-martin:
This should address the review comments if I understood them correctly.

gcsfs/core.py
```diff
@@ -34,6 +38,7 @@
          "publicReadWrite"}
 DEFAULT_PROJECT = os.environ.get('GCSFS_DEFAULT_PROJECT', '')
 DEBUG = False
+RETRY_COUNT = 4
```
martindurant: One final request: could this be made a class attribute? I don't expect people to change it, but it seems to fit more obviously there.

martindurant:
#18 will create a new source of exception, but per file rather than per resquests operation; as with your experience, @jean-philippe-martin , we are seeing some connection errors, but also some incomplete files without error on rare occasions.

gcsfs/core.py
jean-philippe-martin: Done! Actually while doing that I noticed there's already a `retries`, so I switched to using that instead - it's easier to configure the class if there's a single number for retry attempts. I also switched a "+1" to "-1" to match the style of the other retry block, to keep things easier to maintain.

jean-philippe-martin:
Oh, wow, I hadn't seen any incomplete file but that's a lot worse than failures. Thanks for fixing that.

jean-philippe-martin:
I added a bit of documentation. The functions in utils seem too small and simple for tests. No worries about the speed, we all have other things we're doing. Enjoy the conference!

jean-philippe-martin:
@martindurant OK, ready to merge?

martindurant:
Thank you for your work, @jean-philippe-martin !
Now we need to add consistency checks too...

jean-philippe-martin:
Thank you, @martindurant ! I won't be able to contribute for a bit but I'm sure you can do the consistency checks!

PR #20, opened 2017-10-05
Add size and md5 consistency

### LINKED ISSUE #18: Check file consistency on write
GCS provides size and md5 hash information on written files. At the minimum, the former should be checked against the expected upload size, and optionally the md5 also for a stronger guarantee.

martindurant:
Fixes #18 

What should be the default option?

Any need for crc32 check as another option? (I suppose its a bit faster than md5)

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -5,6 +5,8 @@ Google Cloud Storage pythonic interface
 from __future__ import print_function
 
 import array
+from base64 import b64encode
+from hashlib import md5
 import io
 import json
 import logging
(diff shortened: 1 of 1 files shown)


martindurant:
Note that I cannot easily think of a way to induce a consistency failure: perhaps injecting spurious data into the md5 hash object?

Also note that, unlike failures on individual REST calls, a consistency failure is not retriable.

@mrocklin - this, together with retries, would hopefully have avoided, or at least exposed, problems with the massive data transfer you were running.

mrocklin:
I take it that google is calling md5 and crc32 and such on their end?

I don't have strong thoughts on what the default should be.  How fast is crc32?

martindurant:
Yes, google supplies md5 and crc32 on each file when the upload is complete.
A quick test in a console shows the two to be the same speed (~1GB/s for me), which is a little surprising.

mrocklin:
Doing a consistency check by default seems like the right choice to me

mrocklin:
FWIW I would go for something more intense than size.  I would expect google network to max out around 1-200MB/s, which, given your numbers, means 20% or so of a CPU core.  I think that that is a reasonable sacrifice for default behavior.

PR #22, opened 2017-10-11
Add simple info without ls

### LINKED ISSUE #21: Avoid downloading list of files in bucket when object path contains no wildcard
Original discussion [here](https://github.com/dask/dask/issues/2746)

Quick summary: When reading a single file from a bucket containing a very large number of files, the read time can be very long before gcsfs downloads the list of all the files in the bucket. For a bucket with over 4 million files, this can take up to 15 minutes. 

Suggested fix (easy): Check if the object path contains a wildcard. If it does not, then just download file. If it does, then proceed to download list of files. 

Suggested fix (hard, not sure how to do yet, suggested by @martindurant ): implement prefix/delimiter listing, i.e., make it possible to use wildcard in object path (.e.g., `bag.read_text('gs://mybucket/2017/01/*.csv')`) *without* having to download list of files from entire bucket.

martindurant:
Fixes #21 

@halfdanrump - does this do the job? It fails a couple of tests for reasons that are entirely beyond me.


halfdanrump:
@martindurant Yeah, it solves the issue, thank you!! I was getting ready for diving into the code myself, but thank you for fixing it so quickly! :))

martindurant:
I have not, however, managed to fix the tests!
The tests pass with real credentials, but for some reason, some of them do not make VCR recordings that play correctly without credentials.

martindurant:
I intend to implement ls with prefixes like in s3fs in the medium term.

martindurant:
Pushed in separate commit ef42a5360ca577ddacac312a1e44ad7520dba2c1

Issue #35, opened 2017-11-20
RuntimeError: Waited too long for browserauthentication.


rabernat:
I am trying to create a new GCSFileSystem with the command
```python
fs = gcsfs.GCSFileSystem(project='pangeo-181919', token=None)
```

This triggers my browser to open and I have to enter the code. But I can never seem to finish it fast enough before I get the following error:
```
---------------------------------------------------------------------------
ApplicationDefaultCredentialsError        Traceback (most recent call last)
~/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/gcsfs/core.py in connect(self, refresh)
    241             try:
--> 242                 data = self.get_default_gtoken()
    243                 self.tokens[(project, access)] = data

~/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/gcsfs/core.py in get_default_gtoken()
    303     def get_default_gtoken():
--> 304         au = oauth2client.client.GoogleCredentials.get_application_default()
    305         tok = {"client_id": au.client_id, "client_secret": au.client_secret,

~/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/oauth2client/client.py in get_application_default()
   1270         """
-> 1271         return GoogleCredentials._get_implicit_credentials()
   1272 

~/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/oauth2client/client.py in _get_implicit_credentials(cls)
   1260         # If no credentials, fail.
-> 1261         raise ApplicationDefaultCredentialsError(ADC_HELP_MSG)
   1262 

ApplicationDefaultCredentialsError: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.

During handling of the above exception, another exception occurred:

RuntimeError                              Traceback (most recent call last)
<ipython-input-2-f5bb2896043c> in <module>()
----> 1 fs = gcsfs.GCSFileSystem(project='pangeo-181919', token=None)

~/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/gcsfs/core.py in __init__(self, project, access, token, block_size, consistency)
    178         self.consistency = consistency
    179         self.dirs = {}
--> 180         self.connect()
    181         self._singleton[0] = self
    182 

[...]

Issue #31, opened 2017-11-14
does not handle 502 errors (returned by GCS when server is overloaded)


yiga2:
I have faced several times the same 502 error when writing in chunks. Errors are sporadic and look like:
```
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 896, in __exit__
    self.close()
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 868, in close
    self.flush(force=True)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 755, in flush
    self._upload_chunk(final=force)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 777, in _upload_chunk
    validate_response(r, self.location)
  File "/usr/local/lib/python2.7/dist-packages/gcsfs/core.py", line 111, in validate_response
    raise RuntimeError(m)
RuntimeError: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 502 (Server Error)!!1</title>
  <style>
    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}
[...]

Issue #28, opened 2017-11-13
glob through whole tree


yiga2:
Suggestion/request to have a glob option to traverse the full tree e.g.
`GCSFS.glob('bucket/**/*.[gz|zip]', recursive=true)` to list all matching files regardless of where located. This mimics glob in Python 3.6.

Workaround I can think of is to use walk() or ls() then regex

PR #37, opened 2017-11-21
Small fixes

martindurant:
- specify that "project" is what Google calls project_id
- longer wait minimum wait time fro browser auth
- retries for 501, 502 status
- no more writing after force flush
- only set force if write was successful
- add glob **


martindurant:
Fixes #35, #31, #28

PR #39, opened 2017-11-29
metadata on resumable creation

yiga2:


martindurant:
Please merge following #37 , which fixes the `forced` order and code 502.

yiga2:
My bad - my repo was not in sync.
Done.

martindurant:
OK, I can merge this, since it's very unobtrusive, but I would like to see a follow up PR with the other parts I mentioned, as well as tests.

yiga2:
Will do - I have a need for setting metadata on existing objects too. 
For now I use `gsutil setmeta` but interestingly it has a batch option to set in bulk. I will check its code to get inspired for the same here.
Thanks @martindurant, my first merge ! (am only a coder by trade)

PR #47, opened 2017-12-08
Print token url rather than open browser

mrocklin:
Previously we would open a browser for the user.  Unfortunately this was not
helpful when operating on remote servers.  Now we print a URL to the console
and let them navigate on their own.

Additionally this no longer times out if the user takes a while.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -18,7 +18,6 @@ import requests
 import sys
 import time
 import warnings
-import webbrowser
 
 from requests.exceptions import RequestException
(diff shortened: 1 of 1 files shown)


martindurant:
OK, so now we don't ever time out at all? That's OK with me, but perhaps the message should suggest ^C if the user wants to cancel.

mrocklin:
I think that using Ctrl-C is normal behavior at this point.  I would not expect any user to stick around waiting for twenty seconds in order to cancel.

PR #51, opened 2017-12-29
[WIP] Add fuse draft

mrocklin:
This is mostly here for conversation.  This is nowhere near working yet.

### DIFF
--- a/gcsfs/gcs_fuse.py
+++ b/gcsfs/gcs_fuse.py
@@ -1,0 +1,129 @@
+import errno
+import logging
+import stat
+import sys
+
+from distributed.utils import log_errors
+from fuse import FUSE, FuseOSError, Operations, LoggingMixIn
+import gcsfs
+import pandas as pd
+
+
+class GCS(LoggingMixIn, Operations):
+    def __init__(self, fs, bucket):
+        print('hello')
+        self.bucket = bucket.rstrip('/')
+        self.files = dict()
+        self.fd = 0
+        self.fs = fs
+
+    def _path(self, path):
+        print('path:', path)
+        return self.bucket + '/' + path.lstrip('/')
+
+    def create(self, path, mode):
+        file = self.fs.open(self._path(path), mode=mode)
+        fd = self.fd
+        self.fd += 1
+        self.files[fd] = file
+        return fd
+
+    def open(self, path, flags):
+        file = self.fs.open(self._path(path))
+        fd = self.fd
+        self.fd += 1
+        self.files[fd] = file
+        return fd
+
+    def read(self, path, length, offset, fh):
+        file = self.files[fh]
+        file.seek(offset)
+        data = file.read(length)
+        return data
+
+    def write(self, path, buf, offset, fh):
+        file = self.files[fh]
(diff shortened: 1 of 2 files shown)


mrocklin:
OK, some questions:

1.  What is the right way to implement `getattr`, a method that expects to get information about a path.  That path may point to either a file or a directory
2.  In particular, `GCSFileSystem.info` on a directory doesn't return what we want it to return
3.  Any thoughts on permissions?  Are there ways to see if something is publicly visible or not?  Or should we just default to `0o600` for all things?

gcsfs/gcs_fuse.py
```diff
@@ -0,0 +1,129 @@
...
+        print(paths)
+
+        for p in dirents + list(paths):
+            yield p
+
+    def rmdir(self, path):
+        return self.fs.rmdir(self._path(path))
+
+    def mkdir(self, path, mode):
+        return os.mkdir(self._full_path(path), mode)
```
martindurant, line 88: `os`->`self.fs`?
Will be no-op on none top-level directories.

gcsfs/tests/test_core.py
```diff
@@ -59,6 +59,13 @@ def test_info(token_restore):
         assert gcs.info(a) == gcs.ls(a, detail=True)[0]
 
 
+@my_vcr.use_cassette(match=['all'])
+def test_info_directory(token_restore):
```
martindurant, line 63: You need to set the various env variables in the test settings module and run with recording on (will produce a YAML file) before tests will pass.

mrocklin: I think that this won't work regardless

```python
In [1]: from gcsfs import GCSFileSystem

In [2]: gcs = GCSFileSystem()

In [3]: gcs.ls('pangeo-data')
Out[3]: 
['pangeo-data/newman-met-ensemble/',
 'pangeo-data/newmann-met-ensemble-netcdf/',
 'pangeo-data/test997/',
 'pangeo-data/test998/',
 'pangeo-data/test999/']

In [4]: gcs.info('pangeo-data/test997/')
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-4-210f02b52f31> in <module>()
----> 1 gcs.info('pangeo-data/test997/')

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/gcsfs/core.py in info(self, path)
    468             return out[0]
    469         else:
--> 470             raise FileNotFoundError(path)
    471 
    472     def url(self, path):

FileNotFoundError: pangeo-data/test997/
```

martindurant:
Probably to do this right, directory listings should be made using prefix/delimiter as happens in s3fs. Currently, the only way to get a list a "directories" is to list all files in a bucket, and do some text processing on the names; this is implemented when listing, and so it could be used for info too. That complicated things though, because there is an intent in #22 to use HEAD on known full paths rather than bucket listing, accounting for cases where a bucket contains a large number of keys.
In addition, all of this might conflict with the serialisation problem for the directories cache - the more information we keep, the less we need to do new listings (i.e., more api calls), but the bigger the file-system object gets.

Issue #50, opened 2017-12-22
Use requests.Session to avoid repeated SSL handshakes


mrocklin:
This can be nice when the other side is relatively fast

```python
In [1]: import requests
In [2]: %timeit requests.get('https://en.wikipedia.org')
10 loops, best of 3: 116 ms per loop
In [3]: s = requests.Session()
In [4]: %timeit s.get('https://en.wikipedia.org')
10 loops, best of 3: 48.5 ms per loop
```

It's not clear yet how effective this will be in practice.

PR #55, opened 2018-01-03
All HTTP calls through a session

martindurant:
One requests.Session per GCSFileSystem instance. Should cut down on SSL handshaking.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -228,6 +228,7 @@ class GCSFileSystem(object):
             Force refresh, even if the token is expired.
         """
         token = self.input_token
+        self.session = requests.Session()
         project, access = self.project, self.access
         if token is not None and str(token) != 'cloud':
             if 'type' in token or isinstance(token, str):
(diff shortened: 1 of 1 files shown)


martindurant:
Fixes #50

martindurant:
Q: would it be better to have a single global Session for everything? In typical dask usage, I think we may be creating new GCSFileSystem instances, which would have new sessions and not benefit from sharing.

mrocklin:
If the different GCSFileSystem objects had different permission would they want different Sessions?

Alternatively, perhaps we should have a singleton `GCSFileSystem.current()`?

martindurant:
No, we can still maintain a different header token - which fixes the user identity - for each file-system instance rather than have it set for the global session. The SSL connection should still be sharable between them, the interchange of certificates does not depend on the user token.

mrocklin:
I'm inclined to merge this for now and then, once we hit performance issues with many file systems resolve that later, perhaps by caching file system objects.

martindurant:
Agreed.

Issue #24, opened 2017-10-31
`ls` takes a long time with buckets with lots of files


bmabey:
I have a bucket with several directories, each of which has thousands of files. When querying one of the directories that has 3600 files a call to `fs.ls("bucket/my-dir")` takes several minutes. I actually don't know how long it takes since I've never let it finish. In contrast, the google client takes 3.4 seconds to return the directory listing `[f for f in bucket.list_blobs(prefix='my-dir')]` (the comprehension is to force iteration over the iterator object they return).

martindurant:
Indeed, gcsfs has not yet implemented file listing with prefix/delimiter in the way that s3fs has. The method used there could be copied (they should be very similar), but of course this takes some effort.
#22 might be of interest to you: it enables you to get info and open files for reading when you already know the specific path and don't need to list the bucket.

bmabey:
Okay, taking a look at the implementation we would need to switch out the use of `list_bucket` to use this endpoint instead: https://cloud.google.com/storage/docs/json_api/v1/objects/list

Is that what you were thinking?

martindurant:
Yes, I think that's it. It would require changing what we store in the cache.

martindurant:
Similar, but not duplicate of #21

asford:
Resolved with #57.

PR #57, opened 2018-01-04
Per-directory metadata cache.

asford:
- [x] Test case updates to reflect new caching logic.
- [x] VCR-test update.
- [x] Docstring updates to clarify file interface.
- [x] Documentation updates for caching logic.
- [x] Profile vs previous cache implementation.
- [x] Integrate cache-deletion logic and tests from #49.
- [x] Update `glob` logic to restrict search to subdirs and prefixes.

@martindurant  *Work in progress solution for #24 and #21, likely supersedes #22. Would you mind taking a quick look at this pull for a sanity check? I've performed some initial integration and manual testing in my deployment and this implementation appears to resolve the primary performance issues I've encountered.*

Refactors the `GCSFileSystem` to operate on a per-directory object metadata cache, rather than a full-bucket cache, to support file reads in buckets with multiple directory structures. This resolves performance issues due to full-bucket listing when reading a subset of keys from a bucket or when `glob`ing within a subdirectory of the bucket.


gcsfs/core.py
```diff
@@ -64,19 +65,33 @@ def quote_plus(s):
     return s
 
 
+def norm_path(path):
+    """Canonicalize path by split and rejoining."""
+    # TODO Should canonical path include protocol?
```
martindurant: Generally speaking, we should strip the protocol as early as possible within this library.

gcsfs/core.py
```diff
@@ -159,6 +174,9 @@ class GCSFileSystem(object):
         (see description of authentication methods, above)
     consistency: 'none', 'size', 'md5'
         Check method when writing files. Can be overridden in open().
+    cache_timeout: float, seconds
```
martindurant, line 257: I like this idea.

gcsfs/core.py
```diff
@@ -344,32 +373,168 @@ def _call(self, method, path, *args, **kwargs):
...
+
+        while next_page_token is not None:
+            page = self._call(
+                'get', 'b/', project=self.roject, pageToken=next_page_token)
+
+            assert page["kind"] == "storage#buckets"
+            items.extend(page.get("items", []))
+            next_page_token = page.get('nextPageToken', None)
+
+        result = {
```
martindurant, line 584: Why the dict? As far as I can see, the only place that this is used, we immediately pick out the 'items' key.

gcsfs/core.py
```diff
@@ -389,7 +554,6 @@ def mkdir(self, bucket, acl='projectPrivate',
                    predefinedDefaultObjectAcl=default_acl,
                    json={"name": bucket})
         self.invalidate_cache(bucket)
-        self.invalidate_cache('')
 
     def rmdir(self, bucket):
         """Delete an empty bucket"""
```
martindurant: I wonder, if you delete the last key within a given prefix, which calls invalidate cache on the parent, do we expect the apparent directory to disappear?
e.g.,

gcs.ls('bucket/')
['bucket/thing/']
gcs.ls('bucket/thing/')
['bucket/thing/key']
gcs.rm('bucket/thing/key')
gcs.ls('bucket/')
[]  # directory should be gone

gcsfs/core.py
```diff
@@ -398,65 +562,77 @@ def rmdir(self, bucket):
             for v in self.dirs[''][:]:
                 if v['name'] == bucket:
                     self.dirs[''].remove(v)
-        self.dirs.pop(bucket, None)
```
martindurant: We do want to remove the entry from a cached bucket listing, no? Also, the references to `dirs` should be renamed anyway.

martindurant: Oh, I see - you expect invalidate, below, to do this. Still true about `dirs`.

gcsfs/core.py
```diff
@@ -398,65 +562,77 @@ def rmdir(self, bucket):
...
-                        out.append(directory)
-        if detail:
-            return out
+        """List objects under the given '/{bucket}/{prefix} path."""
+        if path in ['/', '']:
+            out = self.buckets
         else:
-            return [f['name'] for f in out]
+            if not path.endswith("/"):
+                path = path + "/"
```
martindurant: Is it not in general reasonable to do `ls` on a pull key name, and expect to get the same information as `info` on that file would bring?

gcsfs/core.py
```diff
@@ -398,65 +562,77 @@ def rmdir(self, bucket):
...
+            bucket, key = split_path(path)
+
+            if not detail:
+                result = []
+
+                # Convert item listing into list of 'item' and 'subdir/'
+                # entries. Items may be of form "key/", in which case there
+                # will be duplicate entries in prefix and item_names.
+                item_names = [
+                    f["name"][len(key):] for f in listing["items"]
```
martindurant: Not sure what `len(key):` is doing here.

gcsfs/core.py
```diff
@@ -398,65 +562,77 @@ def rmdir(self, bucket):
...
+                ]
+                prefixes = [p for p in listing["prefixes"]]
+
+                return list(set(item_names + prefixes))
+
+            else:
+                item_details = listing["items"]
+                pseudodirs = [{
+                        'bucket': bucket,
+                        'name': key + prefix,
```
martindurant: key should be bucket?

gcsfs/core.py
```diff
@@ -398,65 +562,77 @@ def rmdir(self, bucket):
...
+                return item_details + pseudodirs
 
     def walk(self, path, detail=False):
+        """ Return all real keys belows path. """
         bucket, prefix = split_path(path)
+
         if not bucket:
-            raise ValueError('Cannot walk all of GCS')
+            raise ValueError(
+                "walk path must include target bucket: %s" % path)
```
martindurant: `path` is always empty here, so it is not very useful to report it. "Path must include at least a bucket" ?

gcsfs/core.py
```diff
@@ -398,65 +562,77 @@ def rmdir(self, bucket):
...
+
         path = '/'.join([bucket, prefix])
-        files = self._list_bucket(bucket)
+
         if path.endswith('/'):
```
martindurant: So by convention directories end with '/' and files do not? The user may expect `walk('bucket/path')` to get files below 'bucket/path/' too; also actual keys may end with '/', although I am not sure how that gets listed with the delimited. Should be a test for this.

gcsfs/core.py
```diff
@@ -500,11 +676,13 @@ def exists(self, path):
...
             else:
-                if bucket in self.ls(''):
+                if bucket in self.buckets:
                     return True
                 else:
                     try:
-                        self._list_bucket(bucket)
+                        # Bucket may be present & viewable, but not owned by
+                        # the current project. Attempt to list.
+                        self._list_objects(path)
```
martindurant, line 780: exists will be True whether path points to a directory or a file?

gcsfs/core.py
```diff
@@ -514,13 +692,37 @@ def exists(self, path):
...
-        if out:
-            return out[0]
-        else:
-            raise FileNotFoundError(path)
+        bucket, key = split_path(path)
+        if not key:
+            # Return a pseudo dir for the bucket root
+            return {
+                'bucket': bucket,
+                'name': "/",
```
martindurant: This may be what a directory entry looks like, but in user-facing methods, the name shoulod be expanded, in this case to 'bucket/'

gcsfs/core.py
```diff
@@ -586,9 +788,9 @@ def rm(self, path, recursive=False):
             for p in self.walk(path):
                 self.rm(p)
```
martindurant, line 896: I wonder if you happen to know if there is a bulk-delete option in GCS?

mrocklin: It would be very nice if this were the case.  Some of @jhamman 's benchmarks with Zarr spend a lot of time removing tiny files.

asford: Unfortunately the GCS api doesn't have a bulk-delete operation. There are a number of possibilities to speed up object deletion. The easiest, and what's implemented in gsutil, would be to issue a number of concurrent delete requests. You probably have a better grounding with if/how this should be integrated into existing async runloop, but [requests-futures](https://github.com/ross/requests-futures) would be an easy, standalone solution.

martindurant:
```
Total time: 9.03903 s
File: /Users/mdurant/code/gcsfs/gcsfs/gcsfuse.py
Function: getattr at line 67

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    67                                               def getattr(self, path, fh=None):
    68       207          292      1.4      0.0          try:
    69       207      8959980  43284.9     99.1              info = self.gcs.info(''.join([self.root, path]))
    70       103          195      1.9      0.0          except FileNotFoundError:
    71       103         1122     10.9      0.0              raise FuseOSError(ENOENT)
    72       104          210      2.0      0.0          data = {'st_uid': 1000, 'st_gid': 1000}
    73       104           87      0.8      0.0          perm = 0o777
    74
    75       104          166      1.6      0.0          if info['storageClass'] == 'DIRECTORY' or 'bucket' in info['kind']:
    76         3            4      1.3      0.0              data['st_atime'] = 0
    77         3            3      1.0      0.0              data['st_ctime'] = 0
    78         3            2      0.7      0.0              data['st_mtime'] = 0
    79         3           16      5.3      0.0              data['st_mode'] = (stat.S_IFDIR | perm)
    80         3            2      0.7      0.0              data['st_size'] = 0
    81         3            3      1.0      0.0              data['st_blksize'] = 0
    82                                                   else:
    83       101        34912    345.7      0.4              data['st_atime'] = str_to_time(info['timeStorageClassUpdated'])
    84       101        21242    210.3      0.2              data['st_ctime'] = str_to_time(info['timeCreated'])
    85       101        20089    198.9      0.2              data['st_mtime'] = str_to_time(info['updated'])
    86       101          326      3.2      0.0              data['st_mode'] = (stat.S_IFREG | perm)
    87       101          129      1.3      0.0              data['st_size'] = info['size']
    88       101           90      0.9      0.0              data['st_blksize'] = 5 * 2**20
    89       101           85      0.8      0.0              data['st_nlink'] = 1
    90
    91       104           76      0.7      0.0          return data

Total time: 0.121938 s
File: /Users/mdurant/code/gcsfs/gcsfs/gcsfuse.py
Function: readdir at line 93

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    93                                               def readdir(self, path, fh):
    94         1            4      4.0      0.0          path = ''.join([self.root, path])
    95         1           34     34.0      0.0          print("List", path, fh, flush=True)
    96         1       121654 121654.0     99.8          files = self.gcs.ls(path)
    97         1          244    244.0      0.2          files = [os.path.basename(f.rstrip('/')) for f in files]
    98         1            2      2.0      0.0          return ['.', '..'] + files
```

Merged this branch into my more-fuse for this, added the following to GCSFS's init:
```
        import line_profiler
        self.prof = line_profiler.LineProfiler(self.getattr, self.gcs.ls,
                                               self.readdir)
        self.prof.enable()
        import atexit, sys
        atexit.register(lambda: self.prof.disable() or
                        self.prof.print_stats(sys.stdout))
```
and called as
```
> gcsfuse pangeo-data ~/gcs
# other terminal
> time CLICOLOR=0 /bin/ls ~/gcs/newmann-met-ensemble-netcdf
real	0m8.660s
```
(the message saying `ls` happened in the terminal running fuse comes quickly)

martindurant:
Reposted from gitter:
 slow directory listing is due to the fact, that although `readdir()` is returning a totaly reasonable list of files and caching the results correctly, `ls` then does `getattr` on those files **and others having filenames prepended with `._`. Since these are not in the parent’s directory listing (because they don’t exist!) a HEAD is called on each of them. 

This may be an osx-specific behaviour.

I think, if the parent directory is in the cache, trying to do `_get_object()` on a file not listed should be NotFound immediately, without trying the HEAD route (use that only where the parent directory isn’t listed).

Your comment `# Should error on missing cache or reprobe?` - yes, should raise

martindurant:
cc @gbrener

martindurant:
On [my branch](https://github.com/martindurant/gcsfs/tree/dir_fuse)
```
$ gcsfuse pangeo-data ~/gcs
# in other terminal
$ time ls ~/gcs/newmann-met-ensemble-netcdf
conus_ens_001.nc  ...

real	0m0.437s
```

martindurant:
@asford , are you meaning to add more here, or is it only the testing that remains outstanding?

asford:
@martindurant 

Sorry for the slow updates on this, the dissertation has sunk it's claws into me...

I'm not intending on pushing any more logic into the pull, it's mostly just testing updates. I'll rebase 1526f51 into a separate pull with an associated issue.

martindurant:
No worries.

This is an example of the currently failing case (raises exception on current master)

```
def test_small_flush(gcs):
    with gcs.open(fn, 'wb') as f:
        f.write(b'data')
        f.flush(force=False)
```

asford:
Noting issues from https://github.com/pangeo-data/pangeo/issues/112:

- [ ] Default directory cache lifetime should be extended, perhaps indefinite in default case?
- [ ] Forced read through cache if file is missing in exists/info call may not be appropriate. Should respect cache? Different timeout?

martindurant:
I think we should certainly respect the cache if we have it, don't probe for files that should already have shown up. When we modify files, the cache should be invalidated. If files are modified elsewhere, it is OK not to know about them until (much) later unless the user explicitly invalidates the cache.

asford:
Agreed and updated.

asford:
@martindurant This is now is a tests-passing state. I've expanded to GCSFileSystem docstring to include the updated object details semantics.

gcsfs/core.py
asford: This is repacking the result in the form of a de-paginated view of the standard GCS object listing. The `prefixes` list it used later to generated pseudo-directory listings for `ls` and `info` calls.

gcsfs/core.py
asford: I've included it in the logging in case there is some kind of malformed input string.

martindurant:
@asford , at some point we floated the idea of restricting the fields that are pulled down with list_objects; we can do that in a future PR (e.g., `ls` only gives just names or names+simple details, but `info()` gives either names+simple details or does a full call for all information), I want to mention it here in case you think any of your work inhibits doing that. I suspect it should be fine.

asford:
It should be doable. I think we should maintain a cache of the raw results from GCS, which is what this pull implements, and then process the cached results as needed to produce the limited listings.

gcsfs/core.py
asford: This the current semantic. Walks targeting `bucket/key` will walk all objects below `bucket/key/`.

martindurant:
Soryr, you still seem to end up with a conflict - I expect it's small.

asford:
I believe this is essentially ready for final review. I'm going to run some more integration-style testing in my environment, and I think there's some testing to be done on pangeo-data/pangeo#112.

asford:
@martindurant Would you prefer to have this rebased to cleanup the commit history or merged as-is?

martindurant:
I am happy to leave the commit history as is, whichever you prefer.

asford:
Great! I'm then +1 to merge.

martindurant:
@asford , having made this big effort, would you like to become a committer on this repo?

asford:
Sounds great! I'd be glad to lend a hand in keeping this feature working; I suspect we'll find a few more bugs in the future.

mrocklin:
> @asford , having made this big effort, would you like to become a committer on this repo?

+1 !

martindurant:
Sorry, just coming back to this now after a weekend away. @mrocklin , do you have the rights to add @asford , I don't think I do.

mrocklin:
I've just sent @asford an invitation to join.  Welcome @asford , we're lucky to have you!

@martindurant I've also just set it so that you have admin rights.

PR #58, opened 2018-01-04
auth refactor

### LINKED ISSUE #54: Credentials should be resolved from metadata service, if available, in default case.
### Report

I'm seeing intermittent errors using `gcsfs==0.3.0` in a `dask.distributed` cluster that, I believe, are due to an error in credential resolution. Briefly, when writing a large number of partitions to gcs via the `dask.bag.to_textfiles` interface, a number of work partitions failed with the log message:

````
Enter the following code when prompted in the browser:

....

  File "/home/fordas/.conda/envs/dev/lib/python3.5/site-packages/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
  File "/home/fordas/.conda/envs/dev/lib/python3.5/site-packages/gcsfs/core.py", line 597, in __setstate__
    self.connect()
  File "/home/fordas/.conda/envs/dev/lib/python3.5/site-packages/gcsfs/core.py", line 249, in connect
    raise RuntimeError("Waited too long for browser"
RuntimeError: Waited too long for browserauthentication.
````

However a number of the work partitions completed successfully. This occured while using multithreaded workers, so I assume that *some* race condition in token retrieval triggered the browser-based authentication flow. Modifying the filesystem interface in `dask.bytes.core` appeared to resolve the issue:

````
import dask.bytes
import gcsfs

import functools

org_fs = dask.bytes.core._filesystems['gcs']
wrapped_fs = functools.partial(org_fs, token="cloud")


dask.bytes.core._filesystems['gs'] = wrapped_fs
dask.bytes.core._filesystems['gcs'] = wrapped_fs
````

### Credential Resolution

Credential resolution in the default case (i.e. `GCSFileSystem(token=None)`) should resolve credentials from the default service account, rather than falling back to the oauth device flow, if the internal metadata service is available. ~~I'm not familiar with the [oauth2client.client.GoogleCredential.get_application_default()](https://github.com/dask/gcsfs/blob/master/gcsfs/core.py#L302) interface currently used in `master`, but if that interface queries instance metadata by default then this is a "doc bug" to clarify this behavior in the gcsfs interface.~~ The `oauth2client.GoogleCredential.get_application_default` interface appears to provide a token for the GCE Service account when invoked on a GCE instance and does not profive the "refresh_token" expected by `gcsfs`'s token refresh logic. This poison's `gcsfs`'s token cache and blocks authorization if `token=None` mode is used from a `GCE` instance.

### Error Handling & Auth in Dask

Pending the behavior clarification above, it seems strange to me that `dask` would invoke `gcsfs` in a mode that *could* require interactive authorization. I believe `gcsfs` should support a `headless` authorization mode which resolves credentials from (a) the `~/.gcs_tokens` token cache (b) the `gcloud` application default credentials and (c) instance metadata, but provides an informative error if interactive oauth would be required, then register this mode (rather than the current `token=None` mode) with dask. 

This would allows users to authenticate manually via the current token=`None` mode, then utilize the cached token for all further access.

### LINKED ISSUE #56: Connect to GCSFS anonymously
I would like to set up GCSFuse in a docker container pointing to a read-only public bucket.  This container will not have any Google authentication by default.  Is it possible to create a `GCSFileSystem` without associating it to a google account and without using `token='cloud'`?

### PR DESCRIPTION
**Pull request by martindurant**
Fixes #54 
Fixes #56 

Trying to turn this around quickly, reviews welcome. Some of the auth modes will not be testable in typical CI or locally.

### CODE REVIEW

**Comment by martindurant** [comment]
So far, I have confirmed manually that using `token=None`, you get the default token or anon if there are no credentials available, and you can select the various connection methods either by setting `token=` as required or by directly calling the various methods. Is it useful to have `token=False` be a proxy to not calling `connect()` at all?

The attribute `method` is stored so that in testing (or introspection) we can tell which auth method finally succeeded.

Note that for now, `token='cloud'` will fail silently (since it would fall back to anon when `token=None`); perhaps we want to warn and/or raise for some combinations.

**Comment by martindurant** [comment]
(ps: tests pass locally in direct mode, since in the common case, the connection is not affected at all by these changes)

File: gcsfs/core.py [hunk 1]
```diff
@@ -124,24 +124,30 @@ class GCSFileSystem(object):
...
-    - if ``token=None``, GCSFS will attempt to use your default google
-      credentials; if this is not found, you will be given a "device code",
-      which you must enter into a browser where you are logged in with your
-      Google identity.
+    The following modes of authentication are supported:
+    - ``token=None``, GCSFS will attempt to guess your credentials in the
+      following order: gcloud CLI default, gcsfs cached token, google compute
+      metadata service, anonymous.
+    - ``token='gtoken'``, your default gcloud credentials will be used, which
+      are typically established by doing ``gcloud login`` in a terminal.
```
mrocklin: Is `gtoken` a standard term?

gcsfs/core.py
```diff
@@ -124,24 +124,30 @@ class GCSFileSystem(object):
...
+    - ``token=='cache'``, credentials from previoudly successful gcsfs
+      authentication will be used (use this after "browser" auth succeeded)
```
mrocklin: spellcheck previoudly

gcsfs/core.py
```diff
@@ -124,24 +124,30 @@ class GCSFileSystem(object):
...
-      ``c``
-
-    We maintain a cache of refresh tokens in the file ~/.gcs_tokens, so for any
-    pair of (project, access), you will not need to log in once your credentials
-    are verified.
+      containing the JSON returned by logging in with the gcloud CLI tool.
+      This may be at
+      ``~/.config/gcloud/application_default_credentials.json``,
+      `` ~/.config/gcloud/credentials``, or
+      ``~\AppData\Roaming\gcloud\credentials``, etc.
```
mrocklin, line 150: Might we also provide the token explicitly as a dictionary?

gcsfs/core.py
```diff
@@ -217,87 +223,123 @@ def load_tokens():
...
-                    data2 = json.loads(r.content.decode())
-                    if 'error' in data2:
-                        continue
-                    data = data2
+        if method not in ['gtoken', 'cache', 'cloud', 'token', 'anon',
+                          'browser', None]:
+            method = 'token'
+        if method is None:
+            for meth in ['gtoken', 'cache', 'cloud', 'anon']:
+                self.connect(method=meth)
```
mrocklin: Perhaps this should return success or failure?  Perhaps it should raise and we should try-except around it?

mrocklin:
cc @asford

mrocklin:
I'm trying this out locally and running into a couple of small problems:

1.  Tests fail 

```python

gcsfs/tests/test_core.py::test_simple FAILED

==================================================== FAILURES =====================================================
___________________________________________________ test_simple ___________________________________________________

token_restore = None

    def test_simple(token_restore):
        assert not GCSFileSystem.tokens
>       gcs = GCSFileSystem(TEST_PROJECT, token=GOOGLE_TOKEN)

gcsfs/tests/test_core.py:18: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
gcsfs/core.py:188: in __init__
    self.connect(method=token)
gcsfs/core.py:316: in connect
    self.__getattribute__('_connect_' + method)()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <gcsfs.core.GCSFileSystem object at 0x7fc407596978>

    def _connect_token(self):
>       if 'type' in self.token or isinstance(self.token, str):
E       TypeError: argument of type 'NoneType' is not iterable

2.  There are merge conflicts with the fuse branch, though I can guess at how to merge short-term

**Comment by asford** [comment]
I'm testing a bit now on a GCE instance I use as a dev box. On this machine I've the compute sdk installed, but I'm using the service account for credentials:

```
fordas@salish:~/distributed_dev/gcsfs$ gcloud auth list
                  Credentialed Accounts
ACTIVE  ACCOUNT
*       353972334481-compute@developer.gserviceaccount.com
```

In this context `auth2client.client.GoogleCredentials.get_application_default` returns an token component that manages access to the service credentials and does not include a `refresh_token`, `client_secret`. It instead provides access tokens via the `get_access_token` interface, which handles token expiration and refresh.

This currently breaks the current `master` implementation, which expects a refresh token to be available. This will also break this implementation.

My recommendation, though it's a bit of a refactor, would be to store the credential object returned from `get_application_default` and then access the access token via `credential.get_access_token().access_token`. I can follow up later today to determine if this will also work for situations in which the user has obtained credentials via `gcloud auth login`.

gcsfs/core.py
martindurant: No, but I wanted something different from the previously provided "cloud" term. "gcloud" might otherwise have been more reasonable.

gcsfs/core.py
martindurant: Yes, you can provide a dictionary - it does say that, but could provide an example.

gcsfs/core.py
martindurant: Basically, if `self.token` remains unset (None), auth failed for that method. I had tried return True/False, but it felt pretty ugly. The exception is `'anon'`, so that could instead give `self.token=False` or other special value.

asford:
I've tested again after setting the application default credentials to my user account via `gcloud auth application-default login`. In this context `auth2client.client.GoogleCredentials.get_application_default` returns a `GoogleCredentials` object that can also handle access token generation and refresh.

Given that behavior, my current recommendation would be to hold the credential object returned by the `get_application_default` and use the interface `credential.get_access_token().access_token` to manage access. `gcsfs` would then not be responsible for managing auth token management and refresh.

In a broader sense, I would suggest emphasizing use of the default app credentials instead of obtaining and directly caching a credential object for `gcsfs`. In "off cloud" contexts, the user can then manage authorization via the `gcloud auth application-default` command, which provides `login`, `revoke`. In "on cloud" this then provides easy access to the default service account, with the option for a user to override this setting by `gcloud auth login`.

asford:
@martindurant How quickly are you trying to turn this around? If you can wait until tomorrow I can open a PR-onto-this-PR that implements the 'application-default' behavior I've described above.

mrocklin:
(for context @asford I've asked @martindurant to accelerate some work here for an upcoming conference talk this coming Monday)

martindurant:
@asford , @mrocklin , would it be a good idea to merge this, to allow for the service account PR to follow and fix merge conflicts in the fuse branch (I can handle that).
Again, vcr tests can wait, but should be re-recorded and working before release.

mrocklin:
This is fine with me.  @asford might have other thoughts.

asford:
No problem with that plan on my end, we've all been there. Let's plan you merging what's needed for your talk on Monday. I can setup a follow-up PR that handles application default credentials afterward.

In light of that, can we agree that the internal API implemented here *may* need to change to reflect the final authorization story. I think we'll be able to continue supporting all the named token modes (`cloud`, `anon`, `browser`, etc...) but I'd like to have a chance to think through the client library integration and make sure it's being used properly.

@martindurant As a side note, it may be better if `gtoken` was named `application-default`, as that's the name of the related `gcloud auth` subcommand.

martindurant:
@asford , I specifically decided against `'application_default'`, because it sounds like we think of it as the default, which it isn't. I've decided that `'google_default'` is sufficiently unambiguous.

Note as another point for follow up, `_call()` should maybe check for the continued validity of the token as given by its expiry and call connect again `self.connect(method=self.method)`.

Issue #61, opened 2018-01-06
Unsuccessful recovery from ConnectionError 


jhamman:
I'm trying to push a large-ish dataset to GCS via xarray/dask/zarr/gcsfs. Things are generally working during the setup and for the first part of the upload. However, after a bit, I'm getting a `ConnectionError` that is not recoverable. I'm pushing from a server at the University of Washington to bucket at "US-CENTRAL1". I would image the network at UW is pretty stable.

#### Version info:
xarray: jhamman:fix/zarr_set_attrs
dask: 0.16.0
zarr: master
gcsfs: master

Details of the full traceback are in this gist: https://gist.github.com/jhamman/25ddda993ad5b768e4b8289904be6779

cc @mrocklin, @martindurant 

xref: https://github.com/pangeo-data/pangeo/issues/48, https://github.com/pydata/xarray/pull/1800

mrocklin:
Out of curiosity was using `retries=` in dask effective?

jhamman:
@mrocklin - yes, retries=10 did seem to solve this. Perhaps that's all we need here. https://github.com/pydata/xarray/pull/1811 may be able to close this issue for me.

mrocklin:
We should still resolve connection issues within gcsfs.

yiga2:
`POST` requests were not wrapped in a retry logic back then but they are now (just). 
@jhamman can this be closed ?

PR #62, opened 2018-01-08
using google.auth

### LINKED ISSUE #59: Invalid use of service credentials return via 'google_default'
*Opening as a tracking issue for follow-up discussed #58. I will be able to follow-up on this issue the week of 2018-01-28. From that thread:*

[Testing #58] on a GCE instance I use as a dev box. On this machine I've the compute sdk installed and I'm I'm using the service account for application default credentials:

```
fordas@salish:~/distributed_dev/gcsfs$ gcloud auth list
                  Credentialed Accounts
ACTIVE  ACCOUNT
*       353972334481-compute@developer.gserviceaccount.com
```

In this context `auth2client.client.GoogleCredentials.get_application_default` returns an token component that manages access to the service credentials and does not include a `refresh_token`, `client_secret`. It instead provides access tokens via the `get_access_token` interface, which handles token expiration and refresh. It may also be used to retrieve auth headers for requests.

This currently breaks the current `master` implementation, which expects a refresh token to be available. ~~This will also break this implementation.~~

After setting application default credentials to a user account via `gcloud auth application-default login`  `auth2client.client.GoogleCredentials.get_application_default` returns a `GoogleCredentials` object exposing  `refresh_token` et al. This credentials object is capable of handling access token and auth header generation.

Given that behavior, my current recommendation would be to hold the credential object returned by the `get_application_default` and use the interface `credential.get_access_token().access_token` to manage access. `gcsfs` would then not be responsible for managing auth token management and refresh. It would also be desirable to wrap the device credentials generated via the `browser` auth mode in a the same oauth handler and remove the token management logic currently implemented in `gcsfs`.

In a broader sense, I would suggest emphasizing use of the default app credentials instead of obtaining and directly caching a credential object for `gcsfs`. In "off cloud" contexts, the user can then manage authorization via the `gcloud auth application-default` command, which provides `login`, `revoke`. In "on cloud" this then provides easy access to the default service account, with the option for a user to override this setting by `gcloud auth login`.

### LINKED ISSUE #60: `oauth2client` is deprecated.
*Ancillary issue that would likely be linked to #59*

As per [the module documentation](https://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html) `oauth2client` is officially deprecated in favor of [`google-auth`](https://google-auth.readthedocs.io). As `google-auth` is an officially support project with a stable API, any further development of the `gcsfs` authorization system should target this library.

The current oauth implementation should likely replaced via a standard implementation using `google-auth` for token managment and the related [`google-auth-oauthlib.flow`](http://google-auth-oauthlib.readthedocs.io/en/latest/reference/google_auth_oauthlib.flow.html) for device authorization.

martindurant:
Use new purpose-built package to generate sessions and handle refresh,
simplifying our code.

Follows from #58 
Fixes #59 
Fixes #60 
May help with #61


mrocklin:
cc @asford, whenever you're around

martindurant:
One outstanding question: if you use the google default credentials, you get the default project too, but I think every other method requires you to provide a project *if* you want to be able to list buckets. anon, of course, makes no use of the project.
Second: passing a token explicitly is now broken. Do we still want to allow a dict as input, or should it be a Credentials object (reminder, there is now a conversion function for the particularl dict style used in the previous version).

Naturally, documentation etc will need to be updated.

mrocklin:
Using a default if available makes sense to me?  I'm not sure I fully
understand.

On the second question I'm pretty sure I don't have enough experience here
to comment intelligently.

asford:
Picking up the default project along with default credentials would be ideal, if possible. This would allow zero-configuration use of `gcsfs` in cases when someone is already using `google-cloud-sdk`, which seems like the most common user scenario.

In my opinion I would be best to just converge on using the `google.auth` API and accept an explicit `Credentials` object. Does `google.auth` have support for reading credentials objects from files in the json format generated by `gcloud auth`?

Thanks for moving quickly on this!

PR #64, opened 2018-01-11
Manage paths starting with '/'

martindurant:
Assume that a path starting with '/' is intended to have the bucket
first, so remove the character and process as before.

See discussion in https://github.com/pangeo-data/pangeo/issues/66


mrocklin:
cc @rabernat

mrocklin:
Thanks @martindurant !

PR #68, opened 2018-02-07
Add missing import needed for `token='cloud'`

bnaul:
Currently using `token='cloud'` gives the following error:
```
GCSFileSystem(project='blah', token='cloud')
...
~/miniconda3/lib/python3.6/site-packages/gcsfs/core.py in _connect_cloud(self)
    235
    236     def _connect_cloud(self):
--> 237         credentials = gauth.compute_engine.Credentials()
    238         self.session = AuthorizedSession(credentials)
    239

AttributeError: module 'google.auth' has no attribute 'compute_engine'
```

Adding the explicit `compute_engine` import fixes this.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 import array
 from base64 import b64encode
 import google.auth as gauth
+import google.auth.compute_engine
 from google.auth.transport.requests import AuthorizedSession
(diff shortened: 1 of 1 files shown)


martindurant:
They must have changed something, because I'm sure this was working (people are using it right now!).
Thanks for the PR.

PR #72, opened 2018-02-14
Fix small file upload

martindurant:
(see a slightly different version in
https://github.com/dask/gcsfs/commit/30be23c53debaf42d89787633f68f6a8c2c855be )

Fixes #70 

cc @asford

Issue #70, opened 2018-02-14
Small file upload fails


bnaul:
```
In [5]: import pandas as pd
   ...: import dask.dataframe as dd
   ...: dd.from_pandas(pd.DataFrame({'x': [1]}), npartitions=1).to_csv(f'gs://{bucket}/test.csv')
   ...:
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
~/miniconda3/envs/model/lib/python3.6/site-packages/gcsfs/core.py in flush(self, force)
    905         if self.location is not None:
--> 906             self._upload_chunk(final=force)
    907         if force:

~/miniconda3/envs/model/lib/python3.6/site-packages/gcsfs/core.py in _upload_chunk(self, final)
    930             headers=head, data=data)
--> 931         validate_response(r, self.location)
    932         if 'Range' in r.headers:

~/miniconda3/envs/model/lib/python3.6/site-packages/gcsfs/core.py in validate_response(r, path)
    131         else:
--> 132             raise RuntimeError(m)
    133

RuntimeError: b"Invalid request.  The number of bytes uploaded is required to be equal or greater than 262144, except for the final request (it's recommended to be the exact multiple of 262144).  The received request contained 7 bytes, which does not meet this requirement."
```

The same behavior happens with a simple string write:
```
In [8]: with gcsfs.GCSFileSystem(project=project).open(f'{bucket}/test.txt', 'w') as f:
   ...:     f.write('test')
...
RuntimeError: b"Invalid request.  The number of bytes uploaded is required to be equal or greater than 262144, except for the final request (it's recommended to be the exact multiple of 262144).  The received request contained 4 bytes, which does not meet this requirement."
```

But weirdly enough, changing the write type to binary resolves this:
```
In [9]: with gcsfs.GCSFileSystem(project='model-streetlight-toronto').open('persona-upload/test.txt', 'wb') as f:
   ...:     f.write(b'test')
   ...:
In [10]:  # OK
```

I'm trying to remember off the top of my head if this is new behavior but I'm not sure.
```
gcsfs==0.0.4
google-cloud-storage==1.7.0
```

martindurant:
This has been reported elsewhere (I can't find it now) and is due to TextIOWrapper calling `flush()` many times on the underlying file - this doesn't happen in binary, which is why that case worked. A fix will be forthcoming.

martindurant:
30be23c53debaf42d89787633f68f6a8c2c855be

PR #73, opened 2018-02-16
Fix flush-on-small-block-size errors and lift block size constants.

asford:
_Potential alternate/add-on commit to #72 cherry-picked from #57 for #70._

Resolve error when writing small partitions via dask.bag.to_textfiles
when partition size is below minimum GCS multipart upload
size.

Close logic in dask.bytes.core calls flush(force=False), followed by
flush(force=True) on GCSFile. Current logic initializes multipart upload
on non-force flush and then attempts to write a non-final block
below the minimum GCS upload block size.

Fixup logic to skip flush if buffer size is below minimum upload size on
non-forced flush and instead issue a warning. This, incidentally, avoids
initialization of multipart upload in cases where final file size will
be below the minimum block size, which was resulting in duplicate
uploads for small output partitions.

Update core.py to lift GCS block size limits into module level
constants. Replace use of constants in core.py with symbolic names.


asford:
@martindurant Sorry for the slip on this from yesterday. Please feel free to merge/rebase/etc with your fix in #72.

mrocklin:
cc @bnaul

martindurant:
Looking shortly. Note that this was also a bug for any TextIOWrapper use of a GCSFile, not specific to dask's read_text.

gcsfs/core.py
```diff
@@ -893,17 +897,33 @@ def flush(self, force=False):
             raise ValueError('Flush on a file not in write mode')
         if self.closed:
             raise ValueError('Flush on closed file')
+
         if self.buffer.tell() == 0 and not force:
             # no data in the buffer to write
             return
+        if self.buffer.tell() < GCS_MIN_BLOCK_SIZE and not force:
+            warnings.warn(
+                "GCSFile.flush(force=False) with buffer size (%s) below minimum GCS chunk size (2 ** 18), "
```
martindurant: I'm not certain I want this warning showing up just  because the user has used text-mode on a gcsfs output. Since the operation doesn't actually do anything, but the data is sure to get written eventually, I would be happy to have this silent.

gcsfs/core.py
```diff
@@ -980,8 +1000,8 @@ def _simple_upload(self):
 
     def _fetch(self, start, end):
         # force read to 5MB boundaries
-        start = start // (5 * 2**20) * 5 * 2**20
-        end = (end // (5 * 2 ** 20) + 1) * 5 * 2 ** 20
+        start = start // (READ_BLOCK_SIZE) * READ_BLOCK_SIZE
```
martindurant: These two lines can be safely removed, google no longer seems to require this.

gcsfs/core.py
```diff
@@ -893,17 +897,33 @@ def flush(self, force=False):
...
         if force and self.forced:
             raise ValueError("Force flush cannot be called more than once")
-        if force and not self.offset and self.buffer.tell() <= 5 * 2**20:
-            self._simple_upload()
-        elif not self.offset:
-            self._initiate_upload()
+
+
+        if not self.offset:
+            if force and self.buffer.tell() <= self.blocksize:
```
martindurant: Actually, I think we should should do this for any file smaller than DEFAULT_BLOCK_SIZE.
Along the same vane, maybe there's no need to call `_initiate_upload` until we've accumulated more than DEFAULT_BLOCK_SIZE, and simple upload might still happen.

gcsfs/core.py
asford: I definitely struggled with this. It seems strange to not warn in the case of an ignored-flush, but I suppose that dropping it would actually match the expected semantic. Dropping to `DEBUG` level logging sound reasonable?

martindurant: Yes, that sounds reasonable.

gcsfs/core.py
asford: I'm not sure I understand the rational for overriding the user-specified block size? Isn't the intended function of that feature to allow explicit control over cutoffs for an upload request, whether simple or chunked?

Agreed on the second point.

martindurant: Hm, OK maybe you're right. Honestly, I don't expect anyone to use a non-standard block size for writing anyway.

asford:
Some expected Travis errors due to read block alignment changes in 05928e47a5e2882bda06bf61024b70a93bd19356. I'm seeing passes locally.

@martindurant Would you mind rerecording locally to verify and update the VCR?

martindurant:
Yes, I can handle new recordings if you are convinced that this is ready to go.

asford:
Yes, this is ready to merge.

PR #74, opened 2018-02-16
Revert "Fix flush-on-small-block-size errors and lift block size constants."

asford:
Reverts dask/gcsfs#73

@martindurant 

I just identified a regression with this change related to flush behavior on read-only files, sorry for the preemptive merge.


martindurant:
Sorry, I don't understand. 
Can you build on the previous commit rather than reverting and starting anew?

asford:
Yes, I'll do that. Opening another pull now.

PR #75, opened 2018-02-16
Flush on open read-only file should be no-op, not error.

asford:
`flush` on an open, but read-only, file should be a no-op rather than raising a ValueError. Compare to builtin file behavior:

```
with open("read_only", "r"):
    r.flush()
```

Fixes regression from #73

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -893,13 +893,14 @@ class GCSFile:
             blocks are allowed to be. Disallows further writing to this file.
         """
 
-        if self.mode not in {'wb', 'ab'}:
-            raise ValueError('Flush on a file not in write mode')
         if self.closed:
             raise ValueError('Flush on closed file')
         if force and self.forced:
             raise ValueError("Force flush cannot be called more than once")
 
+        if self.mode not in {'wb', 'ab'}:
+            assert not hasattr(self, "buffer"), "flush on read-mode file with non-empty buffer"
+            return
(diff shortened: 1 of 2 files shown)


asford:
@martindurant Sorry, but I'm having trouble executing against the default test bucket to update the VCR recordings. As before, I've the tests passing locally including the updated coverage to address the errors I've seen in my environment.

martindurant:
This is what I do to re-record:

```
py.test -vv  # get some failures
rm gcsfs/tests/recordings/test_[test_name].yaml   # for the failing tests
GCSFS_RECORD_MODE=all GCSFS_TEST_PROJECT=<your-project> GCSFS_TEST_BUCKET=<a-bucket> GCSFS_GOOGLE_TOKEN=~/.config/gcloud/application_default_credentials.json py.test gcsfs/tests/test_core.py::test_previously_was_failing -vv   # for each failing test
py.test -vv
```

asford:
Resolved. Was an issue with how I handled authentication, as I'm developing on a GCS instance. For future reference, I needed to (a) login under a user account as the default service credentials rather and (b) explicitly specify the default credentials json.

martindurant:
OK, cool.
I must admit, I don't *like* flush being legal on a read-mode file, but if that's how the standard library is... For the close() after forced-flush, can you add a logging statement when this happens?

asford:
I agree, the behavior was surprising.

Issue #77, opened 2018-02-22
Objects not processed


martindurant:
@asford https://github.com/dask/gcsfs/blob/master/gcsfs/core.py#L559 - should this not be moved to the line above like `"items": [self._process_object(bucket, i) for i in items],` ? As it is, the `items` are being processed and then ignored.

asford:
Yes, this is not well expressed in the current implementation. `_process_object` performs an in-place update of the input object dict as well as returning the modified input dict. This is obviously surprising behavior, as one would expect the function to either modify the object in-place *or* return a modified copy of the object.

The current implementation does function, as https://github.com/dask/gcsfs/blob/master/gcsfs/core.py#L559 is evaluated for the side-effect of updating the object listing, but should be changed. I'll open a PR.

martindurant:
OK, got it. I'll remove the `items =` part at some point then, so I don't get confused again.

PR #79, opened 2018-02-22
Tidy `_process_object` as outlined in #77.

asford:
*From #77:*

Yes, this is not well expressed in the current implementation. _process_object performs an in-place update of the input object dict as well as returning the modified input dict. This is obviously surprising behavior, as one would expect the function to either modify the object in-place or return a modified copy of the object.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -458,10 +458,19 @@ class GCSFileSystem(object):
 
     @classmethod
     def _process_object(self, bucket, object_metadata):
-        object_metadata["size"] = int(object_metadata.get("size", 0))
-        object_metadata["path"] = posixpath.join(bucket, object_metadata["name"])
+        """Process object resource into gcsfs object information format.
+        
+        Process GCS object resource via type casting and attribute updates to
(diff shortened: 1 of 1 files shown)


gcsfs/core.py
```diff
@@ -458,10 +458,19 @@ def buckets(self):
...
+        """Process object resource into gcsfs object information format.
+        
+        Process GCS object resource via type casting and attribute updates to
+        the cache-able gcsf object information format. Returns an updated copy
+        of the object resource.
 
-        return object_metadata
+        (See https://cloud.google.com/storage/docs/json_api/v1/objects#resource)
+        """
+        result = dict(object_metadata)
```
martindurant, line 469: Why this, isn't it already a dict?

asford: The eliminates the side-effect of processing the source object by, instead, creating a shallow copy of the object data. This eliminates the confusing logic you identified in the original bug, in which the function was being called for its side effect but the return value was being discarded.

martindurant: OK, got it. +1 from me.

Issue #83, opened 2018-02-26
invalid_scope: Empty or missing scope not allowed


rabernat:
After overcoming #82, I am on to a new error.

I set my `GOOGLE_APPLICATION_CREDENTIALS` to point to an appropriate .json file. I verified that the config is working with the following code
```python
In [10]: from google.cloud import storage

In [11]: storage_client = storage.Client()

In [12]: list(storage_client.list_buckets())
Out[12]:
[<Bucket: pangeo>,
 <Bucket: pangeo-data>,
 <Bucket: pangeo-data-private>,
 <Bucket: zarr_store_test>]
``` 

Now I am trying to do the same thing from gcsfs. I get an error `invalid_scope: Empty or missing scope not allowed`.

```python
In [8]: fs = gcsfs.GCSFileSystem()

In [9]: fs.buckets
DEBUG:gcsfs.core:_list_buckets(args=(), kwargs={})
DEBUG:gcsfs.core:_call(args=('get', 'b/'), kwargs={'project': 'pangeo-181919'})
ERROR:gcsfs.core:_call exception: ('invalid_scope: Empty or missing scope not allowed.', '{\n  "error" : "invalid_scope",\n  "error_description" : "Empty or missing scope not allowed."\n}')
Traceback (most recent call last):
  File "/nobackup/rpaberna/conda/envs/pangeo/lib/python3.6/site-packages/gcsfs/core.py", line 431, in _call
    r = meth(self.base + path, params=kwargs, json=json)
  File "/nobackup/rpaberna/conda/envs/pangeo/lib/python3.6/site-packages/requests/sessions.py", line 521, in get
    return self.request('GET', url, **kwargs)
  File "/nobackup/rpaberna/conda/envs/pangeo/lib/python3.6/site-packages/google/auth/transport/requests.py", line 198, in request
    self._auth_request, method, url, request_headers)
  File "/nobackup/rpaberna/conda/envs/pangeo/lib/python3.6/site-packages/google/auth/credentials.py", line 121, in before_request
    self.refresh(request)
  File "/nobackup/rpaberna/conda/envs/pangeo/lib/python3.6/site-packages/google/oauth2/service_account.py", line 322, in refresh
    request, self._token_uri, assertion)
  File "/nobackup/rpaberna/conda/envs/pangeo/lib/python3.6/site-packages/google/oauth2/_client.py", line 145, in jwt_grant
    response_data = _token_endpoint_request(request, token_uri, body)
[...]

PR #84, opened 2018-02-27
Add auth scopes

martindurant:
Perhaps solves some auth problems as shown in #83

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -322,8 +322,7 @@ class GCSFileSystem(object):
             credentials = self.tokens[(project, access)]
             self.session = AuthorizedSession(credentials)
 
-    @staticmethod
-    def _dict_to_credentials(token):
+    def _dict_to_credentials(self, token):
         """
         Convert old dict-style token.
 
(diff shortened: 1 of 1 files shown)


bnaul:
I was about to make almost exactly the same PR 😄 . The `google.cloud.storage` client does this as
```
SCOPE = ('https://www.googleapis.com/auth/devstorage.full_control',
         'https://www.googleapis.com/auth/devstorage.read_only',
         'https://www.googleapis.com/auth/devstorage.read_write')

credentials, _ = google.auth.default()
self._credentials = google.auth.credentials.with_scopes_if_required(credentials, SCOPE)
```
I assume they're equivalent but figured I'd share what I was going with just in case there's any subtle difference...

martindurant:
Have you had a chance to use the two methods to see if there is a difference, @bnaul ? Since I've never had any problem gaining credentials...

bnaul:
Both seem to work fine for me; this is running on a k8s pod where the service account has permissions to read/write to a bucket but the node doesn't have any special storage scopes. With the current master I see the same scope issues that have been reported but this resolves it.

martindurant:
OK, excellent - thanks @bnaul

PR #85, opened 2018-02-27
Batch deletion

### LINKED ISSUE #69: Use batch requests for delete
Although there is no batch delete as such, there is such a concept as batch request in GCS: https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch

martindurant:
Fixes #69.
Could be extended to batch access control changes.
Should improve time for creation of zarr datasets where previous contents need to be cleared.


martindurant:
Once more, the tests pass locally, but not on Travis...

PR #92, opened 2018-03-22
Fixes pangeo-data/pangeo #176

martindurant:
Google apparently changed their error message when a key does not exist.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -152,9 +152,9 @@ def validate_response(r, path):
         except:
             msg = str(r.content)
 
-        if "Not Found" in m or 'No such object' in m:
+        if r.status_code == 404:
             raise FileNotFoundError(path)
-        elif "forbidden" in m:
+        elif r.status_code == 403:
             raise IOError("Forbidden: %s\n%s" % (path, msg))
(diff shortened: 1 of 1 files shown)


gcsfs/core.py
```diff
@@ -152,7 +152,7 @@ def validate_response(r, path):
         except:
             msg = str(r.content)
 
-        if "Not Found" in m:
+        if "Not Found" in m or 'No such object' in m:
             raise FileNotFoundError(path)
```
mrocklin: Is there any way to make these checks less fragile?  For example is there a status code that is both sufficiently meaningful and less likely for google to change?

martindurant: According to https://cloud.google.com/storage/docs/json_api/v1/status-codes , yes.

rabernat:
Thanks @martindurant for the quick fix!

mrocklin:
@rabernat are you able to verify that this resolves the issue?

mrocklin:
If so then I'll push a new image up

rabernat:
> @rabernat are you able to verify that this resolves the issue?

Yes, it resolves the issue. 👍

PR #93, opened 2018-03-22
use versioneer

martindurant:


mrocklin:
Thanks!  I appreciate this.

martindurant:
Does this look correct?

mrocklin:
Operationally it seems to work well :)

```python
In [1]: import gcsfs

In [2]: gcsfs.__version__
Out[2]: '0.0.5+6.g69fbff0'
```

I don't know why, but Dask also has the following line in setup.py

          cmdclass=versioneer.get_cmdclass(),

martindurant:
As far as I understood, that it only required if the package has associated executable scripts.

mrocklin:
gcsfuse?

martindurant:
That's a good point :)

PR #97, opened 2018-04-12
Instance creation fixes

martindurant:
- explicitly copy attributes from singleton upon creation, if init
  parameters match
- introduce secure_serialize parameter, which if False, allows passing
  raw credentials arount to avoid re-authentication
- allow mkdir in FUSE (incomplete)

Ref: https://github.com/dask/gcsfs/issues/91


martindurant:
Well OK, I think this is useful in every case regardless.

PR #103, opened 2018-04-24
adding token='cache' use after token='browser' in doc

rsignell-usgs:


martindurant:
Fair enough.
I thought I had already done this, but apparently only in the [docstring](https://github.com/dask/gcsfs/blob/master/gcsfs/core.py#L177).

mrocklin:
Thanks @rsignell-usgs !

PR #109, opened 2018-07-25
 add userProject to ≈all requests, to support requester-pays buckets

ryan-williams:
first pass, works for me locally, perhaps there is a more principled way to do this?

a weird thing is that afaict if you want to write to a bucket you own, but that bucket is requester-pays, you have to provide `userProject` anyway, even though presumably that machinery is not relevant

there are a couple of requests I didn't add it to yet (`put_object`, bulk-delete iirc)

also haven't looked at adding tests

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -516,7 +516,7 @@ class GCSFileSystem(object):
             raise FileNotFoundError(path)
 
         result = self._process_object(bucket, self._call('get', 'b/{}/o/{}',
-                                                         bucket, key))
+                                                         bucket, key, userProject=self.project))
 
         return result
 
@@ -854,7 +854,7 @@ class GCSFileSystem(object):
     def cat(self, path):
         """ Simple one-shot get of file data """
         details = self.info(path)
-        return _fetch_range(details, self.session)
+        return _fetch_range(details, self.session, { 'userProject': self.project })
 
     @_tracemethod
     def get(self, rpath, lpath, blocksize=5 * 2 ** 20):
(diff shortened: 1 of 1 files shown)


ryan-williams:
not super clear on the test failure

>  vcr.errors.CannotOverwriteExistingCassetteException: No match for the request (<Request (GET) https://www.googleapis.com/storage/v1/b/gcsfs-testing/o/?delimiter=%2F&userProject=test_project>) was found. Can't overwrite existing cassette ('/home/travis/build/dask/gcsfs/gcsfs/tests/recordings/test_simple.yaml') in your current record mode ('none').

seems plausible that the new `userProject` param is breaking some assertion, but haven't dug in

martindurant:
I'm afraid *all* the tests will have to be regenerated, because with the extra parameter, the HTTP calls no longer match the ones in the recordings. It might be possible to get around this in the VCR config, but it's tricky territory. In general, I wonder how this maps with the different auth mechanisms. Do calls still succeed when the project is None?

ryan-williams:
Interesting; maybe the "right thing" to do here is gate [adding the `userProject` param] on [whether the bucket is requester-pays].

then, the tests would presumably just run on non-RP buckets like they always have, and the RP-specific paths won't be untested, which is maybe fine.

duplicating the tests (all cases, or a representative sample) on an RP bucket would probably also not be so bad, if we cared enough to do it

martindurant:
This has been implemented a few times and debated a lot in follow-up PRs

PR #110, opened 2018-08-17
Do not subclass dask's FileSystem

martindurant:
The empty parent class was removed


martindurant:
@jcrist , it seems we don't have any dask/gcsfs cross tests

PR #111, opened 2018-09-06
Single-request cat, for trial with zarr

martindurant:
Should reduce latency, when we know we want the whole key every time,
as is the case when accessing gcs via the mapper.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -7,7 +7,7 @@ from __future__ import print_function
 import decorator
 
 import array
-from base64 import b64encode
+from base64 import b64encode, b64decode
 import google.auth as gauth
 import google.auth.compute_engine
 from google.auth.transport.requests import AuthorizedSession
@@ -853,8 +853,13 @@ class GCSFileSystem(object):
     @_tracemethod
     def cat(self, path):
         """ Simple one-shot get of file data """
-        details = self.info(path)
-        return _fetch_range(details, self.session)
+        u = 'https://www.googleapis.com/download/storage/v1/b/{}/o/{}?alt=media'
+        u2 = u.format(*split_path(path))
+        r = self.session.get(u2)
+        r.raise_for_status()
+        md = b64decode(r.headers['X-Goog-Hash'].split('md5=')[1])
+        assert md5(r.content).digest() == md, "Checksum failure"
+        return r.content
 
     @_tracemethod
     def get(self, rpath, lpath, blocksize=5 * 2 ** 20):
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -853,8 +853,13 @@ def url(self, path):
     @_tracemethod
     def cat(self, path):
         """ Simple one-shot get of file data """
-        details = self.info(path)
-        return _fetch_range(details, self.session)
+        u = 'https://www.googleapis.com/download/storage/v1/b/{}/o/{}?alt=media'
+        u2 = u.format(*split_path(path))
```
alimanfoo: The object path needs to be URL quoted, e.g.:

```python
        bucket, object = split_path(path)
        object = quote_plus(object)
        u2 = u.format(bucket, object)
 ```

File: gcsfs/core.py [hunk 2]
```diff
@@ -853,8 +853,13 @@ def url(self, path):
...
+        r = self.session.get(u2)
+        r.raise_for_status()
+        md = b64decode(r.headers['X-Goog-Hash'].split('md5=')[1])
```
alimanfoo: This might not work in future if google decided to add another hash after the md5. Maybe to be safe need to split first on comma, then look for the md5 hash. Also allow for possibility that 'X-Goog-Hash' header is not present, and also allow for possibility that header is present but does not contain an md5 value.

gcsfs/core.py
```diff
@@ -1447,8 +1460,11 @@ def _fetch_range(obj_dict, session, start=None, end=None):
         head = {'Range': 'bytes=%i-%i' % (start, end - 1)}
     else:
         head = None
-    back = session.get(obj_dict['mediaLink'], headers=head)
-    data = back.content
+    r = session.get(obj_dict['mediaLink'], headers=head)
+    r.raise_for_status()
+    md = b64decode(r.headers['X-Goog-Hash'].split('md5=')[1])
+    assert md5(r.content).digest() == md, "Checksum failure"
```
alimanfoo: Could factor out checksum verification code used here and in cat.

martindurant: Ah, actually I don't think this should be here - you don't expect to get any hash when downloading only part of the file.

martindurant:
(note that tests fail, because VCR, the library used for mocking, would need to have the recordings re-rendered for the altered calls)

mrocklin:
Checking in.  What is the status here?

alimanfoo:
Would be great to get this merged, definitely reduces overhead.

martindurant:
I am struggling getting the tests to work properly. I with there was something like moto or azurite for GCS, but I haven't found one yet.

PR #114, opened 2018-10-22
Raise if auth fails; use rewrite instead of copy

martindurant:


gcsfs/core.py
```diff
@@ -434,6 +434,8 @@ def connect(self, method=None):
         else:
             self.__getattribute__('_connect_' + method)()
             self.method = method
+        if self.session is None:
+            raise RuntimeError("Authentication Failed")
```
mrocklin: Is there some way that we can point users towards correct behavior when this happens?  For example perhaps there is a doc page that they should read if this occurs?

martindurant: The best reference is the docstring of this class, so not sure. Could link to the API doc of the same.

By the way, any idea why Travis isn't running tests here?

mrocklin: > By the way, any idea why Travis isn't running tests here?

￼Github failed last night and has spent most of today getting back up to speed.

> The best reference is the docstring of this class, so not sure. Could link to the API doc of the same.

OK, should we point users to that?  I think I trust you to figure out the best way to accomplish this, I'm mostly stating the design goal that exceptions point users towards proper behavior.  If we want people to have a good time with this library (which I think that we do) then pointing them towards correct behavior in stressful situations goes a long way.

mrocklin: I can imagine many solutions to this including:

1.  Writing a doc page
2.  Pointing them to API generated docs of this class
3.  Explaining things in the exception
4.  etc..

I don't know what's best though.  It's likely that there is some better solution that you can come up with given your experience with this library.

PR #115, opened 2018-10-24
Add pyarrow hook to dask shim

martindurant:
Fixes https://github.com/dask/dask/issues/4009

(copies code from fsspec - could have been implemented in pyarrow, or by adapting gcsfs to be a subclass of fsspec)

### DIFF
--- a/gcsfs/dask_link.py
+++ b/gcsfs/dask_link.py
@@ -48,6 +48,72 @@ def register():
         def size(self, path):
             return self.info(path)['size']
 
+        def _get_pyarrow_filesystem(self):
+            """Make a version of the FS instance which will be acceptable to pyarrow
+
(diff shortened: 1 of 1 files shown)


mrocklin:
I take it that there is some code elsewhere that looks for that special method name?

martindurant:
Here: https://github.com/dask/dask/blob/master/dask/dataframe/io/parquet.py#L675 (i.e., our own code) - so this is one of those cases I would like to see refactored.

martindurant:
May find this interesting: https://github.com/martindurant/filesystem_spec/pull/16/commits/92eff11cd85d82b1244210e8fc8ebe6c18185d5d

martindurant:
ping @cdagnino , have you had a chance to check this for your use case yet?

PR #118, opened 2018-11-07
Add metadata methods

martindurant:
Fixes https://github.com/dask/gcsfs/issues/38


martindurant:
(tests not created, recordings not updated)

martindurant:
@yiga2 , a quick implementation that seems to work

martindurant:
@yiga2 , you are welcome to play with this before I merge

yiga2:
@martindurant I did - works fine ! Tested the `put` too. 

Unless no need to hold off, may want to bundle with another (unrelated) change - if applicable - as per my comment [here](https://github.com/dask/gcsfs/issues/12#issuecomment-437725658)

Issue #120, opened 2018-11-14
Service account token, authorization error if passed as dict object


VMois:
I have created a service account to limit access to a bucket. Using gcloud I have generated a key json file that looks like this:

```json
{
  "type": "service_account",
  "project_id": "---",
  "private_key_id": "---",
  "private_key": "---",
  "client_email": "--",
  "client_id": "---",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "---"
}
```

I have tried simple test for reading bucket:

```python
import gcsfs

fs = gcsfs.GCSFileSystem(
    project='project',
    token='/home/<user>/key.json'
)

print(fs.ls('bucket'))

```

And it works as expected with properly permissions etc. But, now, I want to read a json file as a python dict and pass it as a token. New code:

```python
import gcsfs
import json

key_file = open('/home/<user>/key.json', 'r')
key = json.load(key_file)

fs = gcsfs.GCSFileSystem(
    project='project',
    token=key
)

print(fs.ls('bucket'))
``` 

During execution I received the following error:
```
Traceback (most recent call last):
  File "/home/vmois/gstorage_test.py", line 10, in <module>
    token=key
  File "/home/vmois/python-venv/dask/lib/python3.6/site-packages/gcsfs/core.py", line 302, in __init__
    self.connect(method=token)
  File "/home/vmois/python-venv/dask/lib/python3.6/site-packages/gcsfs/core.py", line 421, in connect
    self._connect_token(method)
  File "/home/vmois/python-venv/dask/lib/python3.6/site-packages/gcsfs/core.py", line 385, in _connect_token
    credentials = self._dict_to_credentials(token)
  File "/home/vmois/python-venv/dask/lib/python3.6/site-packages/gcsfs/core.py", line 356, in _dict_to_credentials
    None, refresh_token=token['refresh_token'],
KeyError: 'refresh_token'
```

I'm using the same file but in different formats. Can we somehow improve token handling?

I think some changes in  `_connect_token` method. Line 384 we can add check if a token is service account like on line 378.

martindurant:
[...]

PR #121, opened 2018-11-14
Add check if dictionary token is service account token

VMois:
Related issue: #120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ matrix:
   include:
     - python: 2.7
     - python: 3.6
-    - python: 3.7
 
 install:
   # install fuse libs
@@ -21,9 +20,9 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n test python=$TRAVIS_PYTHON_VERSION pytest requests pandas decorator
-  - source activate test
-  - conda install -c conda-forge google-auth google-auth-oauthlib
+  - conda create -n test python=$TRAVIS_PYTHON_VERSION pytest requests pandas decorator -y
+  - conda activate test
+  - conda install -c conda-forge google-auth google-auth-oauthlib -y
   - pip install vcrpy fusepy
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
 
   # Install dependencies
   - conda create -n test python=$TRAVIS_PYTHON_VERSION pytest requests pandas decorator -y
-  - conda activate test
+  - source activate test
   - conda install -c conda-forge google-auth google-auth-oauthlib -y
   - pip install vcrpy fusepy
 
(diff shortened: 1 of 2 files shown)


martindurant:
I don't see why py35 should have failed where the others passed, but rerunning it just in case it's "one of those things". I don't actually case about py35 any more, can probably change the set matrix to 2.7/3.6/3.7 (and maybe even drop 2.7 in January).

VMois:
Maybe, there is a bug or something that causes failure. On Friday, I will try to change the way dict token is checked and we will check if py35 will fail.

VMois:
I think my code is causing an error. I have tested this version and another way of checking dict object. Version where I'm using checking in `_dict_to_credentials` is failing, but another part where I;m checking inside `_connect_token` is passing all the tests. This is a code bad version code:
```python
    def _dict_to_credentials(self, token):
        try:
            token = service_account.Credentials.from_service_account_info(
                token, scopes=[self.scope])
        except:
            token = Credentials(
                None, refresh_token=token['refresh_token'],
                client_secret=token['client_secret'],
                client_id=token['client_id'],
                token_uri='https://www.googleapis.com/oauth2/v4/token',
                scopes=[self.scope]
            )
        return token
```

For now, I will put dict check to `_connect_token`, but I don't know if it's good solution. Maybe, it's better to figure out how to fix above code. Any thoughts @mrocklin ?

martindurant:
Are you able to run the failing test on py35 with real credentials? It may just be vcrpy funniness See the settings.py file for the env vars you need to set. Whether the service creds check happens in the token making method or where it gets applied shouldn't matter as far as I can tell.

VMois:
@martindurant Sorry for late response. On real credentials py3.5 (using GCSFS_RECORD_MODE=all ) the failing test is successful. 

P.S It's not related to this issue, but when I have changed  GCSFS_RECORD_MODE=once, *test_ls2* is failed. Traceback:
```
    def test_ls2(token_restore):
        with gcs_maker() as gcs:
>           assert TEST_BUCKET in gcs.ls('')
E           AssertionError: assert 'dask-test-run' in ['anaconda-enterprise', 'anaconda-public-data', 'artifacts.test_project.appspot.com', 'test_project_cloudbuild', 'dataflow-anaconda-compute', 'gcsfs-testing', ...]
```

martindurant:
I wouldn't worry about that failure, you seem to be getting the recorded values (i.e., things that make sense on my system).

martindurant:
OK, merging this now. I don't care about py35 any more, and it seems travis does not yet support 37, it seems, so this will do for now.
Thanks for handling things.

alimanfoo:
FWIW Travis does support python 3.7 but needs some tweaks to config, see
e.g. .travis.yml in numcodecs.

martindurant:
Thanks, I'll look it up for next time, when the current google client key will be expiring

PR #123, opened 2018-12-10
Overwrite flag to GCSFileSystem.get method

bachsh:
This comes in handy for me when working with large files and wanting to only copy them once to set an environment

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -903,7 +903,7 @@ class GCSFileSystem(object):
         return r.content
 
     @_tracemethod
-    def get(self, rpath, lpath, blocksize=5 * 2 ** 20):
+    def get(self, rpath, lpath, blocksize=5 * 2 ** 20, overwrite=True):
         """Download remote files to local
 
         Parameters
@@ -914,7 +914,11 @@ class GCSFileSystem(object):
             Local location
         blocksize: int
             Chunks in which the data is fetched
+        overwrite: bool
+            If False, download the file only if local path doesn't exist
         """
+        if not overwrite and os.path.exists(lpath):
+            return
         with self.open(rpath, 'rb', block_size=blocksize) as f1:
             with open(lpath, 'wb') as f2:
                 while True:


gcsfs/core.py
```diff
@@ -914,7 +914,11 @@ def get(self, rpath, lpath, blocksize=5 * 2 ** 20):
             Local location
         blocksize: int
             Chunks in which the data is fetched
+        overwrite: bool
+            If False, download the file only if local path doesn't exist
         """
+        if not overwrite and os.path.exists(lpath):
+            return
```
martindurant, line 921: Should be exception, FileExistsException?

bachsh: why an exception? The reason behind this addition is when you want to download a large file but only want to do so if the file doesn't already exist on your system. If the caller has to catch an exception then it doesn't really make their lives easier

bachsh: Now that I think of it, the real functionality needed here is `rsync`. WDYT?

martindurant: An exception, to my mind, fits in better with the normal file operations: "write to file, no clobber" would be an error if the file exists in any situation I can think of. Perhaps you could rename your parameter something more specific like `ignore_if_local_file_exists`.

Is there a python library for `rsync`? The file-system is, of course, encoded in python, so your helper library would need to be able to call methods within it to be useful for such a situation.

martindurant:
Any change here? I will close as stale if not.

gcsfs/core.py
bachsh: I agree with what you say about the exception, but the point in this change is to make it easy to sync the file. If we need to catch the exception outside there is no point in this. In light of this I think I will retract the PR.

martindurant: In that case, I would insist on three possibilities `overwrite=True` (default, and previous behaviour), `overwrite=False` (exception if exists) or `overwrite='ignore'` (sync only, skip if exists). The latter could be called other possibilities such as "skip".

Issue #119, opened 2018-11-12
Req: specific the Content-Type of the upload


yiga2:
It would be useful to specify the content-type when write to a file, especially if text/plain as browser can render without saving and opening outside.

Requires `GCSFile().open()` to accept, in `wt` or `wb` mode only, an optional `content-type` argument, which is passed to either `simple upload` or `resumable upload` (initialize only).

yiga2:
It's useful for quickly checking the content for troubleshooting, in a browser (through the G console UI) but also the mobile Cloud Console app.

While we are at it, could you add 'content-encoding', same benefit if the content is gzipped text (browser/app decompresses on the fly) - see https://cloud.google.com/storage/docs/transcoding. 
`None` as default.

martindurant:
It would be fine to add content-type and content-encoding to open() and put(). Indeed. maybe arbitrary header parameters could be passed through.
Would you have time to attempt this? Is the documentation clear on how to create VCR recordings for tests?

yiga2:
Sure thing - I will make time on Monday and will read the VCR stuff.

jmwoloso:
@yiga2 any progress on this?

jmwoloso:
i think this is close to being available natively in `gcsfs` already after some experimentation.

you _can_ pass `metadata` via the `open` method and _it does_ get passed along to GCS, however, it doesn't do any overwriting (just duplicating; see screenshot below). I haven't looked at the `gcsfs` implementation yet, but seems like we should be able to pass in any of the `metadata` to the underlying library as headers and/or supported arguments if available.

```
import json
import gcsfs

fs = gcsfs.GCSFileSystem()
j = json.dumps({"success": True})

with open("path_to_bucket/_blob_name.json", "w", metadata={"Content-Type": "application/json"}) as f:
    f.write(j)
```

looking at the file metadata in the GCS bucket we can see the metadata was passed through but it didn't overwrite the `Content-Type` parameter, instead it was just duplicated.

![metadata](https://user-images.githubusercontent.com/7530947/49821844-a7cc9a80-fd38-11e8-9235-7cf9dec779a3.png)

yiga2:
@jmwoloso this is not a duplicate. "your" content-type is custom metadata and is stored actually as `x-goog-meta-Content-Type`. You can pass custom metadata as you did but setting so-called fixed-key metadata require a minor enhancement.

martindurant:
@yiga2 , were you planning on submitting a PR to deal with this?

[...]

PR #124, opened 2019-02-02
feat: #119 Add option for content-type

arpit1997:
**Description**

Solves issue #119. Adds configurable content type for uploads instead of hardcoded `application/octet-stream`. 

**Tests**
 - Added tests for testing if passed content-type is stored or not
 - Added tests for testing content-type is `application/octet-stream` if not provided.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1080,7 +1080,7 @@ class GCSFileSystem(object):
 
     @_tracemethod
     def open(self, path, mode='rb', block_size=None, acl=None,
-             consistency=None, metadata=None):
+             consistency=None, content_type='application/octet-stream', metadata=None):
         """
         See ``GCSFile``.
 
@@ -1092,12 +1092,12 @@ class GCSFileSystem(object):
         const = consistency or self.consistency
         if 'b' in mode:
             return GCSFile(self, path, mode, block_size, consistency=const,
-                           metadata=metadata)
+                           content_type=content_type, metadata=metadata)
         else:
             mode = mode.replace('t', '') + 'b'
             return io.TextIOWrapper(
                 GCSFile(self, path, mode, block_size, consistency=const,
-                        metadata=metadata))
+                        content_type=content_type, metadata=metadata))
 
     @_tracemethod
     def touch(self, path, acl=None, metadata=None):
@@ -1173,7 +1173,7 @@ class GCSFile:
 
     @_tracemethod
     def __init__(self, gcsfs, path, mode='rb', block_size=DEFAULT_BLOCK_SIZE,
-                 acl=None, consistency='md5', metadata=None):
+                 acl=None, consistency='md5', content_type='application/octet-stream', metadata=None):
         """
         Open a file.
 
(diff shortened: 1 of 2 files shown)


yiga2:
Thanks @arpit1997 ! While you are at it, would you consider adding `content-encoding` too ? Default is `None`. See the benefit here: https://cloud.google.com/storage/docs/transcoding

Basically if file is gzipped text, a browser can display its decompressed content on-the-fly. Useful to quickly check w/o having to download+ungzip, including on mobile !

arpit1997:
@yiga2 Yes I can add it. I read about the transcoding and giving an option from our side is not hard but they have mentioned that it's quite easy to specify the wrong content encoding. Maybe we can add the option and add something of a warning that use it wisely.

Meanwhile can you review the current one and see if something needs to be changed? Also can you help out in travis tests and don't know how to fix them.

martindurant:
To make tests pass, you need to make new vcr recordings of all the calls in each test that fails. To do that, you need to set the various environment variables referenced in tests/settings.py, with `GCSFS_RECORD_MODE=all`, and see if things pass. Then they should pass with `pytest` alone and no environment variables; iterate until done. You may need to delete the .yaml file corresponding to the failing test if it passes with the supplied environment variables but you still get those vcr errors without - look in tests/recordings/, each file is named after the test that uses it. Don't forget to check in new files when adding tests.

yiga2:
@arpit1997 @martindurant sorry for the late reaction. 
On second thought, I rather enhance the `setxattrs()` method and use the (new) metadata or metageneration value as indicator that the upload is complete. This is especially important on serverless functions which can terminate in a middle of a loop and execute all over again.

- `PATCH`  instead of `PUT` allows to add, modify and delete
- Uses partial response for performance (only 'metadata')
- Returns full metadata - better than `getattr()` (one attr at a time) or `info()` (all object attributes)

So I have changed as follows:
```
    def setxattrs(self, path, content_type=None, content_encoding=None, **kwargs):
        """ Set/delete/add writable metadata attributes

        Parameters
        ---------
        content_type: str
            Sets the Content-Type. Note: gcsfs defaults to 'application/octet-stream' on uploads
        content_encoding: str
            To enable transcoding - see https://cloud.google.com/storage/docs/transcoding
        kw_args: key-value pairs like field="value" or field=None 
            Custom metadata pairs. Value must be string to add or modify, or None to delete

        Returns
        -------
        Entire metadata after update (even if only path is passed)
        """
        bucket, key = split_path(path)
        metadata = self._call('patch', "b/{}/o/{}", bucket, key,
                              fields='metadata',
                              json={'contentType':content_type,
                                    'contentEncoding':content_encoding,
                                    'metadata': kwargs})
        return metadata
```

Feel free to wrap in a PR and test script.

martindurant:
@arpit1997 , do you intend to add, or are you just stalled on vcrpy?

arpit1997:
Let me have a look on this. I will first get the vcrpy tests passing then
look into it. Will update the PR over the weekend.

aberres:
Any news from this one? Was planning to uplod compressed CSVs and ended up here ;-)

aberres:
I tried @yiga2's approach posted above. It works nicely.

yiga2:
@aberres You don't need to set the content-type to upload a gzip. 

The only benefit of it when used together with content-encoding, is to see its uncompressed content directly in a browser.

Note that above `setxattrs()` has been merged.

martindurant:
This PR has gone somewhat stale, is there anything to be done here?

jodorning:
can this be approved please?

martindurant:
This requires resolving conflicts and re-recording tests. I may have time to do that, but I would not like to promise.

PR #129, opened 2019-02-14
Skip gcsfs registration to dask if dask is not installed

qibin-google:


gcsfs/dask_link.py
```diff
@@ -118,7 +120,6 @@ def walk(self, path, refresh=False):
     dask.bytes.core._filesystems['gs'] = DaskGCSFileSystem
 
 
-try:
+# Register to dask if dask is installed
+if 'dask' in sys.modules:
```
martindurant: This doesn't do what you think, it requires dask to have been already imported. Therefore, if gcsfs happens to be imported before dask, register will never be called.

gcsfs/dask_link.py
```diff
@@ -118,7 +120,6 @@ def walk(self, path, refresh=False):
...
     register()
-except ImportError as e:
-    print(e)
```
martindurant, line 124: I suggest the easiest thing to do here is to change the print to a logger.debug statement, so it doesn't normally show unless the user enables debug logging.

qibin-google: Hi Martin,

Thanks for pointing it out. Yes, logger.debug will be much easier. Thanks for the suggestion.

Best,

Qibin

gcsfs/dask_link.py
```diff
@@ -1,5 +1,10 @@
+import logging
+
 from .core import GCSFileSystem
 
+logging.basicConfig()
```
martindurant: this should be up to the user - for gcsfs, it gets done if the environment variable is set on import

gcsfs/dask_link.py
```diff
@@ -1,5 +1,10 @@
...
+logger = logging.getLogger('dask_link')
```
martindurant: "gcsfs" - this is then consistent throughout the package

gcsfs/dask_link.py
```diff
@@ -121,4 +126,4 @@ def walk(self, path, refresh=False):
 try:
     register()
 except ImportError as e:
-    print(e)
+    logger.debug(e)
```
martindurant, line 128: should be a string

gcsfs/dask_link.py
qibin-google: deleted

gcsfs/dask_link.py
qibin-google: Done.

gcsfs/dask_link.py
qibin-google: Here is an implicit conversion like the current implementation (print(e)). Maybe no need to add str(e). I can change it if you insist.

martindurant: Seems you are correct, sorry for the noise.

PR #131, opened 2019-02-15
Add new type of Exception to handle 429 Errors

gogasca:
Add new type of Exceptions

### DIFF
--- a/gcsfs/utils.py
+++ b/gcsfs/utils.py
@@ -91,12 +91,8 @@ def read_block(f, offset, length, delimiter=None):
 class RateLimitException(Exception):
     """Holds the message and code from cloud errors."""
     def __init__(self, error_response=None):
-        if error_response:
-            self.message = error_response.get('message', '')
-            self.code = error_response.get('code', None)
-        else:
-            self.message = ''
-            self.code = None
+        self.message = error_response.get('message', '')
+        self.code = error_response.get('code', None)
         # Call the base class constructor with the parameters it needs
         super(RateLimitException, self).__init__(self.message)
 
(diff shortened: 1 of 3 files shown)


gcsfs/utils.py
```diff
@@ -88,6 +88,19 @@ def read_block(f, offset, length, delimiter=None):
     return bytes
 
 
+class RateLimitException(Exception):
+    """Holds the message and code from cloud errors."""
+    def __init__(self, error_response=None):
+        if error_response:
```
martindurant: Do you see both variants in practice?

gcsfs/utils.py
```diff
@@ -119,4 +132,6 @@ def is_retriable(exception):
     errs += [str(e) for e in errs]
     if isinstance(exception, HtmlError):
         return exception.code in errs
+    if isinstance(exception, RateLimitException):
```
martindurant, line 132: Can you please include a link that says this type of exception is retriable? If I were google, I could imagine not recommending to retry calls that are already over some rate quota.

gcsfs/utils.py
gogasca: Updated code. Just see one option

gcsfs/utils.py
gogasca: Done. I will retry as client (papermill). Do you suggest the client owns retry logic as this PR? Or gcsfs should do the retries with backoff? (I see some logic in core GCSFileSystem _call method.

martindurant: It is correct that gcsfs should own the retries if we are convinced that the exception should indeed be retried in normal usage.

gogasca: I agree with your initial statement, for 429 Exception, GCS system allows only writing/updating object once per second. In papermill we were are sending more updates than that..., let me know if this PR is good for now or any other changes are recommended. Also how often the gcsfs package gets updated? Thanks

martindurant: I think you are right, and this work well from my point of view.

martindurant:
There is no regular release cycle here, but it can be done whenever I have a little spare time. I'll try to fit one in.

PR #132, opened 2019-02-18
Make all API calls retriable (including on RateLimitException) + setxattrs() to patch metadata

yiga2:
1. wrap `POST` requests into retriable loop
2. setxattrs(): 
a. `PATCH`  (add/modify/delete keys) metadata instead of `PUT` which replaces
b. sets Content-Type and Content-Encoding
c. returns resulting metadata

Not part of this request but retry loop could be extended to the few `session.get()` out there and `put_object()`. Or use one single modified `_call` for all `session` requests

### DIFF
--- a/gcsfs/dask_link.py
+++ b/gcsfs/dask_link.py
@@ -1,4 +1,8 @@
+import logging
+
 from .core import GCSFileSystem
+
+logger = logging.getLogger('gcsfs')
 
 
 def register():
@@ -121,4 +125,4 @@ def register():
 try:
     register()
 except ImportError as e:
-    print(e)
+    logger.debug(e)


yiga2:
`test_attrs` expected to fail due to the proposed `setxattrs()` changes

gcsfs/core.py
```diff
@@ -488,6 +491,25 @@ def _call(self, method, path, *args, **kwargs):
             out = r.content
         return out
 
+    @_tracemethod
+    def _post(self, path, **kwargs):
```
martindurant: Can this not be achieved via `_call(method='post')`?

gcsfs/core.py
```diff
@@ -488,6 +491,25 @@ def _call(self, method, path, *args, **kwargs):
...
+
+        for retry in range(self.retries):
```
martindurant: May be best to factor out the retry logic "call to be retries" and "number of retries" to be a function unto itself, and if it runs out of retries, it raises the last exception.

gcsfs/core.py
```diff
@@ -619,7 +641,7 @@ def _list_buckets(self):
 
         while next_page_token is not None:
             page = self._call(
-                'get', 'b/', project=self.roject, pageToken=next_page_token)
+                'get', 'b/', project=self.project, pageToken=next_page_token)
```
martindurant: interesting that this was not caught before

gcsfs/core.py
```diff
@@ -957,17 +979,29 @@ def getxattr(self, path, attr):
         meta = self.info(path).get('metadata', {})
         return meta[attr]
 
-    def setxattrs(self, path, **kwargs):
-        """ Set user-defined metadata attributes
+    def setxattrs(self, path, content_type = None, content_encoding=None, **kwargs):
```
martindurant: Please fix the spacing

gcsfs/core.py
```diff
@@ -957,17 +979,29 @@ def getxattr(self, path, attr):
...
+        """ Set/delete/add writable metadata attributes
 
         Parameters
         ---------
-        kw_args : key-value pairs like field="value", where the values must be
-            strings. Replaces existing metadata.
+        content_type: str
+            Sets the Content-Type. Note: gcsfs defaults to 'application/octet-stream' on uploads
```
martindurant: "If not specified, defaults to 'application/octet-stream'"

gcsfs/core.py
```diff
@@ -957,17 +979,29 @@ def getxattr(self, path, attr):
...
+        content_encoding: str
+            set 'gzip' + content_type 'text/plain' for transcoding (https://cloud.google.com/storage/docs/transcoding)
```
martindurant: I don't know what this means. What else might be accepted here?

martindurant:
Supersedes #124 ?

yiga2:
Strictly, no, as the original idea - that @arpit1997 implemented - is/was to set the content-type as part of the `_initiate_upload()` or single `post` requests. My proposal is to do through a patch once object exists. 

Both have merits.

gcsfs/core.py
yan-hic: Not sure as the method/request itself would not defaults or change the content_type. It would leave it as initially set, that is `application/octet-stream` unless https://github.com/dask/gcsfs/pull/124 is implemented

Suggested: "Changes the Content-Type"

martindurant: If not None, set the content-type to this value

gcsfs/core.py
yan-hic: Suggested: "Specify 'gzip' as per https://cloud.google.com/storage/docs/transcoding guidance"

Not sure we would want cross-checking. Let the user read the document for proper combinations

martindurant: Is 'gzip' the only possible thing that can be given here, aside from None?

gcsfs/core.py
yan-hic: Not directly as `_call()` returns `response.json` whereas most `post` requests need to complete response to read the headers and/or json body

I am not familiar enough with `requests` to know how to combine.

martindurant: Fair enough, I suppose they are fundamentally different

gcsfs/core.py
yan-hic: `gzip` is the most commonly used and referenced as browsers can gunzip on-the-fly but specs lists 3: https://tools.ietf.org/html/rfc7231#section-3.1.2.1
Like for Content-Type, I would let the user decides the value to pass. No need for validation AFAIC

gcsfs/core.py
yan-hic: I understand and it does make sense to consider using a generic `_call` but this is beyond my expertise so please do all edits you want, or if someone else can take over. 
To add on my previous comment that `post` requests need more than the response json, `_fetch_range` could also benefit from retries but it uses `raise_for_status()`.

To me any request should be retried if error is transient hence retriable.

yiga2:
@martindurant  sorry for having bundled so much in one PR , with intermediary commits. I am struggling with Git.
But in essence, the proposed changes do work in cloud functions using the forked repo. Here there are:
- use a basic `_call` everywhere  (basic because it returns requests.Response and calling functions picks the content it needs) 
- added the new 429 error as retriable
- fixed `setxattrs()` (from the original request) to not pass content_type if None (as it resets it)

I understand it requires formal testing with vcr but would like someone else to take on.

yiga2:
As we have just faced 429 errors after the 4 initial retries, I have also added fine-tuning of the retry logic to match https://cloud.google.com/storage/docs/exponential-backoff + increase to 6

martindurant:
If vcr is the only thing preventing this being green, I'll try to pencil in some time to do the recording for you. OK if I push to your branch?

yiga2:
Absolutely !

Note that some tests failed because in order for a common `_call`, I had to make `_fetch_range()` a class function. Hence args have changed (no need to pass session). 
The only function left is `put_object` in the root module but I think this one is obsolete and/or may not be used by anyone - hard to tell.

PR #134, opened 2019-02-21
check existence of bucket when instantiating GCSFile()

yiga2:
Addresses https://github.com/dask/gcsfs/issues/33

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1214,6 +1214,11 @@ class GCSFile:
         bucket, key = split_path(path)
         if not key:
             raise OSError('Attempt to open a bucket')
+        elif not gcsfs.exists(bucket):
+            raise OSError(f'Bucket {bucket} does not exist')
+        elif mode not in {'rb', 'wb'}:
+            raise NotImplementedError('File mode not supported')
+
         self.gcsfs = gcsfs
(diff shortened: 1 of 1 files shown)


martindurant:
This is reasonable, but I'm not convinces it's necessary: there may be various reasons that you cannot write to a file, such as permissions.

yiga2:
Understood as `exists()` does indeed check on permission (through `ls`). We
are not impacted by this (anymore) – was just cleaning up. The current
generic error is good enough. Let me cancel the PR and close the issue.



*From:* Martin Durant <notifications@github.com>
*Sent:* Thursday, February 21, 2019 12:00 PM
*To:* dask/gcsfs <gcsfs@noreply.github.com>
*Cc:* Yannick Einsweiler <ye@hintsight.net>; Author <
author@noreply.github.com>
*Subject:* Re: [dask/gcsfs] check existence of bucket when instantiating
GCSFile() (#134)



This is reasonable, but I'm not convinces it's necessary: there may be
various reasons that you cannot write to a file, such as permissions.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/dask/gcsfs/pull/134#issuecomment-466079226>, or mute
the thread

PR #137, opened 2019-03-05
Add GCSFileSystem.get_dir

davidbrochart:
Recursively download remote directory to local

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -932,7 +932,9 @@ class GCSFileSystem(object):
                 lsubpath = os.path.join(lpath, subpath)
                 rsubpath = os.path.join(rpath, subpath)
                 ldirname = os.path.dirname(lsubpath)
-                os.makedirs(ldirname, exist_ok=True)
+                if not os.path.exists(ldirname):
+                    # python2 doesn't have exist_ok argument in makedirs
+                    os.makedirs(ldirname)
             else:
                 rsubpath = rpath
                 lsubpath = lpath
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -906,32 +906,7 @@ class GCSFileSystem(object):
         return r.content
 
     @_tracemethod
-    def get_dir(self, rpath, lpath, blocksize=5 * 2 ** 20):
-        """Download remote directory to local
-
-        Parameters
-        ----------
-        rpath: str
-            Remote location
-        lpath: str
-            Local location
-        blocksize: int
-            Chunks in which the data is fetched
-        """
-        os.mkdir(lpath)
-        if not rpath.endswith('/'):
-            rpath += '/'
-        rls = self.ls(rpath)
-        for rsubpath in rls:
-            if rsubpath.endswith('/'):
-                lname = os.path.basename(rsubpath[:-1])
-                self.get_dir(rsubpath, f'{lpath}/{lname}')
-            else:
-                lname = os.path.basename(rsubpath)
-                self.get(rsubpath, f'{lpath}/{lname}', blocksize)
-
-    @_tracemethod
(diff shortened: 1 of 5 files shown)


gcsfs/core.py
```diff
@@ -905,6 +905,31 @@ def cat(self, path):
...
+            Local location
+        blocksize: int
+            Chunks in which the data is fetched
+        """
+        os.mkdir(lpath)
+        if not rpath.endswith('/'):
+            rpath += '/'
+        rls = self.ls(rpath)
+        for rsubpath in rls:
+            if rsubpath.endswith('/'):
```
martindurant: Keys ending with '/' are allowed, and are not necessarily directories

gcsfs/core.py
```diff
@@ -905,6 +905,31 @@ def cat(self, path):
...
+        Parameters
+        ----------
+        rpath: str
+            Remote location
+        lpath: str
+            Local location
+        blocksize: int
+            Chunks in which the data is fetched
+        """
+        os.mkdir(lpath)
```
martindurant: Might exist - is OK to raise exception here?

gcsfs/core.py
```diff
@@ -905,6 +905,31 @@ def cat(self, path):
                     assert md5(r.content).digest() == md, "Checksum failure"
         return r.content
 
+    @_tracemethod
+    def get_dir(self, rpath, lpath, blocksize=5 * 2 ** 20):
```
martindurant: Why this name? I would have thought perhaps `sync` or `get(path, recursive=True)`.

gcsfs/core.py
```diff
@@ -905,6 +905,31 @@ def cat(self, path):
...
+            Remote location
+        lpath: str
+            Local location
+        blocksize: int
+            Chunks in which the data is fetched
+        """
+        os.mkdir(lpath)
+        if not rpath.endswith('/'):
+            rpath += '/'
+        rls = self.ls(rpath)
```
martindurant: Could use `walk` here to avoid recursion?

davidbrochart:
Thanks @martindurant, I tried to take your suggestions into account in my last commit.

gcsfs/core.py
```diff
@@ -942,14 +917,31 @@ def get(self, rpath, lpath, blocksize=5 * 2 ** 20):
...
         """
-        with self.open(rpath, 'rb', block_size=blocksize) as f1:
-            with open(lpath, 'wb') as f2:
-                while True:
-                    d = f1.read(blocksize)
-                    if not d:
-                        break
-                    f2.write(d)
+        if recursive:
+            rpath = os.path.join(rpath, '') # add trailing slash
```
martindurant: `os.path` will fail on windows, use explicit "/" instead

davidbrochart:
Thanks @martindurant, it should be good now, please let me know if you see anything else.

martindurant:
Thank you, this is going in.

If you have time, you might be interested in https://filesystem-spec.readthedocs.io/en/latest/ , an attempt to make an abstraction of the various file-systems that are around (including [gcsfs](https://filesystem-spec.readthedocs.io/en/latest/)). There is an [issue](https://github.com/martindurant/filesystem_spec/issues/17) there around using globs in various operations, and this idea of recursion could fall there too. If you have the impetus, you might like to sketch out or even make a PR for how to generalise what's where to fsspec, to globs and directory trees, and to all the methods that might use them. btw: `walk` in the sense you have used it here is called `find` in fsspec, where `walk` produces iterators as `os.walk` does.

davidbrochart:
Thanks for the merge and for the guidance @martindurant! I will look at filesystem_spec. Also, a `put(recursive=True)` could be useful.

martindurant:
Indeed - and I imagine you will want to factor out the functionality which gets the common root of a set of files and applies a different root on the target file-system.

PR #139, opened 2019-03-12
Add GCSFileSystem.put(recursive=True)

davidbrochart:
This is the counterpart of GCSFileSystem.get(recursive=True) (PR #137).

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -948,8 +948,8 @@ class GCSFileSystem(object):
 
     @_tracemethod
     def put(self, lpath, rpath, blocksize=5 * 2 ** 20, acl=None,
-            metadata=None):
-        """Upload local file to remote
+            metadata=None, recursive=False):
+        """Upload local files to remote
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


martindurant:
Given that the code here is nearly identical to `get()` one from before, can you think of a way to abstract out the code?

davidbrochart:
Not sure, for instance one might think that `GCSFileSystem.walk` and `os.walk` are similar, but the former gives a file for each iteration, whereas the latter gives a list of files for each iteration. Also, they are slightly asymmetrical:
- `get('a', 'b', recursive=True)` will copy **the content** of 'a' in 'b'
- `put('a', 'b', recursive=True)` will copy 'a' (and its content) in 'b'

davidbrochart:
Maybe this needs further specification. Following the behavior of the `cp` Unix-like command:
- `get/put('a', 'b', recursive=True)`:
if directory `b` exists locally/remotely, copy `a` inside `b` (we'll have `b/a`).
if directory `b` doesn't exist locally/remotely, copy `a` as `b`.

If we agree with that, I'll review both get and put.

martindurant:
I suppose that's correct, but with the added twist that on gcsfs, a particular path can be both a file and a directory; and that "/" on the end of the path should not be determinate.

davidbrochart:
I rewrote `get` and `put`. This is much cleaner now, and it respects the rules that I previously mentioned. Still, I could not abstract out the code from the two methods because of small differences here and there between local file system and GCS file system.

martindurant:
Thank you, I will merge this now. 
I still think there is a good opportunity for cross-filesystem actions like this to be formulated in terms of fsspec, but that can wait.

PR #144, opened 2019-04-08
Exclude explicit method "cloud" search for compute_engine.Credentials().

### LINKED ISSUE #143: method=None inclusion of "cloud" duplicates google_default=google.auth.default but with bad timeouts
`gcsfs` has some local logic it applies when the caller does not specify the desired authentication method.  This logic is overlapped with precedence logic in `google.auth.default()` `Credentials` finder, but the latter seems to take more care to not test for `cloud` credentials (`compute_engine.Credentials()`) when not running in that environment.

Perform the following tests when running in a non Google Cloud environment:

```
gcs = GCSFileSystem(check_connection=False)
gcs.ls('anaconda-public-data/nyc-taxi/csv/2015/')
```

The result is a slow retry/timeout, finishing with an exception:

```
_call out of retries on exception: HTTPConnectionPool(host='metadata.google.internal', port=80): Max retries exceeded with url: /computeMetadata/v1/instance/service-accounts/default/?recursive=true (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe238e555f8>: Failed to establish a new connection: [Errno -2] Name or service not known'))
  File "/opt/conda/lib/python3.7/site-packages/urllib3/connection.py", line 159, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw)
  File "/opt/conda/lib/python3.7/site-packages/urllib3/util/connection.py", line 57, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/opt/conda/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known
```

If you run instead:

```
gcs = GCSFileSystem(token='anon', check_connection=False)
gcs.ls('anaconda-public-data/nyc-taxi/csv/2015/')
```

... you get instant success when `anon` is specified.

If you run instead:

```
gcs = GCSFileSystem(token='google_default', check_connection=False)
gcs.ls('anaconda-public-data/nyc-taxi/csv/2015/')
```

... and `google_default` brings a near-instant exception - can't authenticate this way.

`cache` fails equally quickly.

The sequence for the iterate/auto mode seems identical to what `google.auth.default()` does, except its `_get_gce_credentials` seems to be able to make an abort decision before the metadata registry is attempted to be connected at `metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/?recursive=true`, which will fail (slowly) in general environments.

Proposal is to remove the explicit `cloud` auth test from the loop, since the logic is already included in `google_default`, but more-safely.

#99 and #100 are related.

javabrett:
This is already covered by google_default (google.auth.default()) but with superior detection
and avoiding expensive connection timeouts e.g. to metadata.google.internal.

Added documentation on token='anon' option.

Fixed #143.

I don't know that we have any integration tests to cover this.  Is anyone able to run this in GCE to confirm that metadata credentials are still found by `google_default` (as they should be).

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -98,6 +98,9 @@ Several modes of authentication are supported:
     - you can also generate tokens via Oauth2 in the browser using ``token='browser'``,
       which gcsfs then caches in a special file, ~/.gcs_tokens, and can subsequently be accessed with ``token='cache'``.
 
+    - anonymous only access can be selected using ``token='anon'``, e.g. to access
+      public resources such as 'anaconda-public-data'.
+
 The acquired session tokens are *not* preserved when serializing the instances, so
 it is safe to pass them to worker processes on other machines if using in a
 distributed computation context. If credentials are given by a file path, however,
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -426,7 +426,7 @@ class GCSFileSystem(object):
                           'browser', None]:
             self._connect_token(method)
         elif method is None:
-            for meth in ['google_default', 'cache', 'cloud', 'anon']:
+            for meth in ['google_default', 'cache', 'anon']:
                 try:
                     self.connect(method=meth)
                     if self.check_credentials and meth != 'anon':


martindurant:
So, to be sure: if method is 'google_default', you will get the "cloud" variant automatically, if it is available? It it possible that the default picks another method, but the user want to force the "cloud" method?

javabrett:
I don't currently have the capacity to test this myself on GCE.  Problems were found running locally or in AWS, where `compute_engine` is looked-for and asked to access the metadata registry at `metadata.google.internal`, which of-course fails everywhere except a running GCE environment.  Problem is it fails with a long hostname-lookup timeout etc.

Passing `token`/`method` = `None` will force `google_default` first:

https://github.com/dask/gcsfs/blob/08475833e5df090e57a5256c98f9503aedc2c20b/gcsfs/core.py#L428-L429

`_connect_google_default` calls `google.auth.default`:
https://github.com/dask/gcsfs/blob/08475833e5df090e57a5256c98f9503aedc2c20b/gcsfs/core.py#L335-L336

`google.auth.default` lays out its 5-step default priority:

https://github.com/googleapis/google-auth-library-python/blob/2c6ad78917e936f38f87c946209c8031166dc96e/google/auth/_default.py#L220-L252

Step 4 is the GCE which involves the metadata hunt, but only after it decides it is worth proceeding.

That code is:

https://github.com/googleapis/google-auth-library-python/blob/2c6ad78917e936f38f87c946209c8031166dc96e/google/auth/_default.py#L190-L217

Looks OK to me.  If `token='google_default'` is passed explicitly, we should see the same behaviour but with no failover to `cache` or `anon`.  `cloud` seems duplicate and unnecessary in the loop.  `cloud` can still be explicitly requested.

Would be nice to get this tested in GCE though.

martindurant:
OK yes, since "cloud" is still an option for those that want it, I am fine with removing it from the set that are tried automatically - especially since the google method should pick it up anyway.

martindurant:
Thanks for doing this.

PR #150, opened 2019-06-05
Canonicalize remote path in `GCSFileSystem.get`

### LINKED ISSUE #149: URI's protocol is not removed in GCSFileSystem.get
I've been using `GCSFileSystem.get` with `recursive=True`, and ran into some unexpected (to me) behavior: the first few characters of the local file names were missing. It looks like this happens when a `gs://` protocol is in the path.

After some digging, I realized that the recursive `get` is using `GCSFileSystem.walk` to get the paths to copy (which returns canonicalized paths from `gcsfs.core.norm_path`), but then it uses the length of original remote path (including the `gs://` protocol) to piece together the output paths.

My question is: Is it required that there is no protocol in the path passed to `get`?
If not, I'm happy to open up a PR that simply uses `norm_path` in `get`. Wanted to make sure I'm not just using the library in an unexpected way first though.

pakelley:
Resolves #149

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -308,11 +308,12 @@ def test_get_put(token_restore):
             assert gcs.cat(TEST_BUCKET+'/temp') == data
 
 
+@pytest.mark.parametrize("protocol", ['', 'gs://', 'gcs://'])
 @my_vcr.use_cassette(match=['all'])
-def test_get_put_recursive(token_restore):
+def test_get_put_recursive(token_restore, protocol):
     with gcs_maker(True) as gcs:
         with tempdir() as dn:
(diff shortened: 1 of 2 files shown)


gcsfs/tests/test_core.py
```diff
@@ -308,19 +308,20 @@ def test_get_put(token_restore):
             assert gcs.cat(TEST_BUCKET+'/temp') == data
```
pakelley, line 310: `parametrize` seemed like a simple way of making sure different protocols work, but I'm happy to make an explicit test for that if that's preferred

pakelley:
amended that commit with a couple lines in the unit test I missed

martindurant:
Thanks!

Issue #151, opened 2019-06-10
Can't create GCSFileSystem with instance of `google.oauth2.service_account.Credentials`


jiajie-chen:
When trying to create an instance of `GCSFileSystem` by passing in a service account `Credentials` object, an error is thrown in the `_connect_token` method.

**System Details:**
Python version 3.6.7 64-bit, on Ubuntu 18.04.1 (WSL)
`gcsfs==0.2.2`
`google-auth==1.6.3`
`google-auth-oauthlib==0.3.0`

**Expected Behavior:**
For `GCSFileSystem` to successfully connect with provided the service account credentials.

**Actual Behavior:**
`ValueError: Token format no understood`

**Steps to Recreate:**
Run this code (fill in the relevant details):
```
import google.oauth2.service_account
import gcsfs

# Make google.oauth2.service_account.Credentials object
key_path = "/path/to/key_file.json"
scopes = ["https://www.googleapis.com/auth/devstorage.full_control"]
credentials = (
    google.oauth2.service_account.Credentials.from_service_account_file(
        key_path, scopes=scopes)
)
# Try to make gcsfs (will raise "ValueError: Token format no understood")
project="PROJECT-NAME"
gcsfs.GCSFileSystem(project=project,token=credentials)
```

leehart:
I'm hitting similar confusion with a service account, where the class of object seems ambiguous.
Maybe tripping on `elif isinstance(token, Credentials):` in `core.py`

I can't pass the credentials of a service account as a token (maybe I shouldn't be doing this), e.g. 
```
gcs = gcsfs.GCSFileSystem(project='myproject-jupyterhub', token=gcs_service_account.session.credentials)`
```

The type of the credentials object seems to vary:
```
gcs_service_account.session.credentials
<google.oauth2.service_account.Credentials at 0x7fa6d25aa6d8>
```
```
gcs_service_account.session.credentials
<google.auth.compute_engine.credentials.Credentials at 0x7fd2697608d0>
```

So I guess the check for a valid Credentials object is being confused by the namespace clash? e.g.

```
from google.oauth2.credentials import Credentials as oauth2Credentials
isinstance(gcs_service_account.session.credentials, oauth2Credentials)
```

```
from google.auth.compute_engine import Credentials as authCredentials
isinstance(gcs_service_account.session.credentials, authCredentials)
```

Or maybe this is expected behaviour and the wrong way to roll?

jiajie-chen:
[...]

PR #152, opened 2019-06-11
Fix for Issue #151 - Google Credentials isinstance check

jiajie-chen:
Fixed issue #151 by changing the `isinstance(token, Credentials)` check in `_connect_token` to `isinstance(token, google.auth.credentials.Credentials)`.

This should allow support for any kind of Google Credentials object when authentication with Cloud Storage. Further testing needs to be done to confirm this, however.


martindurant:
I think this is the right thing to do, thank you.

jiajie-chen:
No problem, thanks for merging!

PR #153, opened 2019-06-12
add timeout parameter

aviau:
There is a risk of requests hanging forever if we don't set a timeout.


martindurant:
Good idea! I'll look at this soon. Do you think it's possible to test?

aviau:
I am not fond of timing-dependent tests, and my research shows no such feature in vcrpy.

Current tests are not failing so this shows that the parameter is passed down correctly without breaking anything.

martindurant:
OK, that's fair.

PR #156, opened 2019-06-18
Correct conda channel.

ecrows:
The current command doesn't work with default conda config.  Thanks for the great library.


martindurant:
Hah, quite surprised that this was wrong, thanks for correcting.

PR #165, opened 2019-08-15
Default credentials check to false

### LINKED ISSUE #164: Credentials check fails when using VPC Service Controls
I tried using gcsfs on a VM, using its metadata credentials, in a project contained within a [VPC service perimeter](https://cloud.google.com/vpc-service-controls/docs/overview). gcsfs successfully fetches the credentials, but when it then goes onto [check the credentials by accessing a public GCS bucket](https://github.com/dask/gcsfs/blob/0.3.0/gcsfs/core.py#L412), it fails because the perimeter blocks access to Google APIs for other projects.

I'm aware you can disable the check, but this isn't straightforward in my case, because I'm using [pandas](https://github.com/pandas-dev/pandas/pull/20729) and it doesn't expose the relevant parameter that disables the check.

This may well trip other users using VPC service controls.

leg100:
Fixes #164.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -273,7 +273,7 @@ class GCSFileSystem(fsspec.AbstractFileSystem):
     def __init__(self, project=DEFAULT_PROJECT, access='full_control',
                  token=None, block_size=None, consistency='none',
                  cache_timeout=None, secure_serialize=True,
-                 check_connection=True, requests_timeout=None, **kwargs):
+                 check_connection=False, requests_timeout=None, **kwargs):
         super().__init__(self, **kwargs)
         pars = (project, access, token, block_size, consistency, cache_timeout)
         if access not in self.scopes:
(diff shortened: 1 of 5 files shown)


martindurant:
This failed because one of the calls to GCS is presumably subtly different from how it was before, and vcrpy is *very* finicky. Would you like to have a go at [re-recording](https://gcsfs.readthedocs.io/en/latest/developer.html#testing-and-vcr) the offending test? I've of course had some practice at this, but I don't know when I would get around to doing it for you.

leg100:
Has anyone other than yourself ever recorded the tests? Because the records all have references to the bucket `gcsfs-testing`, which I don't have access to, but presumably you and Travis CI do?

I could override it with my own bucket, but then the records would reference my bucket, and Travis CI won't have access to it...

martindurant:
It's complicated ;)

The file settings.py details how various values are used: the remote *real* bucket used is determined by the `GCSFS_TEST_BUCKET` env var, and at https://github.com/dask/gcsfs/blob/master/gcsfs/tests/utils.py#L33 gets replaced by "gcsfs-testing". This is so that the bucket and credentials used in testing do *not* leak into the public repo.

https://gcsfs.readthedocs.io/en/latest/developer.html#testing-and-vcr

martindurant:
I pushed the recording to this branch, I hope you don't mind.

martindurant:
Thank you!

PR #175, opened 2019-09-12
Make 502 a retriable error

benjeffery:
I've been getting rare (1-in-a few hundredish) `502` errors on writes. In the current code these are not retried. The logic in `is_retriable` includes `502` but only if an `HttpError` exception in raised. In `validate_response` an `HttpError` exception is only raised if the response body is valid JSON with a `message` key. The responses I've been getting from GCS are html pages so an `HttpError` is not raised, and `is_retriable` returns false.

There are two possible fixes for this:

- Raise `HttpError` for errors without a JSON message
- Raise a specific error for `502`'s

This code takes the second option as it seems more conservative.

Thanks for GCSFS!

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -31,7 +31,7 @@ import time
 import warnings
 import random
 
-from requests.exceptions import RequestException
+from requests.exceptions import RequestException, ProxyError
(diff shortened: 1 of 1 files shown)


benjeffery:
Closing to replace with #177

Issue #186, opened 2019-10-22
Performance regression in 0.3.0


bnaul:
When opening a file in a bucket with a lot of files (500k in my case), I see a huge decrease in performance between 0.2.3 and 0.3.0 through master.

Before:
```
[nav] In [1]: import gcsfs
         ...: fs = gcsfs.GCSFileSystem()
         ...: %time fs.open('gs://<bucket>/89257551a5d8c84ba07c42c6af490010.txt').read()
CPU times: user 2.98 ms, sys: 1.18 ms, total: 4.16 ms
Wall time: 63.8 ms
```
Now:
```
[nav] In [1]: import gcsfs
         ...: fs = gcsfs.GCSFileSystem()
         ...: %time fs.open('gs://<bucket>/89257551a5d8c84ba07c42c6af490010.txt').read()
CPU times: user 11.9 s, sys: 4.19 s, total: 16.1 s
Wall time: 2min 55s
```

I don't see this behavior for the same file in a nearly empty bucket, so I'm relatively sure it has something to do with the number of files.

3 minutes is roughly how long it took me to `gsutil ls | wc` the bucket, maybe there's an inadvertent list of all files going on...?

TomAugspurger:
It does look like a `info` is called https://github.com/intake/filesystem_spec/blob/e6b7e1c6b85553f0f4df17f8a8ea8ee049de1f5a/fsspec/spec.py#L871-L877. The spec docstring says


> The default implementation should calls ls and could be overridden by a
> shortcut. kwargs are passed on to ```ls()``.

@martindurant would / could gcsfs implement a faster `info`? Does s3fs need the same?

martindurant:
s3fs does this already, a HEAD request in `info()`, if the parent directory does not already exist in the cache.

TomAugspurger:
Sounds good. @bnaul would you be willing / able to make a PR? Otherwise I
can try to make one by the end of the week.

PR #187, opened 2019-10-22
Implement GCSFileSystem.info() without full bucket ls

bnaul:
Fixes \#186. I basically just took the base `.info()` that @TomAugspurger linked to and tried to `ls()` the specific path instead of the bucket; it seems pretty likely that I'm missing something here since the `s3fs` analog seems a fair bit more complicated...? But it seems to work for a simple test case.


TomAugspurger:
I guess the original issue is the `self.ls(self._parent(path))` in https://github.com/intake/filesystem_spec/blob/e6b7e1c6b85553f0f4df17f8a8ea8ee049de1f5a/fsspec/spec.py#L509

@martindurant do you recall why that's necessary, rather than just doing `self.ls(path, detail=True)`?

Not sure what's up with the CI.

martindurant:
Don't worry about the CI, vcrpy is evil and I can sort it out.

Before going ahead with this, we should consider the use cases: where you want the details of one file among many in a bucket (or prefix), this is clearly faster; this is probably what dask workers normally want. When you want the details of all files, this is clearly slower; this is what the dask client normally wants. 

At the very least, we should check if the parent directory is in the cache, and if it is, use the info there. That should cover walk and glob cases.

The reason for the current implementation, by the way, is that for some backends (ssh, ftp, maybe others), you cannot "list" a single file at all.

martindurant:
On the other hand, a number of the CI problems seem real: to do with files existing and info on directories...

TomAugspurger:
> When you want the details of all files, this is clearly slower; this is what the dask client normally wants.

Isn't that best determined by the caller? If you want the details of all the files aren't you doing `self.ls(directory, details=True)` I don't see why a File.info would need to concern itself with that.

martindurant:
We are hoping to make access times best for people by making a reasonable choice, but we don't know what access pattern they intend to make. I suppose we should follow what s3fs does, and check for parent dir in the cache and use direct info if not there.

TomAugspurger:
You know better than me, but what use-case is served by FileSystem.info()
doing an ls on the parent directory?

martindurant:
The one where you intend to access many/most files in the same directory, but didn't have the foresight to do the `ls()` beforehand. That would happen when opening all of the files in a parquet dataset, for instance, if the paths were taken from metadata rather than by walking the directory tree (I am *not* talking about a Dask workflow here).

martindurant:
Nevertheless, for this PR, the code should check whether the info is already in the cache, and not call remote at all if so. A number of tests seem to be genuinely failing on exactly a loss of sync between expected and actual, i.e., any of the ones that *don't* have vcr stuff in the traceback.

bnaul:
^ ignore that for now, the vcr tests are tripping me up so I need to dig a little more to see what the expected return for `.info()` is for files vs buckets; I think I should be able to sort it out but for vcrpy I'll definitely need some assistance :)

martindurant:
Short version: real http calls can change, so vcr often doesn't find the match when you come back to run a test. You will need to delete the offending YAML file out of tests/recordings, and rerun the test with a set of GCSFS_* environment parameters as listed in the settings.py file https://gcsfs.readthedocs.io/en/latest/developer.html#testing-and-vcr
(then rerun pytest again without those variables to ensure it worked).

bnaul:
all right, I think I have a better handle on what's going on now:
```
# info on file; does not hit cache
fs.info('gs://gcsfs_test/blah.txt')
{'kind': 'storage#object',
 'id': 'gcsfs_test/blah.txt/1571708493657835',
  ...
 'type': 'file'}

# info on file after ls'ing; does hit cache
fs.ls('gs://gcsfs_test/', detail=True)
fs.info('gs://gcsfs_test/blah.txt')
{'kind': 'storage#object',
 'id': 'gcsfs_test/blah.txt/1571708493657835',
  ...
 'type': 'file'}

# info on directory; does not hit cache no matter what, is that ok?
fs.info('gs://gcsfs_test/')
{'name': 'gcsfs_test/', 'size': 0, 'type': 'directory'}
```

i'll take a look at the vcr stuff in a bit, but if you can confirm that the above seems right that'd be helpful

martindurant:
> info on directory; does not hit cache no matter what

There are two cases here: if it's a bucket, we can only know if it exists (and is an implied directory) by trying to list it, so it's OK. If it's any deeper prefix, then it should show up in a parent directory listing if it exists OR if it itself has been listed, then we also know it exists and is a directory. In the latter of those cases, though, we still need to check if there happens to be a key with the same name as the directory, which is allowed on GCS.

bnaul:
I think that makes sense at a high level, I'm out for a few days though so if either of you is interested in taking a stab feel free to push whatever to my branch :)

bnaul:
@martindurant @TomAugspurger I think I got the cache behavior matching what's in master; a couple of tests still fail locally but I think they might just be flaky ones? let me know if it looks generally ok

martindurant:
Can you post which tests are failing - they pass when run against real google?

bnaul:
The ones failing with real google on both master and my branch are `test_pickle` and `test_map_pickle`; it looks like there's also a real issue with `test_rm_recursive` due to slightly incorrect handling of the `['/a', '/a/b']` case:

> If it's any deeper prefix, then it should show up in a parent directory listing if it exists OR if it itself has been listed, then we also know it exists and is a directory. In the latter of those cases, though, we still need to check if there happens to be a key with the same name as the directory, which is allowed on GCS.

If we have listed `gs://bucket/a/` and therefore know it's a directory from the cache, but haven't listed `gs://bucket` and don't know if `a` exists, is it ok to go look that path up remotely? Doesn't that kind of defeat the purpose of checking the cache for `bucket/a/`?

martindurant:
I hate wrapping my head around these permutation! I'll have to think about that one.

bnaul:
Regardless of what the ideal state is, I think this currently matches the cache hit behavior on master; the only test that's still giving me trouble is `test_get_put_recursive`, where I guess something something subtle must have changed with the logic of checking if a directory exists inside a directory; even the old test seems to accidentally put some stuff at `'/temp_dir/temp_dir'` instead of `/temp_dir/` but for whatever reason the test still passes whereas mine doesn't... :/

TomAugspurger:
Hmm, locally your branch is passing aside from a few tests that look like stale caches. Let me try merging master.

TomAugspurger:
I notice that `gcsfs/tests/test_core.py::test_get_put_recursive[gcs://]` pass when run by itself.

There's a leftover `temp_dir` that's lingering between tests.

TomAugspurger:
OK, this is passing for me locally when I rewrite the tests... The diff is a bit messy now, but it's basically just @bnaul's changes + a couple of test setup changes

```diff
diff --git a/gcsfs/core.py b/gcsfs/core.py
index 3f02565..c65ff78 100644
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -682,6 +682,29 @@ class GCSFileSystem(fsspec.AbstractFileSystem):
         self._call('delete', 'b/' + bucket)
         self.invalidate_cache(bucket)
 
+    def info(self, path, **kwargs):
+        """File information about this path."""
+        path = self._strip_protocol(path)
+        # Check directory cache for parent dir
+        parent_path = norm_path(self._parent(path)).rstrip("/")
+        parent_cache = self._maybe_get_cached_listing(parent_path + "/")
+        if parent_cache:
+            for o in parent_cache['items']:
+                if o['name'].rstrip("/") == path:
+                    return o
+        # Check exact file path
+        out = [o for o in self.ls(path, detail=True, **kwargs)
+               if o['name'].rstrip("/") == path]
+        if out:
+            return out[0]
+        # Check parent path
+        out = [o for o in self.ls(parent_path, detail=True, **kwargs)
+               if o['name'].rstrip("/") == path]
+        if out:
+            return out[0]
+        else:
+            raise FileNotFoundError(path)
+
     @_tracemethod
     def ls(self, path, detail=False):
         """List objects under the given '/{bucket}/{prefix} path."""
diff --git a/gcsfs/tests/utils.py b/gcsfs/tests/utils.py
index 3bf4027..ae885e2 100644
--- a/gcsfs/tests/utils.py
+++ b/gcsfs/tests/utils.py
@@ -10,6 +10,7 @@ import sys
 import tempfile
 import vcr
 
+import gcsfs.utils
 from gcsfs.core import GCSFileSystem
 from gcsfs.tests.settings import (TEST_BUCKET, TEST_PROJECT, RECORD_MODE,
                                   GOOGLE_TOKEN, FAKE_GOOGLE_TOKEN, DEBUG)
@@ -194,18 +195,23 @@ def gcs_maker(populate=False):
         try:
             gcs.mkdir(TEST_BUCKET, default_acl="authenticatedread",
                       acl="publicReadWrite")
-        except:
+        except gcsfs.utils.HttpError:
             pass
+
+        # ensure we're empty.
+        gcs.rm(TEST_BUCKET, recursive=True)
+
         for k in [a, b, c, d]:
             try:
                 gcs.rm(k)
-            except:
+            except FileNotFoundError:
                 pass
         if populate:
             for flist in [files, csv_files, text_files]:
                 for fname, data in flist.items():
                     with gcs.open(TEST_BUCKET+'/'+fname, 'wb') as f:
                         f.write(data)
+        gcs.invalidate_cache()
         yield gcs
     finally:
         for f in gcs.find(TEST_BUCKET):
```

TomAugspurger:
I apparently am not able to re-record successfully. I did the first one, but the `GCSFS_RECORD_MODE=once` failed. Presumably, since the recursive `rm` in the `gcs_maker` depends on the content of the bucket...

Looking into if we vcr can ignore certain requests (in the setup & teardown).

martindurant:
> Looking into if we vcr can ignore certain requests

I'm pretty sure no - but you can catch the exceptions.

TomAugspurger:
It's kinda ugly: https://vcrpy.readthedocs.io/en/latest/advanced.html#ignore-requests

I'm trying another approach. Hopefully it's more robust.

TomAugspurger:
I actually may have misunderstood the error earlier. It's saying something about

```pytb
E               vcr.errors.CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('/Users/taugspurger/sandbox/gcsfs/gcsfs/tests/recordings/test_simple.yaml') in your current record mode ('once').
E               No match for the request (<Request (POST) https://oauth2.googleapis.com/token>) was found.
E               No similar requests, that have not been played, found.

```


which may be because I used a service account token. I'll try with an oauth2 token...

martindurant:
You may need to delete the previous run's recording (in on of the tests/recordings/test_*.yaml) each time you try to rerecord

TomAugspurger:
Pushed a (hopefully passing) re-recording. On a call with @martindurant, we discovered that some values may not have been appropriately text-replaced when saved to the recording database. 4b5588556ab9b463f0abce39596e8203c8564531 should fix that.

TomAugspurger:
All green! Thanks for the patience, and for working on this @bnaul.

PR #194, opened 2019-11-07
Pre commit config, flake8, black

TomAugspurger:


.travis.yml
```diff
@@ -22,6 +22,8 @@ install:
 
 script:
   - GCSFS_RECORD_MODE=none py.test -vv  gcsfs
+  - pip install flake8 ; flake8 dask; fi
```
martindurant: dask -> gcsfs on this line and following

martindurant: Also, what's with `fi`?

TomAugspurger:
Ah, it was in an if block in dask.

PR #196, opened 2019-11-09
re-record docs

TomAugspurger:


gcsfs/tests/test_core.py
```diff
@@ -667,15 +667,15 @@ def test_bigger_than_block_read(token_restore):
 
 @my_vcr.use_cassette(match=['all'])
 def test_current(token_restore):
-    from google.oauth2.credentials import Credentials
+    from google.auth import credentials
```
TomAugspurger, line 670: I used a service account, not oauth.

martindurant: Is there a base class that would include both possibilities?

martindurant: (or is this already it...)

TomAugspurger: This is it.

TomAugspurger: It’s a little strange that they name them all Credentials.

PR #202, opened 2019-11-13
Pass through cache_options, deprecate trim

TomAugspurger:


TomAugspurger:
`google-auth` went from 1.6.3 to 1.7.0. Trying with taht.

PR #206, opened 2019-11-25
add user_project to enable requester-pays access

rabernat:
TODO:
- [x] test
- [ ] maybe disable this option in some circumstances


rabernat:
@martindurant - this definitely works for me in our pangeo environment.

This is a fairly urgent situation for our project, as we need to act quick to change our buckets to requester pays. I would appreciate your help in speeding this along. Feel free to edit my branch directly, as you will be quicker than me at figuring out how to test, etc.

cc @jhamman

martindurant:
@TomAugspurger : this change will likely break *all* tests, because the extra parameter will cause vcr not to match the requests. A complete re-recording would be OK, which would give us the chance to add a test. For testing, supplying a non-existent userProject should cause the fetch of any data to raise.

This is an urgent money question for pangeo.

TomAugspurger:
Taking a look now. Will push commits here.

rabernat:
> this change will likely break _all_ tests

😱 yes I can see that!

rabernat:
And to be clear, when I say "urgent" I mean, would be good to resolve within a few days, rather than weeks.

I sincerely appreciate the help!

TomAugspurger:
Added a test that directly calls `GCSFileSystem._call` and verifies that `userProject` is encoded in the url we use. That's probably sufficient for testing.

Going to rerecord everything now.

TomAugspurger:
Pushed a re-record, so this should pass now. https://github.com/dask/gcsfs/pull/206/files/19c5f16e14ad8862b2de7e8705d4f61b5a9a6e14 has the relevant changes.

@rabernat not sure if it matters, but I needed to change

```python
self.user_project = user_project or self.user_project
```

to just

```python
self.user_project = user_project
```

in `GCSFileSystem.__init__`. With the original version, things were failing when replaying the tests. Not sure why though, but I think the second version is more correct. The difference would be if you do something like

```python
client1 = GCSFileSystem(user_project="my_project")
client2 = GCSFileSystem()
```

with the version you originally pushed, `client2.user_project` would take the value from GCSFileSystem, which would be set to `my_project` when `client1` was created. Probably not what we want.

gcsfs/tests/test_core.py
```diff
@@ -733,3 +734,16 @@ def test_attrs(token_restore):
         with pytest.raises(KeyError):
             gcs.getxattr(a, "foo")
         assert gcs.getxattr(a, "something") == "not"
+
+
+@my_vcr.use_cassette(match=["all"])
+def test_request_user_proejct(token_restore):
+    with gcs_maker():
+        gcs = GCSFileSystem(TEST_PROJECT, token=GOOGLE_TOKEN, user_project=TEST_PROJECT)
```
rabernat, line 747: Currently, if `user_project=None` but `project` is set,  `user_project` will be set to `project`.

It might be good to test the case where `project` and `user_project` are different.

rabernat:
> @rabernat not sure if it matters, but I needed to change
> 
> ```python
> self.user_project = user_project or self.user_project
> ```

In my PR, it was `self.user_project = user_project or self.project`. My ideas was to default to the project id.

In any case, I supposed it's fine if we have to be explicit about `user_project`.

TomAugspurger:
Ah, sorry, I misread your original commit. Will think that through.

rabernat:
FWIW, it's more important to me to get this done fast than it is to tweak the API.

TomAugspurger:
👍 will finish it off this morning. Do you know if there's any harm in including `userProject` in requests made to non-requester pays buckets? I think the fact that the tests passed when I recorded them indicates that we're fine, but wanted to double check.

rabernat:
The only possible downside I can think of is a slightly larger payload.

TomAugspurger:
> It might be good to test the case where project and user_project are different.

Agreed. @rabernat I'm going to make a `gcsfs-testing-requesterpays` bucket in the pangeo project if that's OK. Long-term we may want another location for this.

TomAugspurger:
Still working on this. Having to rewrite some testing infra to be more selective about how we rewrite things for vcr. Should be done later today.

TomAugspurger:
Well, this is a real pain. The tldlr is that we do some bulk rewriting of request / responses to strip out things like the bucket & project name from the request URI, header, and body. I *suspect* that this is causing issues when we rewrite too much for the requester pays test, but I'm honestly not sure.

And for some unknown reason, `test_ls2` is failing on this commit for me locally. Hopefully it passes on CI :/

TomAugspurger:
Latest commit restores the fallback to `user_project = user_project or project`, so we'll always include the userProject in the request's query params.

Should be good to go.

TomAugspurger:
Annd, that one failed. I really don't understand vcr. Reverting that commit.

So the tldr is that you'll need to specify `user_project="my_project"` in order to use requester pays.

martindurant:
vcrpy was my choice, sorry. I struggled with it enough myself... If only there was a mgcs or gcsalite emulator. Come on, google!

TomAugspurger:
> vcrpy was my choice, sorry.

Probably still the right choice! Though I've seriously considered investigating how much work writing our own mocking layers would be :) That's for another day though.

Going to merge this and cut a release.

PR #208, opened 2019-11-27
API: user_project falls back to project

### LINKED ISSUE #207: Enable user_project fallback to project_id
As a followup to #206, it'd be convenient to have `user_project` fall back to `project_id` if `user_project` isn't specified but `project_id` is.

Code-wise, this is quite simple

```python
self.user_project = user_project or project_id
```

Unfortunately, this didn't play well with vcr and our CI. I think we were too aggressive with our rewriting of values when saving the requests with VCR, but still need to confirm that.

cc @rabernat. Will see what I can get done this afternoon.

TomAugspurger:
Closes #207

We'll see if this passes. Getting the same `test_ls2` failure locally...

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -248,7 +248,7 @@ class GCSFileSystem(fsspec.AbstractFileSystem):
     Parameters
     ----------
     project : string
-        project_id to work under. Note that this is not the same as, but ofter
+        project_id to work under. Note that this is not the same as, but often
         very similar to, the project name.
         This is required in order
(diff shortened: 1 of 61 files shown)


TomAugspurger:
Yeah, everything is failing. This will need some more thorough investigation.

TomAugspurger:
I think https://github.com/dask/gcsfs/pull/208/commits/2e09b34a9144f58063bf6586a42b74ef65c311d0 will fix the failing tests.

PR #209, opened 2019-12-02
TST: Reduce inter-test state

TomAugspurger:
Previously, auth state was leaking between tests. This meant that we had

```python
test_simple:
    - auth request
    - test requests
test_simple_upload:
    # reuse cached GCSFileSystem
    - test requests
```

So when running just `test_core.py::test_requests`, we would make an auth request (since there's no cache) and fail.


TomAugspurger:
cc @martindurant.

Going to merge this now so that I can make progress on https://github.com/dask/gcsfs/pull/208 more easily. I suspect that if we want to cache this we can by doing it outside of the initial test (maybe in a fixture). But we can do that as a followup if needed.

PR #212, opened 2019-12-02
Refactor authentication

### LINKED ISSUE #211: Requester pays fallback fails for google_default auth
The fallback is failing because `_connect_google_default` sets `self.project`. Going to do a refactor of this to keep all the state setting in `__init__`.

TomAugspurger:
Closes #211.

This refactors all of our auth. The tldr is that we don't want to be setting state on a GCSFileSystem instance outside of `__init__`. To verify that I refactored all the `_connect` methods to not take a `self` and refactored `self.connect` to not assign anywhere.

cc @martindurant. I'm not sure if we should merge this. Running coverage, it doesn't look like our auth setup is thoroughly tested. So the tests passing here aren't a huge confidence boost. I'm going to open an alternative PR for #211 that's much simpler.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -181,6 +181,94 @@ def validate_response(r, path):
             raise RuntimeError(m)
 
 
+def _dict_to_credentials(token, scopes):
+    """
+    Convert old dict-style token.
+
+    Does not preserve access token itself, assumes refresh required.
+    """
+    try:
+        token = service_account.Credentials.from_service_account_info(
+            token, scopes=scopes
+        )
+    except:  # noqa: E722
+        # TODO: catch specific exceptions
+        token = Credentials(
+            None,
+            refresh_token=token["refresh_token"],
+            client_secret=token["client_secret"],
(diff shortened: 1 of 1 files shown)


gcsfs/core.py
```diff
@@ -181,6 +181,94 @@ def validate_response(r, path):
...
+
+def _connect_anon():
+    return requests.Session(), None
+
+
+def _connect_browser(project, access, scopes):
+    flow = InstalledAppFlow.from_client_config(client_config, scopes)
+    credentials = flow.run_console()
+    # XXX: tokens
+    # self.tokens[(project, access)] = credentials
```
martindurant: This is actually quite important: so someone can authenticate via the browser the first time, but afterwards use the same refresh token for future auths.

TomAugspurger: Oh, I think I got that working. One minute.

martindurant:
I have no plant on working on this, so let's close as stale unless there's a need to revive.

Issue #218, opened 2019-12-05
gcsfs 0.5.2 breaks on permission issue


DPGrev:
After updating from 0.5.1 to 0.5.2 we get the following error:

```
<<Project Number>>-compute@developer.gserviceaccount.com does not have serviceusage.services.use access to project <<Project Number>>.
```

This does not happen for GCFS 0.5.1.

TomAugspurger:
Sorry about that, will get a fixed release out soon.

I noticed the same error yesterday when my default Google project didn't match the project I was passing to GCSFileSystem.

Can you confirm that yours match? It's `gcloud config get-value project`

TomAugspurger:
I should mention, this happened when using `google_default` to connect (the default `token`, if successful, or explicitly with `token="google_default"`.

TomAugspurger:
@DPGrev can you check with master? If it's the same issue, the call create the GCSFileSystem will now raise when the projects don't match up.

If it's a different issue, it'd be good to diagnose that before I do a release later today.

sukhinin541:
FWIW, I get the same `does not have serviceusage.services.use access` error in 0.5.3, in particular when I try pd.read_csv(). Also, pandas_df.to_csv() just fails silently. Rolling gcsfs back to 0.5.1 resolves these issues.

TomAugspurger:
@sukhinin541 can you share more information? In particular,

1. What auth mechanism are you using?
2. Does your `gcloud config get-value project` match the project you're using with gcsfs?

@DPGrev could you do the same?

sukhinin541:
@TomAugspurger, here are my answers:

1. When I use gcsfs directly, I'm letting it determine my credentials, i.e. `fs = gcsfs.GCSFileSystem()`
2. `gcloud config get-value project` matches what `fs.project` returns

Also, when I call `read_csv()/to_csv()` Pandas methods, I'm only indirectly using gcsfs, so I'm not sure which auth mechanism is utilized in that scenario.

Let me know if you need more info.

ctivanovich:
I am having the same error, also with `gcsfs-0.5.3`. I similarly have authenticated a service account in my linux environment, and have gcsfs auto-detect the environment's config. I have confirmed I don't this issue with an early version of gcsfs.

TomAugspurger:
[...]

PR #219, opened 2019-12-05
Validate projects

TomAugspurger:
cc @DPGrev, this would have caught the issue earlier, if the cause of #218 is the same as mine.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Version 0.5.3
+=============
+
+* ``GCSFileSystem`` now validates that the ``project`` provided, if any, matches the
+  Google default project when using ``token='google_default'`` to authenticate.
+
 Version 0.5.2
 =============
 
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -2,6 +2,8 @@
 """
 Google Cloud Storage pythonic interface
 """
+import textwrap
+
 import fsspec
 
 import decorator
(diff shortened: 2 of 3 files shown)


TomAugspurger:
cc @martindurant if you have a chance to weigh in on whether

```python
GCSFileSystem(project="foo", token="google_default")
```

should raise when the google default returns a different project than what was requested.

martindurant:
A user can have access to multiple projects, so I'm not sure if we can be definitive about it, without *trying* something against the google api.

TomAugspurger:
> without trying something against the google api.

What do you mean by this? In this case, we are hitting the google API for authorization with `google_default`.

Concretely, on master we have this behavior.

```python
In [1]: !gcloud config get-value project
pangeo-181919

In [2]: import gcsfs

In [3]: fs = gcsfs.GCSFileSystem(token="google_default", project="dask-demo-182016")
/Users/taugspurger/miniconda3/envs/filesystems/lib/python3.7/site-packages/google/auth/_default.py:69: UserWarning: Your application has authenticated using end user credentials from Google Cloud SDK. We recommend that most server applications use service accounts instead. If your application continues to use end user credentials from Cloud SDK, you might receive a "quota exceeded" or "API not enabled" error. For more information about service accounts, see https://cloud.google.com/docs/authentication/
  warnings.warn(_CLOUD_SDK_CREDENTIALS_WARNING)

In [4]: fs.project
Out[4]: 'pangeo-181919'
```

I think it's surprising that `fs.project` ends up as something other than what I passed to `project`.

When project is not specified, then inferring it from the google default makes complete sense.

martindurant:
Oh, I see what you mean. Yes, we should error if we have tried setting a project, but the auth mechanism gives back another project.

TomAugspurger:
Great, thanks for confirming!

PR #224, opened 2019-12-11
Rework requester pays

### LINKED ISSUE #218: gcsfs 0.5.2 breaks on permission issue

TomAugspurger:
Closes #218


TomAugspurger:
The primary drawback here is not supporting billing against a different project than `project`, via a `user_project` keyword. Is that going to be a problem?

martindurant:
`requestor_pays` can be a string? I.e., `False` for normal operation, `True` to charge to the default project and something else to set explicitly?

TomAugspurger:
That seems reasonable.

TomAugspurger:
OK, merging this.

cc @charlesbluca. I plan to release this tomorrow. If / when you upgrade to gcsfs 0.6.0 you'll need to update the storage options to include `requester_pays=True`.

PR #226, opened 2019-12-13
Properly retry failing HTTP requests without JSON response body

aberres:
I was running into the case that a 503 returned from the storage API was not causing a retry but instead ended up as a `RuntimeError`.  
If I understood the code flow correctly the only path which made sense to me is the one where the JSON body is missing. As this can happen according to the docs this MR tries to handle this case.

I cannot validate the fix in production as luckily a 503 does not happen that often.

Personally, I would remove the special handling for the 502 case from `validate_response` as well, but I wanted to not do too many unrelated changes.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -32,7 +32,7 @@ import warnings
 import random
 
 from requests.exceptions import RequestException, ProxyError
-from .utils import HttpError, RateLimitException, is_retriable
+from .utils import HttpError, is_retriable
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


aberres:
When looking through the issues this might fix https://github.com/dask/gcsfs/issues/189

TomAugspurger:
This looks good at a glance, thanks. Will see if @martindurant has time to take a look.

martindurant:
Looks pretty good at a quick glance.

TomAugspurger:
Thanks @aberres!

JacobHayes:
I think this will fix https://github.com/dask/gcsfs/issues/221 as well. For context, the page mentioned in the commit message is here: https://cloud.google.com/storage/docs/json_api/v1/status-codes

PR #227, opened 2019-12-14
Use GCS specific batch endpoint

JacobHayes:
The global batch endpoint is deprecated, which is used for bulk `rm`. See the [deprecation notice](https://developers.googleblog.com/2018/03/discontinuing-support-for-json-rpc-and.html).

This should be covered by the [existing batch test](https://github.com/dask/gcsfs/blob/master/gcsfs/tests/test_core.py#L165).

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -975,7 +975,7 @@ class GCSFileSystem(fsspec.AbstractFileSystem):
             )
             r = self._call(
                 "POST",
-                "https://www.googleapis.com/batch",
+                "https://www.googleapis.com/batch/storage/v1",
                 headers={
                     "Content-Type": 'multipart/mixed; boundary="=========='
                     '=====7330845974216740156=="'


JacobHayes:
Looks like I need to update some VCR files. I'm guessing it broke so many tests b/c the `rm` method is also used to clean up after the tests?

TomAugspurger:
I'll push a commit re-recording.

Will using this endpoint change any user-facing behavior?

martindurant:
No, it retains the same behaviour. I had seen the depr warnings, but didn't think it was happening so soon.

TomAugspurger:
@JacobHayes I'm having trouble pushing to the branch `replicahq/gcsfs:fix-batch-deprecation`. Do you have the checkbox for "allow edits from maintainers" ticked? It could be on my end.

JacobHayes:
Thanks @TomAugspurger and yeah, that checkbox is ticked:

Screen Shot 2019-12-16 at 07 36 50

TomAugspurger:
Merged in https://github.com/dask/gcsfs/pull/229

PR #234, opened 2020-01-26
Fix gcs md5sum

### LINKED ISSUE #232: gcsfs md5sum fails
When uploading a file a little bit larger than chunk_size and using md5sum validation, the md5sum validation fails. This is because the last (partial) chunk is not added to the md5sum. 

In the last chunk (of size < chunk_size) the returned header does not contain "Range". 

`
{'X-GUploader-UploadID': '<>', 'ETag': '<>', 'Vary': 'Origin, X-Origin', 'Content-Type': 'application/json; charset=UTF-8', 'Cache-Control': 'no-cache, no-store, max-age=0, must-revalidate', 'Pragma': 'no-cache', 'Expires': 'Mon, 01 Jan 1990 00:00:00 GMT', 'Date': 'Fri, 03 Jan 2020 10:50:40 GMT', 'Content-Length': '897', 'Server': 'UploadServer',...}
`

In the code this is handled by the following line, which finalizes the upload. Forgotten is to update the md5sum with the last chunk before validating the md5sum.

```
elif l:
            #
            assert final, "Response looks like upload is over"
            size, md5 = int(r.json()["size"]), r.json()["md5Hash"]
            if self.consistency == "size":
                assert size == self.buffer.tell() + self.offset, "Size mismatch"
            if self.consistency == "md5":
                assert (
                    b64encode(self.md5.digest()) == md5.encode()
                ), "MD5 checksum failed"
        else:
            assert final, "Response looks like upload is over"
        return True
```

[Link](https://github.com/dask/gcsfs/blob/d7b832e13de6b5b0df00eeb7454c6547bf30d7b9/gcsfs/core.py#L1194)

kkwekkeboom:
Closes #232 

When uploading a file a little bit larger than chunk_size and using md5sum validation, the md5sum validation fails. This is because the last (partial) chunk is not added to the md5sum.

Also fixed consistency check on size

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1194,13 +1194,14 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
         elif l:
             #
             assert final, "Response looks like upload is over"
-            # update md5 with final chunk of data
-            if self.consistency == "md5":
-                self.md5.update(data)
             size, md5 = int(r.json()["size"]), r.json()["md5Hash"]
             if self.consistency == "size":
+                # update offset with final chunk of data
+                self.offset += l
                 assert size == self.buffer.tell() + self.offset, "Size mismatch"
             if self.consistency == "md5":
+                # update md5 with final chunk of data
+                self.md5.update(data)
                 assert (
                     b64encode(self.md5.digest()) == md5.encode()
                 ), "MD5 checksum failed"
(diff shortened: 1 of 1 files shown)


martindurant:
Can you please add a test for this "chunk and a bit" case?

PR #238, opened 2020-02-14
Fix file not being committed (#237)

### LINKED ISSUE #237: TypeError on commit
It seems that `_upload_chunk()` does not check if `self.offset` has been set.

```
    ...
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 1216, in commit
    self._upload_chunk(final=True)
  File "</usr/local/lib/python3.7/site-packages/decorator.py:decorator-gen-18>", line 2, in _upload_chunk
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 52, in _tracemethod
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 1166, in _upload_chunk
    self.offset + l - 1,
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```
I suppose this happens if the file is small enough so that the final chunk is the first one, and autocommit=False. gcsfs 0.6.0

remram44:
Patch from @martindurant (https://github.com/dask/gcsfs/issues/237#issuecomment-585994780), submitted at his request.

Partially fixes #237 (the `TypeError` should probably still be addressed).

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1168,7 +1168,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
         else:
             if l < GCS_MIN_BLOCK_SIZE:
                 if not self.autocommit:
-                    return
+                    return False
                 elif not final:
                     raise ValueError("Non-final chunk write below min size.")
             head["Content-Range"] = "bytes %i-%i/*" % (self.offset, self.offset + l - 1)


martindurant:
There should ideally also be a test, along the lines of the original issue (showing that it now works). Creating tests in gcsfs is tricky, because of the vcrpy framework - let me know if you have trouble.

martindurant:
(or I may just merge this if neither of us has time)

PR #240, opened 2020-02-29
Safely delete entry from listing cache

### LINKED ISSUE #239: Race condition in _maybe_get_cached_listing
I've mostly been keeping `cache_timeout=None` but as soon as I tried disabling the cache I ran into this pretty quickly: when running with multiple threads touching the same set of files, it seems we can try to delete `path` here twice: https://github.com/dask/gcsfs/blob/master/gcsfs/core.py#L610-L622

```
  File "/Users/brett/model/.venv/lib/python3.7/site-packages/gcsfs/core.py", line 795, in info
    parent_cache = self._maybe_get_cached_listing(parent_path + "/")
  File "<decorator-gen-4>", line 2, in _maybe_get_cached_listing
  File "/Users/brett/model/.venv/lib/python3.7/site-packages/gcsfs/core.py", line 52, in _tracemethod
    return f(self, *args, **kwargs)
  File "/Users/brett/model/.venv/lib/python3.7/site-packages/gcsfs/core.py", line 622, in _maybe_get_cached_listing
    del self._listing_cache[path]
KeyError: 'city_data/portland/transit/'
```
Would it make sense to just `try: del self._listing_cache[path]; except KeyError: pass` to prevent this?

bnaul:
Fixes #239.

@TomAugspurger I'm guessing you'd probably want to do something similar in intake/filesystem_spec#216.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -619,7 +619,7 @@ class GCSFileSystem(fsspec.AbstractFileSystem):
                     cache_age,
                     self.cache_timeout,
                 )
-                del self._listing_cache[path]
+                self._listing_cache.pop(path, None)
                 return None
 
             return listing


TomAugspurger:
Thanks, merging.

I suspect if we really want to make this threadsafe we'll have to change how we handle cached state quite a bit, right?

PR #250, opened 2020-04-13
Setting user-agent header to dask-gcsfs/<version>

vlasenkoalexey:
@sshrdp FYI


gcsfs/core.py
```diff
@@ -25,6 +25,7 @@
 import re
 import requests
 import time
+import versioneer
```
martindurant, line 28: `from .. import version`

martindurant: (might need to change order of inputs in `__init__`)

gcsfs/core.py
```diff
@@ -440,6 +442,11 @@ def _call(self, method, path, *args, **kwargs):
         data = kwargs.pop("data", None)
         r = None
 
+        if headers is None:
+            headers = {}
+        if "User-Agent" not in headers:
+            headers["User-Agent"] = "dask-gcsfs/" + VERSION
```
martindurant, line 448: I suggest `"python-gcsfs/"`, since this is not necessarily used in conjunction with Dask.

vlasenkoalexey:
Thanks for quick reply, I'll resolve your comments.
Unfortunately I wasn't able to make tests to pass, see https://github.com/dask/gcsfs/issues/251

vlasenkoalexey:
Have to rebase, will send another PR for this

Issue #251, opened 2020-04-14
Newly generated VCR files fail on replay


vlasenkoalexey:
I tried to do a small change https://github.com/dask/gcsfs/pull/250 but I can't make VCR tests to pass.
The format of newly generated replay files are significantly different from existing ones, see https://gist.github.com/vlasenkoalexey/7dbb38dfb5b2d09181b9837e6a0c40f5

I tried on MacOS, Ubuntu, and even in Docker following steps from travis config in:
docker run -it --rm --net=host -v ${PWD}:/gcsfs -w /gcsfs continuumio/miniconda3

I guess one of the following libraries was updated: google-auth google-cloud-core google-api-core google-api-python-client
And now request format is different.
If so, either old library version has to be pinned, or test setup has to be fixed.
I'm not very familiar with the protocol to fix this myself.

Issue #231, opened 2019-12-23
gcsfs can't access google storage file, switches to anon user


kapilkd13:
I am trying to read a csv file present in gs bucket directly into dataframe. 
I used `gcloud auth login` to authorize account. But when I try to use gcsfs it thorws:
`gcsfs.utils.HttpError: Anonymous caller does not have storage.objects.list access to <my-bucket>`

Now I ca list/download my files using `gsutil`. I can also list these files by passing `adc` file as token to `fs=gcsfs.GCSFileSystem(token='adc.json')`. My issue is how do I set it as default. Is there a particular location where I put this file. I couldn't find anything in docs.

kapilkd13:
Update: This works with version 0.5.0. Might be an issue with current release.

TomAugspurger:
What's the output of `gcsfs.__version__`? Can you post your code and the traceback? How are you authenticating?

kapilkd13:
gcsfs version is 0.6.0
I am working on a google colab notebook, authenticating it with `gcloud auth login` so to access gcv bucket created in my account.  I also tried uploading application credentials(adc.json) and exporting with `export GOOGLE_APPLICATION_CREDENTIALS="[PATH]"` but both failed similarly.
Here is a minimal code example,
```
import pandas
import gcsfs
```
authenticating with auth login
```
!gcloud auth login
>>> You are now logged in as [kapil.kumar@dxxxxt.in].
df=pd.read_csv("gs://<bucket file path>")
>>> /usr/local/lib/python3.6/dist-packages/gcsfs/core.py in _list_objects(self, path)
    640             return clisting
    641 
--> 642         listing = self._do_list_objects(path)
    643         retrieved_time = time.time()
    644 

</usr/local/lib/python3.6/dist-packages/decorator.py:decorator-gen-128> in _do_list_objects(self, path, max_results)

/usr/local/lib/python3.6/dist-packages/gcsfs/core.py in _tracemethod(f, self, *args, **kwargs)
     50         logger.log(logging.DEBUG - 1, tb_io.getvalue())
     51 
---> 52     return f(self, *args, **kwargs)
     53 
     54 

/usr/local/lib/python3.6/dist-packages/gcsfs/core.py in _do_list_objects(self, path, max_results)
    661             delimiter="/",
    662             prefix=prefix,
--> 663             maxResults=max_results,
    664         ).json()
    665 

[...]

PR #255, opened 2020-05-06
Fixing test infrastructure

vlasenkoalexey:
This should take care of #251 
The actual change is in gcsfs/tests/utils.py, but all replay files had to be regenerated to work with new GCS request format.


vlasenkoalexey:
Tests are passing when I run on my machine under docker, but failing in travis. Can't figure out what's wrong, will take another look tomorrow.

vlasenkoalexey:
Should also take care of #231

gcsfs/tests/test_core.py
```diff
@@ -702,12 +702,12 @@ def test_writable():
 def test_merge():
     with gcs_maker() as gcs:
         with gcs.open(a, "wb") as f:
-            f.write(b"a" * 100)
+            f.write(b"a" * 10)
```
martindurant, line 705: Why were these data sizes changed?

vlasenkoalexey: test was consistently failing with rate exceeded error when executed under default user authentication. As far as I understand, actual value doesn't matter that much.

martindurant: Indeed it doesn't matter for this test, but it is worrying nevertheless. There should be the same number of requests either way, one would think... Never mind!

PR #257, opened 2020-05-06
Setting user-agent header to python-gcsfs/<version>

vlasenkoalexey:
Replacement PR for #250

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -30,7 +30,7 @@ import random
 
 from requests.exceptions import RequestException, ProxyError
 from .utils import HttpError, is_retriable
-from ._version import get_versions
+from . import __version__ as version
 
 logger = logging.getLogger(__name__)
 
@@ -68,7 +68,6 @@ DEFAULT_PROJECT = os.environ.get("GCSFS_DEFAULT_PROJECT", "")
 
 GCS_MIN_BLOCK_SIZE = 2 ** 18
 DEFAULT_BLOCK_SIZE = 5 * 2 ** 20
-VERSION = get_versions()["version"]
 
 
 def quote_plus(s):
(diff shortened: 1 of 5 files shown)


gcsfs/core.py
```diff
@@ -30,6 +30,7 @@
 
 from requests.exceptions import RequestException, ProxyError
 from .utils import HttpError, is_retriable
+from ._version import get_versions
```
martindurant: `from . import __version__ as VERSION` and remove the `VERSION=` line below

vlasenkoalexey: I tried that, but it didn't work. As far as I understand it is causing circular dependency:

ImportError                               Traceback (most recent call last)
<ipython-input-1-c3382239b67a> in <module>
----> 1 from gcsfs.core import GCSFileSystem, quote_plus, validate_response
      2 
      3 gcs = GCSFileSystem('alekseyv-scalableai-dev', token='/Users/alekseyv/.config/gcloud/application_default_credentials.json')
      4 gcs.ls('alekseyv-scalableai-dev')  # no error

~/vlasenkoalexey/gcsfs/gcsfs/__init__.py in <module>
----> 1 from .core import GCSFileSystem
      2 from .mapping import GCSMap
      3 from ._version import get_versions
      4 
      5 __version__ = get_versions()["version"]

~/vlasenkoalexey/gcsfs/gcsfs/core.py in <module>
     32 from .utils import HttpError, is_retriable
     33 #from ._version import get_versions
---> 34 from . import __version__ as VERSION
     35 
     36 logger = logging.getLogger(__name__)

ImportError: cannot import name '__version__'

martindurant: Ah indeed, the order in init needs changing too:
```diff
--- a/gcsfs/__init__.py
+++ b/gcsfs/__init__.py
@@ -1,8 +1,8 @@
-from .core import GCSFileSystem
-from .mapping import GCSMap
 from ._version import get_versions

 __version__ = get_versions()["version"]
 del get_versions
+from .core import GCSFileSystem
+from .mapping import GCSMap
```

vlasenkoalexey: this worked, thanks for suggestion

PR #263, opened 2020-06-03
add _connect_cloud to the list or tried default connect options

dynamix:
So the documentation is in sync with the behaviour.


martindurant:
OK!
I am surprised this wasn't correct...

max-sixty:
FWIW, when I've run on `cloud` but not been on cloud, it's blocked (and not raised an error in a reasonable length of time). Maybe that was a one-off / because of my env though.

PR #271, opened 2020-07-22
Retry after encountering a checksum failure in GCSFileSystem.cat

### LINKED ISSUE #267: Retry after encountering a checksum failure in GCSFileSystem.cat?
We've been using gcsfs fairly heavily for loading data from zarr stores in the cloud; in general it has worked really well.  A basic pattern might look something like this:

```python
import gcsfs
import xarray

PROJECT = "example-project"
STORE = "gs://path-to-store/store.zarr"

gcs = gcsfs.GCSFileSystem(project=PROJECT)
mapper = gcs.get_mapper(STORE)
ds = xarray.open_zarr(mapper)
result = ds.mean().compute()
```

Every now and then -- I honestly don't know what dictates it -- there are days when we encounter transient errors like this when performing the computation:

```
AssertionError                            Traceback (most recent call last)

... (very long traceback)

~/miniconda3/envs/fv3net/lib/python3.7/site-packages/gcsfs/core.py in cat(self, path)
    888                 if key == "md5":
    889                     md = b64decode(val)
--> 890                     assert md5(r.content).digest() == md, "Checksum failure"
    891         return r.content
    892 
AssertionError: Checksum failure
```

Typically one can retry the script a few times and eventually the full computation succeeds.  This can be a bit frustrating, however, because particularly for long-running computations, many downloads will be performed before one of these errors is hit, and re-running the script means starting over from scratch.  Would it be appropriate to add some logic to `GCSFileSystem.cat` that would retry a download if a checksum failure occurred?  I believe there are other places in gcsfs where retry logic is used to work around these kinds of transient issues.

Apologies for not being able to provide a minimal working example -- the transient nature of the issue makes that difficult.  

cc: @nbren12 @oliverwm1 @brianhenn

spencerkclark:
This is a basic attempt to fix #267.  I moved the logic that compares the md5 hash of the content to the value in the header into `validate_response`, and created a retry-able exception that gets thrown if there is a mismatch.  

@martindurant I didn't go quite as far as to factor the retry logic out of `_call` -- I'm pretty new to this codebase -- but this approach seemed simplest to me.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1181,9 +1181,6 @@ class GCSFileSystem(AsyncFileSystem):
         r: requests response object
         path: associated URL path, for error messages
         """
-        if headers is None:
-            headers = {}
-
         if status >= 400:
             error = None
             try:
@@ -1208,7 +1205,7 @@ class GCSFileSystem(AsyncFileSystem):
             else:
                 raise RuntimeError(msg)
         else:
-            if "X-Goog-Hash" in headers:
+            if headers is not None and "X-Goog-Hash" in headers:
                 # if header includes md5 hash, check that data matches
                 bits = headers["X-Goog-Hash"].split(",")
                 for bit in bits:
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1172,7 +1172,7 @@ class GCSFileSystem(AsyncFileSystem):
         else:
             return path.split("/", 1)
 
-    def validate_response(self, status, content, json, path, headers):
+    def validate_response(self, status, content, json, path, headers=None):
         """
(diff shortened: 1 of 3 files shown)


gcsfs/core.py
```diff
@@ -1181,7 +1172,7 @@ def split_path(cls, path):
         else:
             return path.split("/", 1)
 
-    def validate_response(self, status, content, json, path):
+    def validate_response(self, status, content, json, path, headers):
```
martindurant: Perhaps have headers=None, so that you don't need to change the call signature elsewhere

spencerkclark: Thanks @martindurant; should be all set now.

martindurant:
Perfect

PR #277, opened 2020-08-14
add sign to filesystem class

CJ-Wright:
This will enable downstream classes to specify how to sign urls, providing short term URLs for situations where the URL is needed but permissioning needs to be tightly controlled.

Note that if we don't want to add a `google.cloud` dep there is a recipe for how to do this with just `google.oauth2` which is under an Apache-2.0 License.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1215,6 +1215,16 @@ class GCSFileSystem(AsyncFileSystem):
                         if md5(content).digest() != md:
                             raise ChecksumError("Checksum failure")
 
+    def sign(self, path, expiration=0):
+        # if we don't want to use google.cloud we can use
+        # https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/signed_urls/generate_signed_urls.py
+        from google.cloud import storage
+        bucket, key = self.split_path(path)
+        client = storage.Client()
+        bucket = client.bucket(bucket)
+        blob = bucket.blob(key)
+        return blob.generate_signed_url(expiration=expiration, method='GET')
+
 
 GCSFileSystem.load_tokens()
 


martindurant:
Why, what is the license for google.cloud?

CJ-Wright:
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/signed_urls/generate_signed_urls.py

martindurant:
Oh, I see. I think we can do without the dep, and reference the origin of the code in the method comments, and mention the implied license. 

Note that the test is failing on style, please run `black`. It would be nice to have a test, which I suppose would involve requests.get(url), to see it has the same content. vcrpy should cope with recording that fine.

CJ-Wright:
@martindurant do you have a test that I should use as a template for this?

gcsfs/core.py
```diff
@@ -1215,6 +1214,14 @@ def validate_response(self, status, content, json, path, headers=None):
                         if md5(content).digest() != md:
                             raise ChecksumError("Checksum failure")
 
+    def sign(self, path, expiration=100, **kwargs):
+        # if we don't want to use google.cloud we can use
+        bucket, object_name = self.split_path(path)
+        self.connect()
```
martindurant, line 1220: Why would you need this? Connect should have happened at instantiation.

gcsfs/utils.py
```diff
@@ -74,3 +80,104 @@ async def send(self, pre, f, post):
...
+        )
+
+    escaped_object_name = quote(object_name.encode(), safe=b"/~")
+    canonical_uri = "/{}".format(escaped_object_name)
+
+    datetime_now = datetime.datetime.utcnow()
+    request_timestamp = datetime_now.strftime("%Y%m%dT%H%M%SZ")
+    datestamp = datetime_now.strftime("%Y%m%d")
+
+    client_email = google_credentials.service_account_email
```
martindurant, line 111: What happens if this isn't a service account?

martindurant (commented):
Sorry this got lost - it's a useful addition.

martindurant:
@CJ-Wright : the question about service accounts I think needs to be answered before I know what to do with this PR. The answer "this only works for service accounts" might be OK, but not great - but since I don't really understand how the signing works, I don't know what the other options might be.

baszalmstra:
There hasnt been any updates on this for a while. I can take over this PR but I dont really understand the whole service account thing. Would it be ok to raise an exception if there is no service account for now?

PR #294, opened 2020-10-08
log connection exception

dynamix:
The reason why a connection method failed was not logged before, this PR adds the actual exception to the log message.


martindurant:
Please apply `black` to pass linting - otherwise happy to include this.

martindurant:
Ping... 
I would apply the linting myself, but I cannot write to your branch. If I don't hear otherwise, I may make this into a separate PR from my fork.

PR #297, opened 2020-10-23
Override _strip_protocol method

namanjain:
This change overrides the internal `_strip_protocol` method for the fs.
The current implementation uses the `AbstractFileSystem` method which
doesn't return the correct result for GCS urls. Instead, we override
it with `infer_storage_options` method that other remote filesystem
implementation uses.

Fixing this also fixes some other bugs that rely on this method. One of
such bugs was encountered when uploading a file to a folder in GCS and
the implementation before the change infered the remote folder as the
filename. I added a test to check for this bug and it works with the
fix.

This change fixes https://github.com/intake/filesystem_spec/issues/448.


gcsfs/core.py
```diff
@@ -270,6 +271,10 @@ def __init__(
             self._session = None
         self.connect(method=token)
 
+    @classmethod
+    def _strip_protocol(cls, path):
+        return infer_storage_options(path)["path"]
```
martindurant: Seems a little heavy handed. Why not:
```
for protocol in self.protocols:
    if path.startswith(protocol + "://"):
        path = path[len(protocol) + 3:]
```

martindurant (commented):
This PR is failing on style - please run `black`.

Also, I would love a test case that showed the original problem, which is now fixed. In what case does that final "/" matter?

martindurant:
EDIT - maybe test_get_put_file_in_dir does exactly what I asked for, I'll have to read carefully.

gcsfs/core.py
namanjain: The path can also start with `gcs://` or `gs://`. Like I explained in the description, this is how most of the other implementations are handling it as well. I can also change it to what you suggested. Your call.

martindurant: Right, and self.protocols contains both "gs" and "gcs". This is the sort of logic in the superclass, and what almost all of the other implementations do. The only difference you are after, is the remove the `rstrip("/")`, right?

Also, you'll notice that `AbstractFileSystem._strip_protocol` handles lists of paths and non-string paths, which we would lose here; that doesn't appear to break anything, but consistency is good. I wonder, would it be easier to add an option to that method, whether to remove the trailing "/"?

namanjain: Oh I see how the base class method does it. You're right, we just need to remote the `rstrip("/")` call for gcsfs.

Good point on the non-string path handling. I can also convert the path object to string first before sending it to `infer_storage_options` method. I think both the options are pretty doable at this point.

I'm still leaning towards the approach in my PR instead of adding an option to the base method, but that's mostly because adding an option to the method would take longer. That's because I'll have to make changes to fsspec, wait for the new version to release, and then make a change in gcsfs. But I can be easily convinced otherwise (you can even veto it since you are a maintainer for both repos).

martindurant:
I think I prefer using as close to the upstream implementation as possible, so please just copy it, minus the line you don't want.
Please remove the .vscode/settings.json file.

martindurant:
Thanks for implementing, this is going in.

For reference: github does not send notifications when you force-push, so you should ping when you are ready for a new review.

PR #303, opened 2020-11-10
Support ls, info, and exists without storage.objects.get permission

JacobHayes:
Closes https://github.com/dask/gcsfs/issues/245

I'm having a little more trouble with VCR and test access on this one. Aside from that, I'm not sure how to put in a good regression test. We do a little IAM manipulation with the `gcs_maker`, but a separate bucket/user without `storage.objects.get` may seem a little heavy handed. Perhaps I can use a separate VCR cassette inspect the requests and assert no object GETs?

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -589,11 +589,22 @@ class GCSFileSystem(AsyncFileSystem):
             # listing.
             raise FileNotFoundError(path)
 
-        result = self._process_object(
-            bucket, await self._call("GET", "b/{}/o/{}", bucket, key, json_out=True)
-        )
-
-        return result
+        res = None
+        # Work around various permission settings. Prefer an object get (storage.objects.get), but
+        # fall back to a bucket list + filter to object name (storage.objects.list).
+        try:
+            res = await self._call("GET", "b/{}/o/{}", bucket, key, json_out=True)
+        except OSError as e:
+            if not str(e).startswith("Forbidden"):
+                raise
+            resp = await self._call("GET", "b/{}/o/", bucket, prefix=key, json_out=True)
+            for item in resp.get("items", []):
+                if item["name"] == key:
+                    res = item
+                    break
+            if res is None:
+                raise FileNotFoundError(path)
+        return self._process_object(bucket, res)
 
     async def _list_objects(self, path, prefix=""):
         bucket, key = self.split_path(path)
@@ -799,7 +810,8 @@ class GCSFileSystem(AsyncFileSystem):
             return await self._get_object(path)
         except FileNotFoundError:
             pass
-        out = await self._ls(path, detail=True, **kwargs)
+        kwargs["detail"] = True  # Force to true for info
+        out = await self._ls(path, **kwargs)
         out0 = [o for o in out if o["name"].rstrip("/") == path]
         if out0:
             # exact hit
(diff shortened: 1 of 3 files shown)


gcsfs/core.py
```diff
@@ -798,7 +804,8 @@ async def _info(self, path, **kwargs):
             return await self._get_object(path)
         except FileNotFoundError:
             pass
-        out = await self._ls(path, detail=True, **kwargs)
+        kwargs['detail'] = True # Force to true for info
+        out = await self._ls(path, **kwargs)
```
JacobHayes: I was getting `got multiple values for keyword argument 'detail'` after making the `_get_object` change. Perhaps that's a sign of some other unexpected control flow change?

gcsfs/core.py
```diff
@@ -589,11 +589,17 @@ async def _get_object(self, path):
...
             raise FileNotFoundError(path)
 
-        result = self._process_object(
-            bucket, await self._call("GET", "b/{}/o/{}", bucket, key, json_out=True)
-        )
-
-        return result
+        res = None
+        # Using a list with a prefix only requires storage.objects.list permissions, rather than
+        # storage.objects.get.
```
martindurant: I've actually come across the reverse too, where you have get permission, but not list permission

JacobHayes: I wondered about that. I don't love the complexity, but what do you think about:
- try get (code in `master`)
- catch the OSError (which is only raised for Forbidden)
- fallback to this prefix method
- if both fail/no match, make sure to return the right error (`Forbidden` if `get` raised that originally, otherwise `FileNotFound`)
?

martindurant: Yes, that sounds reasonable. I don't know if it's testable :|

JacobHayes:
Working on an additional test - planning to mock out `validate_response` to raise a variety of exceptions depending on the path (to distinguish between get vs list).

gcsfs/tests/test_core.py
```diff
@@ -929,3 +929,66 @@ def test_validate_response():
...
+        ),
+    ],
+    ids=[
+        "missing_with_get_perms",
+        "missing_with_list_perms",
+        "existing_with_get_perms",
+        "existing_with_list_perms",
+    ],
+)
+def test_metadata_read_permissions(
```
JacobHayes, line 970: Don't love how roundabout/messy this test is, so open to other ideas. 😁

martindurant: Don't worry about messy tests! It's when the main code is messy that we are in trouble.

gcsfs/core.py
```diff
@@ -589,11 +589,22 @@ async def _get_object(self, path):
...
-        return result
+        res = None
+        # Work around various permission settings. Prefer an object get (storage.objects.get), but
+        # fall back to a bucket list + filter to object name (storage.objects.list).
+        try:
+            res = await self._call("GET", "b/{}/o/{}", bucket, key, json_out=True)
+        except OSError as e:
+            if not str(e).startswith("Forbidden"):
+                raise
+            resp = await self._call("GET", "b/{}/o/", bucket, prefix=key, json_out=True)
```
martindurant: Should we have limit=1 here (I think GCS allows this) - the correct entry should always be first alphabetically.

Issue #310, opened 2020-11-26
Got HttpError, however status code is not printed in log.


onevirus:
This is error log what I got
```
_call non-retriable exception: 
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/gcsfs/core.py", line 507, in _call
    self.validate_response(status, contents, json, path, headers)
  File "/opt/conda/lib/python3.6/site-packages/gcsfs/core.py", line 1230, in validate_response
    raise HttpError({"code": status})
gcsfs.utils.HttpError
```

**What you expected to happen**:
HttpError status code in log.

**Anything else we need to know?**:
I use gcsfs via papermill.
Is there any configuration to print http status code in log ?

**Environment**:

- Dask version: 
- Python version: 3.6
- Operating System: debian
- Install method (conda, pip, source): conda

martindurant:
The argument to HttpError should also include a "message" to be printed - but the repr of the exception could be changed to include all of the arguments passed to it. Would you like to fix this?

PR #311, opened 2020-11-26
Add status code in HttpError exception log

onevirus:
This should take care of #310 

Sadly, gcs does not include message in many responses.
So, I add status code at the end of exception log.

@martindurant 
I did only _call non-retriable exception case.
Let me know if you want retriable exception case also.

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -946,14 +946,14 @@ def test_validate_response():
     with pytest.raises(HttpError) as e:
         gcs.validate_response(503, b"", None, "/path")
     assert e.value.code == 503
-    assert e.value.message == ""
+    assert e.value.message == ", 503"
 
     # HttpError with JSON body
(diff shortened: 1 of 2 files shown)


martindurant:
Yes, I would vote to log both the code and message, if we have them, in every case.

onevirus:
@martindurant 
I added error code on every HttpError logs.

martindurant:
How about the following, which makes for simpler code, and ensures the code and the message is included every time?
```diff
--- a/gcsfs/utils.py
+++ b/gcsfs/utils.py
@@ -7,8 +7,10 @@ class HttpError(Exception):

     def __init__(self, error_response=None):
         if error_response:
-            self.message = error_response.get("message", "")
             self.code = error_response.get("code", None)
+            self.message = error_response.get("message", "")
+            if self.code:
+                self.message += ", %s" % self.code
         else:
             self.message = ""
```

martindurant:
Perfect!

PR #313, opened 2020-11-30
Improve folder detection in _info

dynamix:
A suggestion to fix https://github.com/dask/gcsfs/issues/312.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -814,10 +814,10 @@ class GCSFileSystem(AsyncFileSystem):
             pass
         kwargs["detail"] = True  # Force to true for info
         out = await self._ls(path, **kwargs)
-        out0 = [o for o in out if o["name"].rstrip("/") == path]
-        if out0:
+        out0 = next((o for o in out if o["name"].rstrip("/") == path), None)
+        if out0 and out0["name"][-1] != "/" and out0["size"] == 0:
             # exact hit
-            return out0[0]
+            return out0
         elif out:
             # other stuff - must be a directory
             return {


martindurant:
What happens if you both have the empty placeholder *and* nested directories? 

This should be tested for the cases that:
- nothing has been listed yet
- that the directory in question has been listed
- that the parent directory has been listed.

With this change, is is still true that `info(path) == ls(path)[0]` in all cases?

dynamix:
You mean in case there is `/foo` and  `/foo/`? Or `/foo/` and `/foo/bar`?

I'll add some test cases at the end of the week, can't answer these out of my head. 

At the moment (with or without this PR) the behaviour seems to be incorrect as `ls` is returning `["foo", "foo"]`.

martindurant:
> You mean in case there is /foo and /foo/? Or /foo/ and /foo/bar?

I suppose both those problematic cases :|

> ls is returning ["foo", "foo"]

If you do with `detail=True`, then one of these is presumably the folder, the other the file. So it's not exactly wrong. Whether or not there should be a "/" suffix is another question.

kaushikb11:
@dynamix @martindurant Any updates on the PR? The inconsistency is causing issues for us while working on remote filesystems. [Ref](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loggers/tensorboard.py#L272)

kaushikb11 (approved):

martindurant:
I would like some additional tests for each of the cases I listed above. It's not obvious to me from the code alone that every case will pass.

gcsfs/core.py
```diff
@@ -814,10 +814,10 @@ async def _info(self, path, **kwargs):
             pass
         kwargs["detail"] = True  # Force to true for info
         out = await self._ls(path, **kwargs)
-        out0 = [o for o in out if o["name"].rstrip("/") == path]
-        if out0:
+        out0 = next((o for o in out if o["name"].rstrip("/") == path), None)
+        if out0 and out0["name"][-1] != "/" and out0["size"] == 0:
```
martindurant, line 818:
```suggestion
        if out0 and (out0["name"][-1] != "/" or out0["size"] == 0):
```

PR #321, opened 2020-12-29
Fix Invalid Dircache with Placeholder Dirs

JacobHayes:
I've come across a few buckets that have "placeholder" directory objects (objects ending in `/`, for example created with "Create Folder" in the Storage Console). After these objects get cached in the dircache, subfolders can't be retrieved. Seems there could be other manifestations, but I noticed it after `glob`bing a subdir with a matching placeholder dir object.

`git bisect` traced this back to https://github.com/dask/gcsfs/pull/289, which just seems like a cache entrypoint (so presumably there are other paths that'd populate the cache). I added a small test for that case and a very narrow fix in `find` (avoid caching those objects, but still return them). It feels like the root cause is a bit off and/or the real fix should be lower level - perhaps this is due to partial responses with `prefix` requests that _shouldn't_ be cached for example?

I'll have spotty availability until February, so this may be more an issue + test case if others want to take over.

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -1045,3 +1045,24 @@ def test_metadata_read_permissions(
             with pytest.raises(expected_error):
                 gcs.info(TEST_BUCKET + file_path)
             assert gcs.exists(TEST_BUCKET + file_path) is False
+
+
+@my_vcr.use_cassette(match=["all"])
+def test_ls_prefix_cache():
+    with gcs_maker(True) as gcs:
+        gcs.touch(f"gs://{TEST_BUCKET}/a/file1")
+        gcs.touch(f"gs://{TEST_BUCKET}/a/file2")
+
+        gcs.ls(f"gs://{TEST_BUCKET}/", prefix="a/file")
+        gcs.info(f"gs://{TEST_BUCKET}/a/file1")
+
+
+@my_vcr.use_cassette(match=["all"])
+def test_placeholder_dir_cache_validity():
+    with gcs_maker(True) as gcs:
+        gcs.touch(f"gs://{TEST_BUCKET}/a/")
+        gcs.touch(f"gs://{TEST_BUCKET}/a/file")
+        gcs.touch(f"gs://{TEST_BUCKET}/b")
+
+        gcs.find(f"gs://{TEST_BUCKET}/a/")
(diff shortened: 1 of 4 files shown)


JacobHayes:
> perhaps this is due to partial responses with prefix requests that shouldn't be cached

Ok, I think this is indeed the case - I added another test case for this. The fix for this seems a bit trickier though (I'm not sure each function knows when a response may be partial).

martindurant:
List responses may be paginated, but they should not be truncated ( https://cloud.google.com/storage/docs/json_api/v1/objects/list#response ). Do you mean that *if* prefix is given for a list/find, we should not cache, but somewhere we are doing? That seems a separate issue from paths ending in "/" (which I really wish had never become a thing).

gcsfs/tests/test_core.py
```diff
@@ -1053,7 +1053,10 @@ def test_placeholder_dir_cache_validity():
         gcs.touch(f"gs://{TEST_BUCKET}/a/file")
         gcs.touch(f"gs://{TEST_BUCKET}/b/file")
 
-        # Confirm the lookup works before the cache is polluted
-        gcs.info(f"gs://{TEST_BUCKET}/b/file")
+        gcs.ls(f"gs://{TEST_BUCKET}/", prefix="a/")
```
martindurant: The purpose of prefix in `ls` is to subselect paths within a listing that is *not* a directory level, i.e., 
`prefix="path/directory/file"` to select "path/directory/file*" and similar without listing the whole of "directory". Perhaps this is the problem? I don't think that a `prefix` value ending in "/" makes sense.

JacobHayes: Yeah the trailing `/` case may be a bit degenerate, but I updated it to use `a/file` to match `file1` and `file2` and the same issue arises.

JacobHayes:
> Do you mean that if prefix is given for a list/find, we should not cache, but somewhere we are doing

Yes, this was the case for `ls`.

> That seems a separate issue from paths ending in "/" (which I really wish had never become a thing).

You're right, the similar outcome (polluted cache/`FileNotFound`) explains why the root cause felt cloudy. 😁 `glob` ends up calling `find` with `/bucket/a/` (no prefix) and the additional caching in `find` caches the `/bucket/a/` _object_ under the `/bucket/` key.

martindurant:
Excellent, thank you

PR #322, opened 2020-12-30
Fix zero cache timeout with intake/filesystem_spec#513

JacobHayes:
After https://github.com/intake/filesystem_spec/pull/513, `cache_timeout=0` now "works" as [documented](https://github.com/dask/gcsfs/blob/b94b9c187129b144e46d5f78ef37e21e39746be2/gcsfs/core.py#L202-L204) (ie: `cache_timeout <= 0` will clear the cache immediately, rather than being equivalent to `None`/no expiration). This requires cache writes to be "atomic" (ie: can't add empty list and then append). I added a small and kinda measly test case that is effectively a noop for existing fsspec versions but would kick in with later ones - let me know if you have any better suggestions for that.


martindurant:
Sorry, this conflicts because the other PR was squash-merged. Can you please fix?

PR #326, opened 2021-01-13
Add callback_timeout argument for sync function

ChanYiLin:
Hi team,

We have encountered an issue that the `get_client` function in the `init` function when useing non-async mode might get too long (> 1 hour). It probably get stuck. 
We would like to set the callback_timeout for the `sync` function and found out there is actually no chance to give the argument since it doesn't pass the parameter during calling the function. The init function also doesn't accept callback_timeout as argument.

This PR add callback_timeout as arguments in the `init` function and pass to the `sync` function.

Please help to review, thanks!

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -237,6 +237,7 @@ class GCSFileSystem(AsyncFileSystem):
         requester_pays=False,
         asynchronous=False,
         loop=None,
+        callback_timeout=None,
         **kwargs,
     ):
         super().__init__(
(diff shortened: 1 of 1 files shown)


martindurant:
Whilst I agree in principle with this, I'd really like to find out why things are stalling, especially in `init` - I haven't heard of this elsewhere. How are you calling gcsfs, and what auth are you using (token file, cluster metadata...) ?

ChanYiLin:
We use the credentials.json and set the env `GOOGLE_APPLICATION_CREDENTIALS` and then call `gcsfs.FileSystem()` directly.
It always works when the scenario is that simple.
However, we try to develop a more complicated usecase to handle heavy IO workload by using multiprocess plus asyncio. First we spawn multiprocess(concurrent.executor) and create new event-loop for every process, then we call `fs=gcsfs.GCSFileSystem(asynchronous=True, loop=new_loop)` in the process to do async IO task. 
We do this whole process for multiple times. In this case, stalling might happen when the GCSFileSystem is initiating in the process we forked.
 
We are still investigating the reason it gets stuck. We know there is a cache for the GCSFileSystem instance and will be cleared in the new process which might be a possible root cause. That's why we would like to add timeout to see is it really the connection timeout or something happens to thread and event-loop.

Thanks!

PR #329, opened 2021-01-14
Add `credentials.apply` in `_connect_token`

### LINKED ISSUE #328: Credentials token is not applied to self.heads if valid
Here is an MVP:

```python
storage_client = storage.Client()  # (assuming default auth is correctly setup and not anonymous)
bucket = storage_client.get_bucket('mybucket')
gcsfs = GCSFileSystem(token=storage_client._credentials) 
# `_connect_token` in `__init__` will not apply the token
print(gcsfs.credentials.token)  # 'ya29.c....'
print(gcsfs.heads)  # {} but should be {'authorization': 'Bearer ya29.c....'}
# the following fails with:
# gcsfs.utils.HttpError: Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object
gcsfs.open('gs://mybucket/myprivatefile')
```

This is fixed by calling `apply`:
```python
gcsfs.credentials.apply(gcsfs.heads)
print(gcsfs.heads)  # {'authorization': 'Bearer ya29.c....'}
gcsfs.open('gs://mybucket/myprivatefile')  # works
```

I believe this could be fixed by adding 
```python
credentials.apply(self.heads)
```
after https://github.com/dask/gcsfs/blob/0.7.1/gcsfs/core.py#L368
but I may be missing edge cases.

econtal:
Fixes #328

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -388,6 +388,8 @@ class GCSFileSystem(AsyncFileSystem):
         else:
             raise ValueError("Token format not understood")
         self.credentials = credentials
+        if self.credentials.valid:
+            self.credentials.apply(self.heads)
 
     def maybe_refresh(self):
         # this uses requests and is blocking


martindurant:
Aha, understood. The assumption was, that the token would always need refreshing on first use, so `apply()` would happen then.

What do you need to finish off this PR?

econtal:
> What do you need to finish off this PR?

Hi @martindurant,
I didn't add any tests, or even looked at what is already tested. Before removing "Draft" I would like to *at least* test locally, and better add new unit tests

martindurant:
I'm not sure it is easy to test this n CI. Maybe make a dummy token out of your existing structure, with (presumably) a fake expiry time in the future, and ensure that instantiating GCSFileSystem ends up with the fake token's values populated in the `heads` attribute?
Honestly, if you say this is working for you, that would be enough.

econtal:
I tested locally and confirms this fixes #329.

`gcsfs.heads` is correctly `{'authorization': 'Bearer ya29.c....'}` and I can read files

martindurant:
Excellent

PR #330, opened 2021-01-14
Fix badges and improve .gitignore

DPGrev:
- The badges were broken on the readme page.
- Added a more complete .gitignore so the user experience of contributing should improve.


martindurant:
Thanks for the improvement

Issue #325, opened 2021-01-11
Unable to upload files larger >= 2GB


agrinh:
I'm experiencing an issue when trying to upload files larger than 2GB. 

**What happened**: I get an overflow error when uploading large files

**What you expected to happen**: Upload to finish successfully

**Minimal Complete Verifiable Example**:

```python
import gcsfs

fs = gcsfs.GCSFileSystem()

# Works fine
with fs.open("gs://bucket/path/to/upload/one", "wb") as f:
    f.write(b"0" * (2 ** 30))

# Raises `ClientOSError: [Errno 1] malloc failure (_ssl.c:2423)`
with fs.open("gs://bucket/path/to/upload/two", "wb") as f:
    f.write(b"0" * (2 ** 31 - 1))

# Raises `OverflowError: string longer than 2147483647 bytes`
with fs.open("gs://bucket/path/to/upload/three", "wb") as f:
    f.write(b"0" * (2 ** 31))
```

**Anything else we need to know?**: I've tried this on various machines and see the same thing each time, definitely not mac only.

**Environment**:

- Python version: 3.7.9
- Operating System: macOS Catalina 10.15.7
- Install method (conda, pip, source): pip
- GCSFS version: 0.7.1

martindurant:
Apparently this is a known issue with aiohttp. You could work around it by splitting your writes into multiple pieces of <2**30bytes.

Our code only currently checks for blocks that are too small, and it seems no one has tried to push a block this big before. The code could be changed to split such big blocks internally to save you the hassle.

agrinh:
> Apparently this is a known issue with aiohttp. You could work around it by splitting your writes into multiple pieces of <2**30bytes.
> 
> Our code only currently checks for blocks that are too small, and it seems no one has tried to push a block this big before. The code could be changed to split such big blocks internally to save you the hassle.

Oh, that's a bit of a surprise. Can you share a link to the known aiohttp issue, please?

Also, is there any chance you could provide some pointers for fixing this? I might be interested in helping out.

martindurant:
Actually, it seems to be deeper, in ssl or httplib, see discussions like https://github.com/psf/requests/issues/2717 (for the requests library).

So there are two possibilities, the first of which might not work, but is far simpler
[...]

PR #334, opened 2021-01-22
Allow large uploads by limiting the chunk size

agrinh:
An intended fix for #325 

Instead of the suggestion from #325 of creating a loop in `_upload_chunk` I use the existing method for handling shortfall. Let me know whether you think this is a good approach.

Seems to work when I'm testing this manually, but the added test locks on verifying the contents (`gcsfs.cat`). I haven't figured out why, but if you have any ideas I'd appreciate it.


martindurant:
This PR is failing because the new test function doesn't not have a vcrpy recording - see the project documentation on how to create one. I think, though, that for the sake of the test you should temporarily decrease the maximum upload size by a lot, since all that data gets recorded into the repo - indeed, what is reasonable for a single HTTP operation?

martindurant:
@agrinh , I have had a go at reworking this. Please see if it solves your problem! Note that there were errors in your version, around whether the large upload was the last write for the file or not. I think everything is fixed, but I have strange connection problems to GCS right now, so I wasn't able to record a run for the test function, which is skipped for now.

agrinh:
@martindurant awesome, thanks! Sorry for the late reply, but I'll have a look this weekend.

agrinh:
@martindurant I think I got it working now, simplified some logic and made sure to clear the buffer and update offset when done, but I'm seeing something really weird. If the upper limit is set to a power of 2, the test works fine, but whenever it is set to something else (like the limit you chose, `10 * 2 ** 20`), the test freezes on the second PUT (with or without my changes). I really can't figure this out, would you have time to take a look? For now, I changed the the upper limit in the test to a power of 2.

Finally, all other tests seem OK except for:
```
ERROR    gcsfs:core.py:555 _call non-retriable exception: You already own this bucket. Please select another name., 409
  File "~/proj/gcsfs/gcsfs/core.py", line 538, in _call
    self.validate_response(status, contents, json, path, headers)
  File "~/proj/gcsfs/gcsfs/core.py", line 1326, in validate_response
    raise HttpError(error)
gcsfs.utils.HttpError: You already own this bucket. Please select another name., 409
```

How should I handle this?

martindurant:
The hang and power-of-two is pretty odd, I'll try to look. 
The error about owning the bucket is fine and can be ignored, no tests should fail because of it.
However, several tests here do fail via VCR, which means that the calls made by the code must have changed, which was not supposed to happen for files smaller than the limit.

agrinh:
@martindurant Feels like something is off with my test environment. When I try running setting the vcr record mode to `"once"` or `"none"` locally after `"all"`, the tests fail from `test_simple`. I tried reverting all recordings and skipping the large upload test that seems to have made the tests happy in the CI (doesn't work locally). Any chance that you could have another try at running the tests, I'm finding it really difficult to debug.

martindurant:
It was not your build system, but timeouts caused by vcrpy attempting to work with large(ish) recordings, and causing some internal timeout. Hoping this will work.

agrinh:
@martindurant great, thanks!

kconvey:
@martindurant Any idea when this might be included in a release?

martindurant:
Soon... I am working through some stuff in fsspec around event loops that might affect gcsfs, so after that is done.

PR #343, opened 2021-02-09
Implement crc32c consistency checks

### LINKED ISSUE #342: Check crc32c checksums
GCSFS will sometimes fail to download data when `consistency="md5"` (see https://github.com/VulcanClimateModeling/fv3config/issues/124).

According to google cloud storage's [docs](https://cloud.google.com/storage/docs/composite-objects#metadata) not all GCS objects have md5 metadata. They recommend using "crc32c" instead since apparently all objects have it.  

After toying around with the python library `crcmod`, this is how one computes the "crc32c" of a bunch of bytes:
```
In [41]: crc32c = crcmod.Crc(0x11EDC6F41, initCrc=0, xorOut=0xFFFFFFFF)

In [42]: crc32c.update(data)

In [43]: crc32c.hexdigest()
Out[43]: '4C349160'
```
I got the random looking hexadecimal numbers from wikipedia. I verified that this checksum works using the google-cloud-storage library. The usage looks pretty similar to "md5". I'm happy to contribute a PR adding a `consistency="crc32c"` option if ya'll are open to it.

nbren12:
Unlike md5, crc32c check-sums are available for every object in google cloud storage including composite objects. This PR adds support for this type of checksum.

Most of the changes refactor all the consistency checking logic to a new file `checkers.py` and adds some unit tests.

Resolves #342


martindurant:
Please ping me when this is ready

nbren12:
@martindurant This should be ready. 

It would be nice to test it with the round-tripped `test_put_get` test, but that is skipped by the test suite. Hopefully the unit tests I added should be a suitable correctness check, and I verified that
```
>>> fs = gcsfs.GCSFileSystem(consistency='crc32c')
>>> fs.cat("gs://bucket/file")
```
works manually.

nbren12:
I also don't know how you feel about adding crc32c to the requirements, I could make it an optional import if you prefer.

martindurant:
Do I understand correctly that the tests you include here don't need any real HTTP responses?

martindurant:
> I also don't know how you feel about adding crcmod to the requirements, I could make it an optional import if you prefer.

I think optional would be better. If the uses picks CRC checking, import it then, or offer a nice exception message.

nbren12:
> Do I understand correctly that the tests you include here don't need any real HTTP responses?

Correct. The new tests are local only w/ some mocked fields.

I will add crcmod as an optional requirement.

nbren12:
Ok. Should be ready for review now @martindurant.

gcsfs/core.py
```diff
@@ -1347,15 +1322,10 @@ def validate_response(self, status, content, json, path, headers=None):
...
-            if headers is not None and "X-Goog-Hash" in headers:
-                # if header includes md5 hash, check that data matches
-                bits = headers["X-Goog-Hash"].split(",")
-                for bit in bits:
-                    key, val = bit.split("=", 1)
-                    if key == "md5":
-                        md = b64decode(val)
-                        if md5(content).digest() != md:
-                            raise ChecksumError("Checksum failure")
+            if headers is not None:
```
martindurant: Should this be optional, can md5 be missing here?

martindurant (commented):
Thanks for the useful refactor here, it was cleanly done.
I have the one small question, but otherwise all looks good.

gcsfs/core.py
nbren12: Some of the tests pass in `headers=None` so I tried to keep the logic similar:
https://github.com/VulcanClimateModeling/gcsfs/blob/c76f93ce1e9bae5ba05fbbd1833d1608495444db/gcsfs/tests/test_core.py#L951-L951

I'm not sure if this can happen with real GCS responses...probably not.

nbren12: I'm making the if-statement identical to the current version to ensure backwards compatibility.

martindurant: Yes, but I mean that we explicitly use MD5 here, which may be unnecessary and could perhaps be missing, like it was for your case. Or maybe in this case it is indeed guaranteed to be there, I don't know.

nbren12: I see your point. I didn't modify this since the "consistency" flag was not used here. I believe this check is only needed for `_cat_file` since `_{get,put,pipe}_file` have their own validation logic using `consistency`. The checking logic is unique to each operation, so would be hard to implement in `_call`.

I'd propose:
- adding validation to `_cat_file`
- removing the checksum logic from `validate_response/_call`.

nbren12: I could do this in a follow-up PR if you think this is otherwise ready to merge @martindurant.

martindurant: Actually let's leave it until someone needs this. If ever there is a similar exception to the one you saw, the user would naturally find their way here.

PR #355, opened 2021-03-06
Find correct

### LINKED ISSUE #346: Bug: Recursive removes always tries to remove the bucket!
Or at least it seems to try to remove the bucket.
This is a bug new in 0.7.2 (released yesterday).
a call to `fs.rm('gs://<bucket>/<path>', recursive=True, batchsize=500)`, will give the following error, which it doesn't give with 0.7.1:
```
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 1236, in rm
    batchsize=batchsize,
  File "/usr/local/lib/python3.7/site-packages/fsspec/asyn.py", line 71, in sync
    raise exc.with_traceback(tb)
  File "/usr/local/lib/python3.7/site-packages/fsspec/asyn.py", line 55, in f
    result[0] = await future
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 1051, in _rm
    await asyncio.gather(*[self._rmdir(d) for d in dirs])
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 798, in _rmdir
    await self._call("DELETE", "b/" + bucket, json_out=True)
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 549, in _call
    raise e
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 531, in _call
    self.validate_response(status, contents, json, path, headers)
  File "/usr/local/lib/python3.7/site-packages/gcsfs/core.py", line 1321, in validate_response
    raise HttpError(error)
Exception: The bucket you tried to delete was not empty., 409
```

### LINKED ISSUE #354: find does not list all sub-directories
**What happened**:
Using version 0.7.2, `fs.glob(f"{path}/*")` and `fs.find(path, withdirs=True)` does not return all sub-directories.
In particular, nested directories that only contain other directories which contain files are not returned.
For example, `fs.find(path, withdirs=True)` for the path `a/b/file` should return an entry for each `a`, `b`, and `file`, but instead only returns `b` and `file`.

Using version 0.6.2, `fs.glob` returns all sub-directories, which is the expected behavior.

```python
from fsspec.core import get_fs_token_paths
bucket = "bucket"
path = f"gs://{bucket}"
fs, fs_token, paths = get_fs_token_paths(path, storage_options = {"token": "cloud"})
fs.touch(f"{path}/a/b/file")
b = set(fs.glob(f"{path}/a/*"))
assert f"{bucket}/a/b" in b  # directories that contain files are returned
a = set(fs.glob(f"{path}/*"))
assert f"{bucket}/a" in a  # directories that only contain other directories should also be returned
```

This is related to #287 which seems to have been partially fixed for directories containing files.

**Environment**:

- Dask version: N/A
- Python version: 3.8.8
- Operating System: debian:10.3-slim

martindurant:
Fixes #354 (thanks for the concise reproducer @brl0 )
Fixes #346 (cc @pvanderlinden - unless you can show that _ls_from_cache is broken)

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1161,24 +1161,27 @@ class GCSFileSystem(AsyncFileSystem):
         sdirs = set()
         cache_entries = {}
         for o in out:
-            par = self._parent(o["name"])
-            if par not in sdirs:
-                sdirs.add(par)
-                dirs.append(
-                    {
-                        "Key": self.split_path(par)[1],
-                        "Size": 0,
-                        "name": par,
-                        "StorageClass": "DIRECTORY",
-                        "type": "directory",
-                        "size": 0,
-                    }
-                )
-            # Don't cache "folder-like" objects (ex: "Create Folder" in GCS console) to prevent
-            # masking subfiles in subsequent requests.
-            if not o["name"].endswith("/"):
-                cache_entries.setdefault(par, [])
-                cache_entries[par].append(o)
+            par = o["name"]
+            while par:
+                par = self._parent(par)
+                if par not in sdirs:
+                    if len(par) < len(path):
+                        break
+                    sdirs.add(par)
+                    dirs.append(
+                        {
+                            "Key": self.split_path(par)[1],
+                            "Size": 0,
+                            "name": par,
+                            "StorageClass": "DIRECTORY",
+                            "type": "directory",
+                            "size": 0,
+                        }
+                    )
+                # Don't cache "folder-like" objects (ex: "Create Folder" in GCS console) to prevent
+                # masking subfiles in subsequent requests.
+                if not o["name"].endswith("/"):
+                    cache_entries.setdefault(par, []).append(o)
         self.dircache.update(cache_entries)
 
         if withdirs:
--- a/gcsfs/tests/recordings/test_du.yaml
+++ b/gcsfs/tests/recordings/test_du.yaml
@@ -17,16 +17,18 @@ interactions:
   response:
     body:
       string: !!binary |
-        H4sIABvMT18C/4WPsQ7DIBBDfwUxt7Bn7I9EJ7gkqMAh7hCpqvx7Qzt1ymjLlp/fGpxD5lnoiVlP
-        Su/7rm9Ks6OCQ28ihSdre+9mJVojQglsHCULTTbrIjV/LxFkoZrU2crBq6vWuUotC5uKQ1/mG2MN
-        eSGDCUIcgF/gWV4/ygdCxTr84P+/HB99IHep4wAAAA==
+        H4sIAIvJQ2AC/4WPMQ7DIBAEv4KoE+hd5iPWCc42CnCIO4SjyH+PSapULne1I82+NTiHzLPQE7Oe
+        lN73Xd+UZkcFR95ECk/W9t7NSrRGhBLYOEoWmmz25KllYVNxZHViOXh1hTXGGvJCBhOEeDl3kZq/
+        lwiyUE1D8Cs8y+tn+UCoWEcf/P+X4wNVOkHA4wAAAA==
     headers:
       Cache-Control:
-      - private
+      - no-cache, no-store, max-age=0, must-revalidate
       Content-Encoding:
       - gzip
       Content-Type:
       - application/json; charset=utf-8
+      Pragma:
+      - no-cache
       Server:
       - scaffolding on HTTPServer2
       Transfer-Encoding:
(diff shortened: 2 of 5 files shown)


brl0:
@martindurant, Thanks for the incredibly fast response.  I installed this PR and can confirm that it fixes the issue I was running into.

gcsfs/core.py
```diff
@@ -1161,24 +1161,27 @@ def find(self, path, withdirs=False, detail=False, prefix="", **kwargs):
...
-            # Don't cache "folder-like" objects (ex: "Create Folder" in GCS console) to prevent
-            # masking subfiles in subsequent requests.
-            if not o["name"].endswith("/"):
-                cache_entries.setdefault(par, [])
-                cache_entries[par].append(o)
+            par = o["name"]
+            while par:
+                par = self._parent(par)
+                if par not in sdirs:
+                    if len(par) < len(path):
```
pvanderlinden, line 1168: I found the cause of this bug, the issue is that self._parent, doesn't work correctly for files with an empty filename (not really a surprise).

Example:
```
{'kind': 'storage#object',
 'id': 'bucket/something/items//1605786947866168',
 'selfLink': 'https://www.googleapis.com/storage/v1/b/bucket/something%2Fitems%2F',
 'mediaLink': 'https://www.googleapis.com/download/storage/v1/b/bucket/something%2Fitems%2F?generation=1605786947866168&alt=media',
 'name': 'bucket/something/items/',
 'bucket': 'bucket',
 'generation': '1605786947866168',
 'metageneration': '1',
 'contentType': 'application/octet-stream',
 'storageClass': 'STANDARD',
 'size': 0,
...
 'type': 'file'}
 ```

The parent for this spark created "file" called `bucket/something/items/` is the directory `bucket/something/items/`, but `_parent` will return `bucket/something/` instead

Comment (martindurant) [reply gcsfs/core.py:1168]: Oh, you mean because there is an actual item whose name ends with "/"?

Comment (martindurant) [reply gcsfs/core.py:1168]: cc https://github.com/dask/gcsfs/pull/355#pullrequestreview-606488402

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: Sort of. the filename is actually `''` which is, if you prefix it with the bucket and directory: `'bucket/something/items/'` (which looks like a directory)

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: I also just realized `ls` is not returning subdirectories anymore (`0.7.1` actually does include those)

Comment (martindurant) [reply gcsfs/core.py:1168]: ? Sorry, ca you please show this again?

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: If I compare that with the builtins of fsspec, ls should actually include the directories as well.

Comment (martindurant) [reply gcsfs/core.py:1168]: Yes yes, it should - and I think it does! We do have tests for `ls`, but if you come up with something showing the problem, I'll fix that too.

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: Never mind the `ls`. Can't seem to reproduce this anymore, maybe I accidentally broke it with some of the manual changes, sorry.

Comment (martindurant) [reply gcsfs/core.py:1168]: That's perfectly fine with me :)
OK to merge from your view?

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: @martindurant It's not fixing the actual issue. Also, I just figured out the other bug.

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: I think this has to do with caching, let me create a test case

Comment (pvanderlinden) [reply gcsfs/core.py:1168]: This probably deserve's it's own bugreport. But let me include it here, as I think it is related to this code.
I first created the structure:
```
fs.touch('bucket/testcase/df/_SUCCESS')
fs.touch('bucket/testcase/df/')
fs.touch('bucket/testcase/df/sub/file')
```

Now I started a new python process (not entirely sure if this is needed):
```
>>> fs.ls('bucket/testcase/df/')
['bucket/testcase/df/',
 'bucket/testcase/df/_SUCCESS',
 'bucket/testcase/df/sub']
>>> fs.find('bucket/testcase/df/')
['bucket/testcase/df/',
 'bucket/testcase/df/_SUCCESS',
 'bucket/testcase/df/sub/file']
>>> fs.ls('bucket/testcase/df/')
['bucket/testcase/df/_SUCCESS']
>>>  fs.expand_path('bucket/testcase/df/', recursive=True)
['bucket/testcase',
 'bucket/testcase/df',
 'bucket/testcase/df/',
 'bucket/testcase/df/_SUCCESS',
 'bucket/testcase/df/sub',
 'bucket/testcase/df/sub/file']
```

pvanderlinden: This is with 0.7.2 as it is on pypi.

PR #362, opened 2021-03-22
Don't validate the header if the parent filesystem isn't using md5 as the consistency method

### LINKED ISSUE #360: Always tries to validate headers with md5
Even though the consistency checker is set to None, the gcsfs always tries to validate the contents of a response when x-goog-hash is present with md5. 
https://github.com/dask/gcsfs/blob/85eb01886c9303af4dde20500a0b81d0055b41e1/gcsfs/core.py#L1348-L1351

The problem with that is `x-goog-hash` is not only specific to md5, the hash might be a crc32 when the object is composite. So the operation throws a `NotImplementedError` and fails abruptly. 

For the check above, either it should check if the consistency setting is `md5` (`self.consistency == 'md5'`), or handle it with try/except in case it fails (`try: checker.check() except NotImplementedError: pass`), or should be completely gone (since it kinda seems like over doing). I could have submitted a patch though there are 3 different actions that might be taken so I'm just posting the issue now

isidentical:
As per https://github.com/dask/gcsfs/issues/360#issuecomment-801079010, resolves #360

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1349,7 +1349,9 @@ class GCSFileSystem(AsyncFileSystem):
             else:
                 raise RuntimeError(msg)
         else:
-            if headers is not None and "X-Goog-Hash" in headers:
+            if self.consistency != "md5":
+                return None
+            elif headers is not None and "X-Goog-Hash" in headers:
                 checker = MD5Checker()
                 checker.update(content)
                 checker.validate_headers(headers)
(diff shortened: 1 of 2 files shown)


martindurant:
Failure is due to moving the event loop in fsspec - I'll make a PR here to fix shortly.

isidentical:
Thanks for clarifying!

isidentical:
rebased

PR #371, opened 2021-04-06
setup: bump fsspec to 0.9

isidentical:
`0.8` doesn't work with `fsspec<0.9` though it is getting installed (and raising errors on import time)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ google-auth>=1.2
 google-auth-oauthlib
 requests
 decorator
-fsspec>=0.9.0
+fsspec>=0.9.0,<0.10.0
 aiohttp
 ujson
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ google-auth>=1.2
 google-auth-oauthlib
 requests
 decorator
-fsspec>=0.8.0
+fsspec>=0.9.0
 aiohttp
 ujson


darkclouder:
Not only that, even version 0.7.2 is broken because of the `fsspec==0.9.0` udpate.

```
  File ".env/lib/python3.8/site-packages/gcsfs/core.py", line 269, in __init__
    self._session = sync(
  File ".env/lib/python3.8/site-packages/fsspec/asyn.py", line 53, in sync
    raise result[0]
  File ".env/lib/python3.8/site-packages/fsspec/asyn.py", line 20, in _runner
    result[0] = await coro
  File ".env/lib/python3.8/site-packages/fsspec/implementations/http.py", line 25, in get_client
    return aiohttp.ClientSession(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'callback_timeout'
```


Had to freeze `fsspec==0.8.7` to make version `0.7.2.` work again.
Please don't use `>=` versions, but restrict them to major releases.
Since there are breaking changes between `0.8.0` and `0.9.0`, I would even say minor releases.


So `fsspec>=0.9.0,<=0.10.0` ?

martindurant:
Given the way that Dask and most related projects have gone, I think it might make more sense to switch to calendar-versioning, and always ensure that fsspec/s3fs/gcsfs are released together. 

I think that for a library like this, most changes are "breaking", since the downstream libraries don't care just about the user API, but about the internal details. Note, however, that s3fs and gcsfs tests are run as part of the fsspec CI, so at least we know when updates are required.

martindurant:
NB: you are welcome to submit a similar PR to the conda-forge feedstock for build 1. It is not possible to update the release on pypi, however.

PR #375, opened 2021-04-19
Tweak docs `loop=loop`

max-sixty:
Closes https://github.com/dask/gcsfs/issues/352, but the actual error was already fixed in a33c0f5, thanks to @Giuzzilla


martindurant:
Fair enough - although I might come back and clean up this doc sentence anyway. 

Thanks for contributing.

PR #380, opened 2021-04-28
fix 400 errors on retries in _call

### LINKED ISSUE #290: Retry after HttpError code 400
Google Cloud Storage occasionally throws an HTTP Error 400 (which is nominally a 'bad request'.  See the [Google Cloud docs on HTTP response 400](https://cloud.google.com/storage/docs/json_api/v1/status-codes#400_Bad_Request)).  But this happens on requests that have worked previously and work again after retrying.  I've seen these spurious HTTP Error 400s when calling `gcs.du` and when using `dask` to read data from Google Cloud.

The error message from GCP is: `Error 400 (Bad Request)!  That's an error.  Your client has issued a malformed or illegal request.  That's all we know.`

Monkey-patching `gcsfs.utils.is_retriable` fixes the issue for me:

```python
import gcsfs

# Override is_retriable.  Google Cloud sometimes throws
# a HttpError code 400.  gcsfs considers this to not be retriable.
# But it is retriable!

def is_retriable(exception):
    """Returns True if this exception is retriable."""
    errs = list(range(500, 505)) + [
        # Jack's addition.  Google Cloud occasionally throws Bad Requests for no apparent reason.
        400,
        # Request Timeout
        408,
        # Too Many Requests
        429,
    ]
    errs += [str(e) for e in errs]
    if isinstance(exception, gcsfs.utils.HttpError):
        return exception.code in errs

    return isinstance(exception, gcsfs.utils.RETRIABLE_EXCEPTIONS)

gcsfs.utils.is_retriable = is_retriable
```

In a perfect world, I guess the best solution would be to ask Google Cloud to not throw spurious HTTP Error 400s. But perhaps a pragmatic approach is to modify `gcsfs` to retry after HTTP Error 400s :)

**Environment**:

- Dask version: 2.28.0
- Python version: 3.8.5
- Operating System: Ubuntu 20.04 on a Google Cloud VM
- Install method: conda

nbren12:
The kwargs was being overwritten by ._get_args. If any retries are
needed, then on the second iteration datain will always be 'None'.  This
will cause uploading steps to return with exit code 400.

This commit fixes this bug by refactoring the `_call` method to minimize
the number of variabes in the scope.

Resolves #290

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -498,6 +498,27 @@ class GCSFileSystem(AsyncFileSystem):
             kwargs["userProject"] = user_project
         return path, jsonin, datain, headers, kwargs
 
+    async def _request(self, method, path, *args, **kwargs):
+        await self._set_session()
+        self.maybe_refresh()
+        path, jsonin, datain, headers, kwargs = self._get_args(path, *args, **kwargs)
+        async with self.session.request(
+            method=method,
+            url=path,
+            params=kwargs,
+            json=jsonin,
+            headers=headers,
+            data=datain,
+            timeout=self.requests_timeout,
+        ) as r:
+
+            status = r.status
+            headers = r.headers
+            info = r.request_info  # for debug only
+            contents = await r.read()
+
+            return status, headers, info, contents
+
     async def _call(
         self, method, path, *args, json_out=False, info_out=False, **kwargs
     ):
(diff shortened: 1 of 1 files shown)


gcsfs/core.py
```diff
@@ -507,26 +528,9 @@ async def _call(
             try:
                 if retry > 0:
                     await asyncio.sleep(min(random.random() + 2 ** (retry - 1), 32))
-                self.maybe_refresh()
-                path, jsonin, datain, headers, kwargs = self._get_args(
```
nbren12, line 511: note that kwargs is overwritten here. This is the bug.

martindurant:
Perfect, thank you!

PR #384, opened 2021-04-29
Use _call within _get_file to ensure retries

### LINKED ISSUE #383: _get_file does not use _call, and so does not have any retry logic
See discussion in https://github.com/VulcanClimateModeling/fv3net/issues/1136.

I'm working on a PR to change the implementation of `_get_file` to use `_call` and hopefully leverage the retry logic in that routine.

oliverwm1:
Use `_call` within `_get_file` to ensure that the retry logic in that function is exercised. This PR makes a significant change in implementation, since now multiple http requests are made within single `_get_file` call (if the file is large) whereas previously a single request was made. Not sure about performance implications (speed, memory usage, reliability).

Resolves #383

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1189,10 +1189,22 @@ class GCSFileSystem(AsyncFileSystem):
         checker = get_consistency_checker(consistency)
         os.makedirs(os.path.dirname(lpath), exist_ok=True)
 
+        bucket, key = self.split_path(rpath)
+        metadata = await self._call("GET", "b/{}/o/{}", bucket, key, json_out=True)
+        data_size = int(metadata["size"])
         with open(lpath, "wb") as f2:
-            headers, content = await self._call("GET", u2, **kwargs)
-            f2.write(content)
-            checker.update(content)
+            offset = 0
+            while True:
+                print(offset)
+                head = {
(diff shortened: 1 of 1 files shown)


martindurant:
@isidentical , this is probably also of interest to you

oliverwm1:
@martindurant @isidentical thoughts on this? It passes the (skipped) test_get_put tests for me locally, and I've manually called `fs.get` on a small and large file with success. I'm not sure about performance implications and also perhaps there are complications for compressed files? (https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding)

martindurant:
I'm pretty sure we need a way to split up the reads and stream to disc, else we risk memory issues.

oliverwm1:
> I'm pretty sure we need a way to split up the reads and stream to disc, else we risk memory issues.

Okay. Other idea is to define a retry decorator that we can apply to `_call` and `_get_file`. Not totally straightforward to get an interface that will work for both functions, but I think it's doable. Trying to avoid more copy and paste.

martindurant:
Perhaps you can pass the output file into `_call` ?

oliverwm1:
Closing in favor of #387

PR #387, opened 2021-05-01
Retry decorator for _call and _get_file

### LINKED ISSUE #383: _get_file does not use _call, and so does not have any retry logic

oliverwm1:
Resolves #383 by retrying both `_get_file` and `_call` with a new `retry_request` decorator. Differently from #384, this PR leaves internals of `_get_file` intact.


gcsfs/core.py
```diff
@@ -1289,53 +1243,6 @@ def split_path(cls, path):
         else:
             return path.split("/", 1)
 
-    def validate_response(self, status, content, path, headers=None):
```
oliverwm1, line 1292: moved to `retry.py` module (which used to be called `utils.py`)

gcsfs/core.py
```diff
@@ -1289,53 +1243,6 @@ def split_path(cls, path):
...
-                )  # text-like
-            else:
-                raise RuntimeError(msg)
-        else:
-            if self.consistency != "md5":
-                return None
-            elif headers is not None and "X-Goog-Hash" in headers:
-                checker = MD5Checker()
-                checker.update(content)
-                checker.validate_headers(headers)
```
oliverwm1, line 1337: I deleted this part of the function when I copied it. I think doing checksum checks is a separate concern from validating request status, and anyway only the md5 is checked here instead of allowing for size, crc, etc.

oliverwm1:
@martindurant let me know what you think of this as an alternative to #384

gcsfs/retry.py
```diff
@@ -0,0 +1,133 @@
...
+        ) as e:
+            if (
+                isinstance(e, HttpError)
+                and e.code == 400
+                and "requester pays" in e.message
+            ):
+                msg = "Bucket is requester pays. Set `requester_pays=True` when creating the GCSFileSystem."
+                raise ValueError(msg) from e
+            if retry == retries - 1:
+                logger.exception("_call out of retries on exception: %s" % e)
```
martindurant: This is no longer exclusively called from `_call`

gcsfs/retry.py
```diff
@@ -0,0 +1,133 @@
...
+            raise ValueError("Bad Request: %s\n%s" % (path, msg))
+        elif error:
+            raise HttpError(error)
+        elif status:
+            raise HttpError({"code": status, "message": msg})  # text-like
+        else:
+            raise RuntimeError(msg)
+
+
+@decorator
```
martindurant, line 104: Is this a new dependency? Could it be done with `functools.wraps`?

martindurant (commented):
On first look, this all seems reasonable. I have a couple of comments. 
Unfortunately, the diff is a little inflated by the formatter changing some lines that were fine previously - perhaps a new version of `black` :|

gcsfs/retry.py
oliverwm1: Good catch.

gcsfs/retry.py
oliverwm1: No, it's listed in `requirements.txt`. Although doesn't look like it's used anywhere in the codebase right now.

martindurant: OK, never mind. It's small anyway.

oliverwm1:
> Unfortunately, the diff is a little inflated by the formatter changing some lines that were fine previously - perhaps a new version of black :|

yes, black's behavior when confronted with split lines that could fit into one line has changed. Anyway, I reverted the formatting changes.

gcsfs/retry.py
oliverwm1: updated to `func.__name__`

martindurant:
> I reverted the formatting changes.

Still passes :)

gcsfs/core.py
martindurant: +1
We check for this separately in checkers.py, but it didn't always use to be that way.

nbren12:
Is this ready to be merged? It would also be nice to issue a new PyPI release.

oliverwm1:
I think it's good to go. @martindurant let me know if you see remaining issues—I'm happy to address.

PR #399, opened 2021-06-30
spec: fix find()/glob() calls with prefix on cache

### LINKED ISSUE #393: Error during glob: TypeError: list indices must be integers or slices, not str"
**What happened**:

I am getting this error on a call to `.glob` after upgrading gcsfs and fsspec both to `2021.05.0`. 
The exact same code works with `gcsfs=0.6.1` and `fsspec=0.6.2`. The only difference between the two envs that work and don't is the versions of these two libraries (and whatever version changes that may have induced indirectly in the respective envs to other packages).
```
  File "/deps/envs/deps/lib/python3.7/site-packages/fsspec/asyn.py", line 72, in wrapper
  File "/deps/envs/deps/lib/python3.7/site-packages/fsspec/asyn.py", line 53, in sync
    raise result[0]
  File "/deps/envs/deps/lib/python3.7/site-packages/fsspec/asyn.py", line 20, in _runner
    result[0] = await coro
  File "/deps/envs/deps/lib/python3.7/site-packages/gcsfs/core.py", line 637, in _glob
    return await super()._glob(path, prefix=prefix, **kwargs)
  File "/deps/envs/deps/lib/python3.7/site-packages/fsspec/asyn.py", line 435, in _glob
    root, maxdepth=depth, withdirs=True, detail=True, **kwargs
  File "/deps/envs/deps/lib/python3.7/site-packages/gcsfs/core.py", line 943, in _find
    par = o["name"]
TypeError: list indices must be integers or slices, not str"   
```

For the glob to continue to work. The glob format I am using has is a wildcard like `gs://bucket/some/path/prefix_*.txt`.

Here is where things get strange. When I fire up the interpreter in the same conda environment that is producing this error, and try the same glob pattern, everything works fine. Unfortunately executing this code path in the exact way this error is produced is very hard to intercept (remote cluster, inside of a container, etc etc). 

I realize below is a somewhat vague bug report. I hope to get some suggestions as to where to look, so I can produce more useful information.

**Environment**:

- Dask version: 2.18.1
- Python version: 3.7
- Operating System: centos
- Install method (conda, pip, source): conda. also tried pip via conda's env.yml under the `-pip` section, same result.

isidentical:
Resolves #393

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -380,11 +380,11 @@ class GCSFileSystem(AsyncFileSystem):
         # Check if parent dir is in listing cache
         listing = self._ls_from_cache(path)
         if listing:
-            f = [f for f in listing if f["type"] == "file"]
-            if f:
-                return f
-            # parent is listed, doesn't contain the path
-            raise FileNotFoundError(path)
+            for file_details in listing:
+                if file_details["type"] == "file" and file_details["name"] == path:
+                    return file_details
+            else:
+                raise FileNotFoundError(path)
 
         if not key:
             # Attempt to "get" the bucket root, return error instead of
@@ -937,7 +937,7 @@ class GCSFileSystem(AsyncFileSystem):
             delimiter=None,
             prefix=prefix,
         )
-        if not out and key:
+        if not prefix and not out and key:
             try:
                 out = [
                     await self._get_object(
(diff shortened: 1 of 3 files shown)


martindurant:
OK will do - keep pinging me until it's done!

PR #401, opened 2021-07-20
Add test to cover get/put with lists

baxen:
There is currently a bug that this would have caught from
intake/filesystem_spec#712

Will fail until intake/filesystem_spec#713 is available.

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -472,6 +472,19 @@ def test_get_put(consistency):
 
 
 @pytest.mark.skipif(ON_VCR, reason="async fail")
+@my_vcr.use_cassette(match=["all"])
+def test_get_put_list():
+    with gcs_maker(True) as gcs:
+        with tmpfile() as fn:
+            gcs.get([TEST_BUCKET + "/test/accounts.1.json"], [fn])
+            data = files["test/accounts.1.json"]
+            assert open(fn, "rb").read() == data
+            gcs.put([fn], [TEST_BUCKET + "/temp"])
+            assert gcs.du(TEST_BUCKET + "/temp") == len(data)
+            assert gcs.cat(TEST_BUCKET + "/temp") == data
+
+
+@pytest.mark.skipif(ON_VCR, reason="async fail")
 @pytest.mark.parametrize("protocol", ["", "gs://", "gcs://"])
 @my_vcr.use_cassette(match=["all"])
 def test_get_put_recursive(protocol):


martindurant:
I will have a look at this tomorrow. However, this test will fail anyway, unless you can record a response session with vcrpy. I can handle that, if needed.

baxen:
> I will have a look at this tomorrow. However, this test will fail anyway, unless you can record a response session with vcrpy. I can handle that, if needed.

Looks like the get/put family of tests are skipped already on vcr (I copied over the skipif "async fail"). But on my laptop, with fsspec 2021.7.0 I've got:

```
================================================================================= FAILURES ==================================================================================
__________________________________________________________________________ test_get_put_list[None] __________________________________________________________________________

consistency = None

    @pytest.mark.skipif(ON_VCR, reason="async fail")
    @my_vcr.use_cassette(match=["all"])
    @pytest.mark.parametrize("consistency", [None, "size", "md5", "crc32c"])
    def test_get_put_list(consistency):
        if consistency == "crc32c" and gcsfs.checkers.crcmod is None:
            pytest.skip("No CRC")
        with gcs_maker(True) as gcs:
            gcs.consistency = consistency
            with tmpfile() as fn:
                gcs.get([TEST_BUCKET + "/test/accounts.1.json"], [fn])
                data = files["test/accounts.1.json"]
                assert open(fn, "rb").read() == data
>               gcs.put([fn], [TEST_BUCKET + "/temp"])

gcsfs/tests/test_core.py:486:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../.pyenv/versions/3.7.4/envs/gcsfs/lib/python3.7/site-packages/fsspec/asyn.py:88: in wrapper
../../.pyenv/versions/3.7.4/envs/gcsfs/lib/python3.7/site-packages/fsspec/asyn.py:69: in sync
    raise result[0]
../../.pyenv/versions/3.7.4/envs/gcsfs/lib/python3.7/site-packages/fsspec/asyn.py:25: in _runner
    result[0] = await coro
../../.pyenv/versions/3.7.4/envs/gcsfs/lib/python3.7/site-packages/fsspec/asyn.py:382: in _put
    rdirs = other_paths(dirs, rpath)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

paths = [], path2 = ['tests-gcsfs/temp'], is_dir = None

    def other_paths(paths, path2, is_dir=None):
        """In bulk file operations, construct a new file tree from a list of files

        Parameters
        ----------
        paths: list of str
            The input file tree
        path2: str or list of str
            Root to construct the new list in. If this is already a list of str, we just
            assert it has the right number of elements.
        is_dir: bool (optional)
            For the special case where the input in one element, whether to regard the value
            as the target path, or as a directory to put a file path within. If None, a
            directory is inferred if the path ends in '/'

        Returns
        -------
        list of str
        """
        if isinstance(path2, str):
            is_dir = is_dir or path2.endswith("/")
            path2 = path2.rstrip("/")
            if len(paths) > 1:
                cp = common_prefix(paths)
                path2 = [p.replace(cp, path2, 1) for p in paths]
            else:
                if is_dir:
                    path2 = [path2.rstrip("/") + "/" + paths[0].rsplit("/")[-1]]
                else:
                    path2 = [path2]
        else:
>           assert len(paths) == len(path2)
E           AssertionError

../../.pyenv/versions/3.7.4/envs/gcsfs/lib/python3.7/site-packages/fsspec/utils.py:368: AssertionError
```

and with a local install of the fix on fsspec it's passing. Added the recording from the passing test with the fsspec patch, which I hope is right?

gcsfs/tests/test_core.py
```diff
@@ -471,6 +471,23 @@ def test_get_put(consistency):
             assert gcs.cat(TEST_BUCKET + "/temp") == data
 
 
+@pytest.mark.skipif(ON_VCR, reason="async fail")
```
martindurant, line 474: Does this, in fact, fail with VCR?

gcsfs/tests/test_core.py
```diff
@@ -471,6 +471,23 @@ def test_get_put(consistency):
...
+@my_vcr.use_cassette(match=["all"])
+@pytest.mark.parametrize("consistency", [None, "size", "md5", "crc32c"])
```
martindurant: I don't think we really care about consistency in this test

martindurant (commented):
Thanks, this looks good, except for the small simplifications, below.

Are you able to create a VCR recording, or does it not work.

gcsfs/tests/test_core.py
baxen: I was not able to test this - I think it runs through a nearly identical code path to tets_get_put which had this mark though which is why i included it

baxen:
> Thanks, this looks good, except for the small simplifications, below.
> 
> Are you able to create a VCR recording, or does it not work.

I am able to run with `GCSFS_RECORD_MODE=all` and the recordings are created, but there are some differences in them relative to what you have in the repo (seems to be related to auth). I'm also not able to run with `GCSFS_RECORD_MODE=once` or `none`, it hangs indefinitely for me. If you have a chance it'd be great if you could add the recordings since I'm not sure they're correct on my side.

martindurant:
I wasn't able to get vcr to play nicely either - so I'll put this in, and worry about that later.

Issue #361, opened 2021-03-12
Should the default consistency for GCSFile be md5?


isidentical:
Composite objects in GCSFs doesn't have a md5 hash, so it kinda feels weird that the default consistency is md5 (also it never gets used since most of the operations around it is done by `_open` which passes the consistency as either None or the instance's consistency [defaults to also None]). Maybe it should be crc32, or None (just like the GCSFileSystem).

PR #403, opened 2021-07-22
Fix typo in GCSFile docstring

bnaul:
The actual allowed value is "crc32c" not "crc32".

Also per #361 might be worth changing the default value here, I also noticed that it's `"none"` in `GCSFileSystem` so maybe those should be made consistent as well?

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1108,7 +1108,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
             Check for success in writing, applied at file close.
             'size' ensures that the number of bytes reported by GCS matches
             the number we wrote; 'md5' does a full checksum. Any value other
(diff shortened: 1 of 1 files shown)


martindurant:
> Also per #361 might be worth changing the default value here,

I like consistency! I don't have a strong opinion on what's best practice. I believe that it's pretty unlikely that, if a transfer reports OK, that any bytes are missing or corrupted...

bnaul:
The way I came across this docstring in the first place was debugging very rare corrupt GCS files: we're seeing (very roughly) about 1 malformed file per 100,000 or so, or 1 per couple of TB. I just now enabled this check in our pipeline so it remains to be seen whether it resolves the issue, but I think it's pretty likely.

martindurant:
Merging the type fix (thanks) and please do let us know about how often this happens or would have fixed things for you.

Issue #405, opened 2021-08-01
`is_retriable` is catching 404 responses


mouradmourafiq:
**What happened**:

[is_retriable](https://github.com/dask/gcsfs/blob/main/gcsfs/retry.py#L54) is catching 404 responses and retrying several times:

```
_get_file_request out of retries on exception: 404, message='Not Found', ...
```

This behavior happens because `ClientResponseError` is not an `HttpError` subclass, so the status code is not even checked:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/gcsfs/retry.py", line 110, in retry_request
    return await func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/gcsfs/core.py", line 993, in _get_file_request
    r.raise_for_status()
  File "/usr/local/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 1000, in raise_for_status
    raise ClientResponseError(
```

**What you expected to happen**:

I expect that the status code to be respected here:

```python
errs = list(range(500, 505)) + [
        # Request Timeout
        408,
        # Too Many Requests
        429,
    ]
```

I also think that any `ClientResponseError`will be retried because it's a child of `ClientError`:

```python
    return isinstance(exception, RETRIABLE_EXCEPTIONS)
```

```python
RETRIABLE_EXCEPTIONS = (
    requests.exceptions.ChunkedEncodingError,
    requests.exceptions.ConnectionError,
    requests.exceptions.ReadTimeout,
    requests.exceptions.Timeout,
    requests.exceptions.ProxyError,
    requests.exceptions.SSLError,
    requests.exceptions.ContentDecodingError,
    google.auth.exceptions.RefreshError,
    aiohttp.client_exceptions.ClientError,  # <----------
    ChecksumError,
)
```

**Minimal Complete Verifiable Example**:

```python
fs.get("/this/path/does/not-exist")
```

**Anything else we need to know?**:

**Environment**:

- Dask version:
- Python version:
- Operating System:
- Install method (conda, pip, source):

martindurant:
Yes, sounds like a regression. Do you happen to have time to implement logic which checks for a return code first, before falling back to retries for generic errors? Is there any case in which 404 (and others) *should* be retriable?

oliverwm1:
[...]

PR #406, opened 2021-08-21
Avoid retrying responses with 404 status code

mouradmourafiq:
related #405

### DIFF
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -77,7 +77,7 @@ def validate_response(status, content, path):
     """
     if status >= 400:
         if status == 404:
-            raise FileNotFoundError
+            raise FileNotFoundError(path)
 
         error = None
         if hasattr(content, "decode"):
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -66,7 +66,7 @@ def is_retriable(exception):
     return isinstance(exception, RETRIABLE_EXCEPTIONS)
 
 
-def validate_response(status, content, path, headers=None):
+def validate_response(status, content, path):
     """
     Check the requests object r, raise error if it's not ok.
 
@@ -76,6 +76,9 @@ def validate_response(status, content, path, headers=None):
     path: associated URL path, for error messages
     """
     if status >= 400:
+        if status == 404:
+            raise FileNotFoundError
+
         error = None
         if hasattr(content, "decode"):
             content = content.decode()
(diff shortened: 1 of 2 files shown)


gcsfs/retry.py
```diff
@@ -121,8 +111,15 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
-                msg = "Bucket is requester pays. Set `requester_pays=True` when creating the GCSFileSystem."
+                msg = (
+                    "Bucket is requester pays. "
+                    "Set `requester_pays=True` when creating the GCSFileSystem."
+                )
                 raise ValueError(msg) from e
+            # Special test for 404 to avoid retrying the request
+            if isinstance(e, aiohttp.client_exceptions.ClientError) and e.code == 404:
+                logger.debug("Request returned 404, no retries.")
+                raise e
```
martindurant, line 132: Just `raise` alone

martindurant (commented):
I don't have a principled object to changing around the codes and messaging here. I wish we could get someone with a deeper understanding of GCS's errors and http/aiohttp in general to give their opinions.

gcsfs/retry.py
md-mq: Just to be consistent with the rest of the file, see this [line](https://github.com/dask/gcsfs/pull/406/files#diff-d89213d3b1e29d457dc7aed940ccb379f48e8ef45f34495e2ab3934e39124534L130) and this [line](https://github.com/dask/gcsfs/pull/406/files#diff-d89213d3b1e29d457dc7aed940ccb379f48e8ef45f34495e2ab3934e39124534L135)

mouradmourafiq:
Given that the tests are failing, should I revert the change related to the [json loading of the error](https://github.com/dask/gcsfs/pull/406/files#diff-d89213d3b1e29d457dc7aed940ccb379f48e8ef45f34495e2ab3934e39124534L83) or should I adapt the tests to the new behavior.

Just to give more information about why I made the change, in my tests with large folders with several nesting levels and files, when errors are raised, `json.loads` is a blocking behavior and impacts performance, additionally the error formatting does not bring any value in terms of debugging, keeping the full content users can decide to handle the errors later for debugging purposes, but most users are will trade that for the performance benefits.

martindurant:
> should I revert the change related to the json loading of the error

If this leads to tests passing locally, then probably yes. Again, the JSON thing could be a new PR, but I'm not convinced it's a problem - decoding should always take a trivial amount of time.

gcsfs/retry.py
```diff
@@ -76,6 +76,9 @@ def validate_response(status, content, path, headers=None):
     path: associated URL path, for error messages
     """
     if status >= 400:
+        if status == 404:
+            raise FileNotFoundError
```
martindurant: Could we have `FileNotFoundError(path)` ?

md-mq: done

martindurant:
Thanks @mouradmourafiq

PR #411, opened 2021-09-06
feat: implement signing of urls

baszalmstra:
This is a continuation of #277 . This PR implements the `sign` method for Google Cloud Storage based on the example from google. This only works when authenticating with a service account. This is made explicit with a check to see if a `signer_email` is present.

I didn't add tests and documentation yet. I first want to validate that this approach is valid.


martindurant:
@mariusvniekerk , do you want to review?

mariusvniekerk:
So why do we need to reimplement this rather than just calling https://github.com/googleapis/python-storage/blob/ef65048656d7c67acc4f7600d2c4004d89fe8ba7/google/cloud/storage/blob.py#L419 ?

baszalmstra:
> So why do we need to reimplement this rather than just calling https://github.com/googleapis/python-storage/blob/ef65048656d7c67acc4f7600d2c4004d89fe8ba7/google/cloud/storage/blob.py#L419 ?

I was ready previous PRs that suggested reimplementing. I was under the impression that not adding a dependency on `google-cloud-storage` was preferred. If its not, that makes life much easier. :)

martindurant:
We already rely on google-auth-oauthlib - can you spec out home much bigger and environment gets from including google-cloud-storage ?

baszalmstra:
> We already rely on google-auth-oauthlib - can you spec out home much bigger and environment gets from including google-cloud-storage ?
What do you mean?

martindurant:
I mean, with conda or venv, create an environment with gcsfs's bare requirements, and then another which also includes g-cloud-storage. Measure the size of the environments in terms of number of packages and disc usage.

baszalmstra:
Creating a virtual env and installing gcsfs in the environment results in a virtual env of 36mb (-+ 24 packages), Installing `google-cloud-storage` in the same environment grows the environment to 44mb (+- 30 packages). Doesn't look like much of a problem to me. 

I changed the code to use the `google-cloud-storage` library.

gcsfs/core.py
```diff
@@ -1088,6 +1088,29 @@ def split_path(cls, path):
...
+
+        Returns
+        -------
+        URL : str
+            The signed URL
+        """
+        from google.cloud import storage
+
+        bucket, key = self.split_path(path)
+        client = storage.Client(credentials=self.credentials.credentials)
```
martindurant: Does this need any other arguments from the filesystem constructor, like `project`?

baszalmstra: yep, the `project` indeed. I changed it!

baszalmstra:
@mariusvniekerk what do you think?

mariusvniekerk:
My view is that you don't really want to own the continual maintenance burden of having your own implementation of the signer.  If you take a dependency and google makes a new signature version (which WILL happen at some point) you'll be forced to go reimplement that vs just update your dep.

PR #416, opened 2021-09-14
core: implement callbacks for put_file/get_file

isidentical:
Implements fsspec callbacks. Also removes the `_isdir()` check from the `get_file` since the object based filesystems doesn't really support directories (there are obviously cases where this would work, but not really well defined and can cause serious problems when doing a lot of `get_file()` calls since it might cost an actual API call (`info()`)).

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -884,6 +884,7 @@ class GCSFileSystem(AsyncFileSystem):
         consistency=None,
         content_type="application/octet-stream",
         chunksize=50 * 2 ** 20,
+        callback=None,
         **kwargs,
     ):
         # enforce blocksize should be a multiple of 2**18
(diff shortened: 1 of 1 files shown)


martindurant:
I know it's a pain to make new recordings, but would you like to add a test here?

isidentical:
Is tests really necessary for callbacks? If so, I can try but as you said it is really problematic to make new recordings.

martindurant:
OK, I agree. It will be interesting to see to what uses callbacks are put in the wild!

isidentical:
> OK, I agree. It will be interesting to see to what uses callbacks are put in the wild!

Thanks to @skshetry 's work, we are now completely moving away from our custom stream based (opening 2 files, one from the local fs and the other one from the remote fs) `get_file`/`put_file` implementations to fsspec's `get_file`/`put_file` without loosing any progress information, which seems to be going very great!

PR #419, opened 2021-09-23
Update checks for 404 during retry

### LINKED ISSUE #418: 'ClientPayloadError' object has no attribute 'code'
During a `fs.cat(...)`, I encountered the following traceback:

```
File "/opt/conda/lib/python3.8/site-packages/fsspec/asyn.py", line 88, in wrapper
  File "/opt/conda/lib/python3.8/site-packages/fsspec/asyn.py", line 69, in sync
    raise result[0]
  File "/opt/conda/lib/python3.8/site-packages/fsspec/asyn.py", line 25, in _runner
    result[0] = await coro
  File "/opt/conda/lib/python3.8/site-packages/fsspec/asyn.py", line 344, in _cat
    raise ex
  File "/opt/conda/lib/python3.8/site-packages/gcsfs/core.py", line 678, in _cat_file
    headers, out = await self._call("GET", u2, headers=head)
  File "/opt/conda/lib/python3.8/site-packages/gcsfs/core.py", line 340, in _call
  File "/opt/conda/lib/python3.8/site-packages/decorator.py", line 221, in fun
  File "/opt/conda/lib/python3.8/site-packages/gcsfs/retry.py", line 130, in retry_request
    if isinstance(e, aiohttp.client_exceptions.ClientError) and e.code == 404:
AttributeError: 'ClientPayloadError' object has no attribute 'code'
```

The following line is checking for an error code on an exception. 

https://github.com/dask/gcsfs/blob/ae87e07d69cc4d583b3f56c8c7f6ba39e7c8c49e/gcsfs/retry.py#L134

However, neither [ClientError](https://github.com/aio-libs/aiohttp/blob/09ac1cbb4c07b4b3e1872374808512c9bc52f2ab/aiohttp/client_exceptions.py#L44) nor [ClientPayloadError](https://github.com/aio-libs/aiohttp/blob/09ac1cbb4c07b4b3e1872374808512c9bc52f2ab/aiohttp/client_exceptions.py#L224) (which inherits from `ClientError`) carry a `code` member.  

The solution is likely to change the `isinstance` to an `HttpError`.

**Environment**:

- Dask version: 2021.9.0
- GCSFS version: 2021.8.1
- Python version: 3.8
- Operating System: Ubuntu 20.04

chrisroat:
Closes #418

### DIFF
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -131,7 +131,7 @@ async def retry_request(func, retries=6, *args, **kwargs):
                 )
                 raise ValueError(msg) from e
             # Special test for 404 to avoid retrying the request
-            if isinstance(e, aiohttp.client_exceptions.ClientError) and e.code == 404:
+            if isinstance(e, HttpError) and e.code == 404:
                 logger.debug("Request returned 404, no retries.")
                 raise e
             if retry == retries - 1:
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -131,6 +131,9 @@ async def retry_request(func, retries=6, *args, **kwargs):
                 )
                 raise ValueError(msg) from e
             # Special test for 404 to avoid retrying the request
+            if isinstance(e, aiohttp.client_exceptions.ClientResponseError) and e.status == 404:
+                logger.debug("Request returned 404, no retries.")
+                raise e
             if isinstance(e, HttpError) and e.code == 404:
                 logger.debug("Request returned 404, no retries.")
                 raise e
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -131,7 +131,10 @@ async def retry_request(func, retries=6, *args, **kwargs):
                 )
                 raise ValueError(msg) from e
             # Special test for 404 to avoid retrying the request
-            if isinstance(e, aiohttp.client_exceptions.ClientResponseError) and e.status == 404:
+            if (
+                isinstance(e, aiohttp.client_exceptions.ClientResponseError)
+                and e.status == 404
+            ):
                 logger.debug("Request returned 404, no retries.")
                 raise e
             if isinstance(e, HttpError) and e.code == 404:


mouradmourafiq:
@chrisroat thank you for fixing this 👍

TiansuYu:
Just want to report on this change: The introduction of making 404 completely not `retriable` has made `gcsfs>=2021.09` not usable for our team. 

We have been using a customized `gcsfs` (adding 404 into `gcsfs.retry.is_retriable`) for quite some time. (GCS has a very low chance of returning 404 for completely green operations, but this behavior shows up quite frequently if you use GCS for CI every day for the entire team.) Thus it is important for us to ignore and retry 404. 

If I understand correctly, here `retry_request` first raises all kinds of 404 (from different classes) before checking for `is_retriable` which makes our customization faulty. 

It can be solved by either we includes everything under `gcsfs.retry` customized (not sure if this a good long term solution, as this module has been twigging quite frequently). Or you can reorder the if else ordering here to check `is_retriable` first. And in the long run, I would request a feature to check if `retry_request` introduces any breaking changes around 404. Would love to hear your feedbacks on this.

mouradmourafiq:
> We have been using a customized gcsfs (adding 404 into gcsfs.retry.is_retriable) for quite some time. (GCS has a very low chance of returning 404 for completely green operations...)

Normally the library should never return 404 for existing blobs, it should return a different status to raise an error and not confuse the user about the existence of an object, hence if a blob does not exist the library should not retry.

> but this behavior shows up quite frequently if you use GCS for CI every day for the entire team. Thus it is important for us to ignore and retry 404.

Is it possible to check that in the CI instead? Also do you have some race condition in your CI?

mouradmourafiq:
@TiansuYu I might know where the errors you are seeing are coming from. Are you using by any chance `gcsfs>=2021.9.0`. 

If that's the case, I think this commit introduced a regression: https://github.com/dask/gcsfs/commit/552ccb3735b9bcfe153977ca78fa842b05ad59bf#diff-efbb8d206455e951a2ed240275e4d88d01119dbefa1dbd37a18a0ab49da334b1R1027

the check `is_dir` is no longer performed and the library perform download calls, with the retry it works because the recursive call does not abort with the effect of the exponential backoff.

TiansuYu:
Sorry I have made a typo above: the error code was actually 400. Sorry you took so much time into it. I hope the general context still has some value in it.

PR #427, opened 2021-10-20
Add support to additional 'fixed-key-metadata'

### LINKED ISSUE #426: Add support for other Fixed-key metadata
While trying to use a bucket lifecycle that depends on the Custom-Time Metadata I releasized that gcsfs onlys supports one fixed-key metadata `content-type`. 

I would like to add support for other metadata described here:

Should I just add a new arguments to all the related function that should set / update those ?

Darune:
Add support for the following additionnal "fixed-key" metadata
- content_encoding
- cache_control
- content_disposition
- content_language
- custom_time

Google documentation:

fixes #426

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1551,10 +1551,10 @@ async def simple_upload(
         + metadata
         + "\n--==0=="
         + "\n{}".format(
-            [
+            '\n'.join([
                 '{}: {}'.format(key, value)
                 for key, value in fixed_key_metadata.items()
-            ].join('\n'))
+            ]))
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks for the PR.
Leading up to pydata, I'm not sure when I'll be able to look at this, perhaps someone else has the time. I do notice that it adds quite a lot of lines, and wonder whether it might be made more compact - i.e., it's not necessary to use the same logic as was used for content-type .

Darune:
i'll open a new one with less change, i'm closing this one

PR #429, opened 2021-10-21
Add support to additional 'fixed-key-metadata'

### LINKED ISSUE #426: Add support for other Fixed-key metadata

Darune:
Initially supporting only content-type we now support the following
- content_encoding
- cache_control
- content_disposition
- content_language
- custom_time

They are now all encapsulated in the `google_metadata` argument
And the code is only where it matters compared #427 

Google documentation:
  cloud.google.com/storage/docs/metadata#mutable

fixed #426

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -61,6 +61,14 @@ QUOTE_TABLE = str.maketrans(
     }
 )
 
+SUPPORTED_FIXED_KEY_METADATA = {
+    "content_encoding": "contentEncoding",
+    "cache_control": "cacheControl",
+    "content_disposition": "contentDisposition",
+    "content_language": "contentLanguage",
+    "custom_time": "customTime",
+}
+
 
 def quote_plus(s):
     """
@@ -735,7 +743,12 @@ class GCSFileSystem(AsyncFileSystem):
     getxattr = sync_wrapper(_getxattr)
 
     async def _setxattrs(
-        self, path, content_type=None, content_encoding=None, **kwargs
+        self,
+        path,
+        content_type=None,
+        content_encoding=None,
+        fixed_key_metadata=None,
+        **kwargs,
     ):
         """Set/delete/add writable metadata attributes
 
@@ -744,8 +757,18 @@ class GCSFileSystem(AsyncFileSystem):
         content_type: str
             If not None, set the content-type to this value
         content_encoding: str
+            This parameter is deprecated, you may use fixed_key_metadata instead.
             If not None, set the content-encoding.
             See https://cloud.google.com/storage/docs/transcoding
+        fixed_key_metadata: dict
+            Google metadata, in key/value pairs, supported keys:
+                - cache_control
+                - content_disposition
+                - content_encoding
+                - content_language
+                - custom_time
+            More info:
+            https://cloud.google.com/storage/docs/metadata#mutable
         kw_args: key-value pairs like field="value" or field=None
             value must be string to add or modify, or None to delete
 
@@ -757,7 +780,12 @@ class GCSFileSystem(AsyncFileSystem):
         if content_type is not None:
             i_json["contentType"] = content_type
         if content_encoding is not None:
+            logger.warn(
+                "setxattrs: content_encoding parameter is now deprecated "
+                "you may use `fixed_key_metadata` instead"
+            )
             i_json["contentEncoding"] = content_encoding
+        i_json.update(_convert_fixed_key_metadata(fixed_key_metadata))
 
         bucket, key = self.split_path(path)
         o_json = await self._call(
@@ -1104,6 +1132,7 @@ class GCSFileSystem(AsyncFileSystem):
         consistency=None,
         metadata=None,
         autocommit=True,
+        fixed_key_metadata=None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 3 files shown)


gcsfs/core.py
```diff
@@ -758,6 +767,7 @@ async def _setxattrs(
             i_json["contentType"] = content_type
         if content_encoding is not None:
             i_json["contentEncoding"] = content_encoding
+        i_json.update(await convert_google_metadata(google_metadata))
```
jhamman: Seems like the `contentEncoding` field was already supported. Is it worth thinking about consolidating here?

gcsfs/core.py
```diff
@@ -1399,6 +1430,17 @@ def _fetch_range(self, start=None, end=None):
...
 
+async def convert_google_metadata(metadata, from_google=False):
+    out = {}
+    if metadata is None:
+        return out
+
+    for key, attribute_name in SUPPORTED_GOOGLE_METADATA.items():
+        src = key if not from_google else attribute_name
+        dst = attribute_name if not from_google else key
+        if metadata.get(src) is not None:
```
jhamman:
```suggestion
        if src in metadata:
```

gcsfs/core.py
Darune: It looks like it's only supported by setxattrs, and removing it might cause a breaking change, My code added the support for it in the open().

If we would like to consolidate here, i see the following options:
1- Add the support in the open() with the content_encoding parameter
2- remove content_encoding from the setxattrs to only support the newly introduced `fixed_key_metadata` but this might break things
3- anything else ?

IMO second option 2 seems better but I'm not sure I'm the right person to take this decision, what do you think ?

gcsfs/core.py
Darune: sure.

side note: This is a bad habit I catched when i had to fix a bug whith the in operator where metadata would be a string instead of a dict and cause a hard to debug crash later on.

gcsfs/core.py
martindurant: I think it's OK to look for the old kwarg, and raising if it is found, pointing to the new way to do it. Alternatively we keep supporting it for a while and deprecate later.

Darune: additionnal option could also be to just log.warn that this parameter is deprecated and that this parameter will be removed in favor of `fixed_key_metadata`

martindurant: If that's not too complicated, sounds like a good compromise

gcsfs/core.py
```diff
@@ -1251,6 +1277,10 @@ def __init__(
             "contentType", "application/octet-stream"
         )
         self.metadata = metadata or det.get("metadata", {})
+        self.fixed_key_metadata = sync(
```
martindurant: This does not need the `sync` if `convert_fixed_key_metadata` is not `async`

gcsfs/core.py
```diff
@@ -1400,6 +1432,19 @@ def _fetch_range(self, start=None, end=None):
             raise
 
 
+async def convert_fixed_key_metadata(metadata, from_google=False):
```
martindurant: What does `from_google` mean?

gcsfs/core.py
```diff
@@ -1418,11 +1463,17 @@ async def upload_chunk(fs, location, data, offset, size, content_type):
...
+    bucket,
+    key,
+    content_type="application/octet-stream",
+    metadata=None,
+    fixed_key_metadata=None,
 ):
     j = {"name": key}
     if metadata:
         j["metadata"] = metadata
+    j.update(await convert_fixed_key_metadata(fixed_key_metadata))
```
martindurant: Likewise, we don't need `await` here, because `convert_fixed_key_metadata` shouldn't be `async`

gcsfs/core.py
Darune: Google returns fixed_key_metadata in camelCase and we expose snake_case parameter

from_google means that we want to convert from camelCase to snake_case
otherwise we convert from snake_case to camelCase.

I was a bit out of inspiration for this one, any suggestion ?

gcsfs/core.py
Darune: I don't really know much about async await. but since convert_fixed_key_metadata is called in an async context, I had the wront impression that calling none async method would cause issue.

gcsfs/core.py
Darune: sure

gcsfs/core.py
Darune: Do you want this change in a separate commit for deprecation tracking when updating the changelog ? should I just edit the changelog myself ?

gcsfs/core.py
martindurant: The name is fine, but docstring would be welcome. If the argument isn't used, it can be removed. 
There is an argument that this function should be marked internal with `_`, since users will never call it.

gcsfs/core.py
martindurant: Directly editing the changelog is best, and you can include it all in here.

Darune: it's hard to guess the version of the next release so I from your version I guessed `2021.11.0`

martindurant: Sounds likely :)

gcsfs/core.py
Darune: sure will do

PR #449, opened 2022-02-16
Fix KeyError when calling `_list_buckets` with disabled caching

### LINKED ISSUE #448: `GCSFileSystem._list_buckets` throws a KeyError when `cache_timeout<=0`
Hi,

When we create a `GCSFileSystem` with disabled caching a KeyError is raised when trying to list existing buckets.
The `GCSFileSystem` is created like this:
```
fs = gcsfs.GCSFileSystem(
    project=self.project,
    token=self.token if self.token else None,
    cache_timeout=-1,
)
```
Error:
```
File "/env/lib/python3.8/site-packages/gcsfs/core.py", line 402, in buckets
    b["name"] for b in sync(self.loop, self._list_buckets, timeout=self.timeout)
File "/env/lib/python3.8/site-packages/fsspec/asyn.py", line 71, in sync
    raise return_result
File "/env/lib/python3.8/site-packages/fsspec/asyn.py", line 25, in _runner
    result[0] = await coro
File "/env/lib/python3.8/site-packages/gcsfs/core.py", line 575, in _list_buckets 
    return self.dircache[""] 
File "/env/lib/python3.8/site-packages/fsspec/dircache.py", line 62, in __getitem__ 
    return self._cache[item]  # maybe raises KeyError 
KeyError: ''
```

We found the problem in the `_list_buckets` [function:](https://github.com/fsspec/gcsfs/blob/6ce46c2abcb856217d204812ae8bd33f3dfc21e7/gcsfs/core.py#L549)
If the `dircache` is empty, the bucket info is requested via API and then saved to the cache, and then `return self.dircache[""]` is called.
But since the `cache_timeout` is set negative, the cache entries are immediately invalidated so that a KeyError is raised.

I was able to monkey-patch the function by replacing 
```
self.dircache[""] = [
    {"name": i["name"] + "/", "size": 0, "type": "directory"} for i in items
]
```
([line 572](https://github.com/fsspec/gcsfs/blob/6ce46c2abcb856217d204812ae8bd33f3dfc21e7/gcsfs/core.py#L572)) with
```
buckets = [
    {"name": i["name"] + "/", "size": 0, "type": "directory"} for i in items
]
self.dircache[""] = buckets
return buckets
```

SophieHallstedtQC:
Fixes #448

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -569,9 +569,11 @@ class GCSFileSystem(AsyncFileSystem):
                 items.extend(page.get("items", []))
                 next_page_token = page.get("nextPageToken", None)
 
-            self.dircache[""] = [
+            buckets = [
                 {"name": i["name"] + "/", "size": 0, "type": "directory"} for i in items
             ]
+            self.dircache[""] = buckets
+            return buckets
         return self.dircache[""]
 
     def invalidate_cache(self, path=None):


martindurant:
Thank you, looks great.

PR #452, opened 2022-03-14
Close auth session

JacobHayes:
I noticed that this `requests.Session()` wasn't being closed when running in (py)tests with `filterwarnings = ["error"]` and getting this warning (raised as an error):
```
ResourceWarning: unclosed <ssl.SSLSocket fd=24, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0,
```

After this change and reinstalling the package locally, the issue was fixed. Easier to see when [ignoring whitespace only changes](https://github.com/fsspec/gcsfs/pull/452/files?w=1)

### DIFF
--- a/gcsfs/credentials.py
+++ b/gcsfs/credentials.py
@@ -156,13 +156,14 @@ class GoogleCredentials:
             return  # anon
         if self.credentials.valid:
             return  # still good
-        req = Request(requests.Session())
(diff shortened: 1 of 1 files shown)


martindurant:
Interesting. Not sure why this is necessary, but thank you for making it happen!

PR #457, opened 2022-03-16
Fix check connection

janjagusch:
While working on #455 I accidentally encountered some strange behavior when setting `GCSFileSystem.check_connection=True`. The main problem is that we're trying to call the `ls` method from inside a `GoogleCredentials` instance, where this method is not defined. I moved infering the connection method and checking the connection into the `GCSFileSystem` class. I also had to adjust the connection check, as I couldn't access the `anaconda-public-data` bucket. To be honest, I'm unsure whether the connection check is sufficient, I'd be glad to get some feedback on whether we could introduce a more rigid check here. I also changed the default value for `check_connection` to `True`, as I feel like this is a more sensible default and because it makes it easier to test this functionality.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -13,7 +13,6 @@ import warnings
 import weakref
 
 import fsspec
-import google
 from fsspec.asyn import AsyncFileSystem, sync, sync_wrapper
 from fsspec.callbacks import NoOpCallback
 from fsspec.implementations.http import get_client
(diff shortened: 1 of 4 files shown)


martindurant:
Before having a look at your code, I suspect that this feature was simply not used and thus neglected. Do you think we should remove it?

martindurant:
Let's see if anyone else on here has any opinion for a day or so.

janjagusch:
> Let's see if anyone else on here has any opinion for a day or so.

@martindurant, considering we haven't received any other ideas, should we just remove the `check_connection` feature?

janjagusch:
> Yes, I think so.

Cool, I've deprecated the `check_connection` functionality.

gcsfs/core.py
```diff
@@ -271,14 +267,36 @@ def __init__(
...
+            try:
+                self.credentials = GoogleCredentials(project, access, meth)
+                logger.debug("Connected with method %s", meth)
+                break
+            except google.auth.exceptions.GoogleAuthError as e:
+                # GoogleAuthError is the base class for all authentication
+                # errors
+                logger.debug('Connection with method "%s" failed' % meth, exc_info=e)
+        else:
+            raise RuntimeError("All connection methods have failed!")
```
martindurant: anon never fails

gcsfs/credentials.py
```diff
@@ -209,20 +213,6 @@ def connect(self, method=None):
             None,
         ]:
             self._connect_token(method)
-        elif method is None:
```
martindurant: Was there a reason to move this block to GCSFileSystem? Something in the tests seems to be getting here with `None` and failing.

gcsfs/core.py
janjagusch: thanks, that's good to know. i think it's still a good idea to raise instead of leaving `self.credentials=None`, in case something unexpected happens. but I can add a comment, indicating that the else branch should technically never be executed.

gcsfs/credentials.py
janjagusch: yes, the reason i initially moved this outside the `GoogleCredentials` class was that that chunk of code tires to call the `self.ls` method, which is only implemented for the `GCSFileSystem` class. now that we've deprecated the `self.check_credentials` argument, we technically move it back into the `GoogleCredentials` class, even though I personally prefer having the credential guessing outside the credentials class.

about the test failure: it should be enough the delete the `test_googlecredentials_none` test case or to make it assert that an error is raised. providing `token=None` is not valid anymore, since the credential guessing happens outside of this class.

gcsfs/core.py
martindurant: ok

gcsfs/credentials.py
martindurant: Let's move it back and maintain the previous test.

janjagusch:
i've incorporated your suggestions, @martindurant. i'd suggest squashing the commits when merging since my commit history got a bit messy.

martindurant:
All good, thank you!

Issue #473, opened 2022-05-18
put and pipe methods do not invalidate cache for small files


machow:
While investigating #184 , I noticed that the put and pipe methods--when handling small files--return early, skipping the invalidate_cache statement. This causes calls to things like `ls` to be out of date.

Below is an example, where two subfolders get created with a file:

* `gcsfs_demo/a/some.txt`
* `gcsfs_demo/b/some.txt`.

However, the creation of the second file doesn't invalidate the cache:

```python
import fsspec
fs = fsspec.filesystem("gs")

# add a subfolder that gets listed ----
# lists folder a

fs.pipe("ci-pins/gcsfs_demo/a/some.txt", b"")
fs.ls("ci-pins/gcsfs_demo")

# add a subfolder that doesn't get detected ----
# only lists a, not b

fs.pipe("ci-pins/gcsfs_demo/b/some.txt", b"")
fs.ls("ci-pins/gcsfs_demo")                                 


# bust cache to list both directories ----

fs.invalidate_cache()
fs.ls("ci-pins/gcsfs_demo")
```

* pipe method return: https://github.com/fsspec/gcsfs/blob/main/gcsfs/core.py#L1017
  - earlier return statement
* put method return: https://github.com/fsspec/gcsfs/blob/main/gcsfs/core.py#L1053

PR #474, opened 2022-05-18
Fix cache invalidation in put and pipe (#473)

machow:
Addresses #473, by ensuring invalidate cache is run when small files are created using `.pipe` or `.put`, and adds a test for each.

I had three failing tests, which might be related to me needing to set some test config? Here are the failing test names:

```
FAILED gcsfs/tests/test_core.py::test_request_user_project - ValueError: User-provided project 'project' does not match the google default project 'None'. Either
FAILED gcsfs/tests/test_core.py::test_request_header - TypeError: Invalid variable type: value should be str, int or float, got None of type <class 'NoneType'>
FAILED gcsfs/tests/test_credentials.py::test_googlecredentials_none - ValueError: User-provided project 'myproject' does not match the google default project 'None'. Either
```

Happy to tweak / add anything useful!

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -921,7 +921,7 @@ def test_pipe_small_cache_validity(gcs):
     folder = f"{TEST_BUCKET}/{str(uuid4())}"
 
     gcs.pipe(f"gs://{folder}/a/file.txt", b"")
-    
+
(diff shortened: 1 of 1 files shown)


martindurant:
Please run linting on this - it is failing on whitespace. You can `pre-commit install` (if pre-commit itself is in your environment), if you want that automated.

PR #477, opened 2022-06-07
small test of CI

martindurant:


shuuji3:
Hi, I've investigated these two test failures in #479. Could you review my (partial) fix?

PR #479, opened 2022-06-08
Fix failed tests `test_attrs()` and `test_mkdir_with_path()`

shuuji3:
Hello, this PR is a trial to fix `test_attrs()`, one of the current pytest failures.

I noticed this PATCH call didn't actually update the `metadata` field but PUT did:
https://github.com/fsspec/gcsfs/blob/c2d2d89b540fc289794c69d2741d20e5934a18fa/gcsfs/core.py#L837-L845

I'm not sure this is the best way to fix it since the official documentation recommends us to use a PATCH request for the update of user-defined metadata: [Objects: update  |  Cloud Storage  |  Google Cloud](https://cloud.google.com/storage/docs/json_api/v1/objects/update). Maybe this is an issue of fsouza/fake-gcs-server.

I also delete the following line as the assigned values seem not used anywhere:
https://github.com/fsspec/gcsfs/blob/c2d2d89b540fc289794c69d2741d20e5934a18fa/gcsfs/core.py#L846

Here is a test GitHub Actions run on my fork branch: shuuji3/gcsfs/pull/1

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -789,6 +789,8 @@ def test_array(gcs):
         assert out == b"A" * 1000
 
 
+# https://github.com/fsspec/gcsfs/pull/479
+@pytest.mark.xfail("fake-gcs-server:latest only supports PUT for metadata, not PATCH")
 def test_attrs(gcs):
     gcs.touch(a)
     assert "metadata" not in gcs.info(a)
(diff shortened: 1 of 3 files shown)


shuuji3:
Another test failure for `test_mkdir_with_path` is caused by this new feature of fsouza/fake-gcs-server: https://github.com/fsouza/fake-gcs-server/pull/805. The behavior reflects an actual Cloud Storage behavior so I think we should use another unique directory name for testing.

gcsfs/tests/test_core.py
```diff
@@ -1071,11 +1071,11 @@ def test_dir_marker(gcs):
...
-    assert gcs.exists("new")
+        gcs.mkdir("gcsfs-test-dir/path", create_parents=False)
+    assert not gcs.exists("gcsfs-test-dir")
+    gcs.mkdir("gcsfs-test-dir/path", create_parents=True)
+    assert gcs.exists("gcsfs-test-dir")
 
     # these lines do nothing, but should not fail
-    gcs.mkdir("new/path", create_parents=False)
-    gcs.mkdir("new/path", create_parents=True)
+    gcs.mkdir("gcsfs-test-dir/path", create_parents=False)
```
shuuji3: If we comment out this line (L1080), the test will pass but it still raise  the following error:

```
gcsfs.retry.HttpError: A Cloud Storage bucket named 'gcsfs-test-dir' already exists.
Try another name. Bucket names must be globally unique across all Google Cloud projects,
including those outside of your organization., 409
```

How should we candle this case?

martindurant:
Would you mind testing PUT versus PATCH on real google?

gcsfs/tests/test_core.py
martindurant: I am not bothered by the warning

martindurant:
It makes sense to me that PUT should replace, and PATCH amend the attrs. In our usage, we are after amending.

martindurant:
So should we set xfail on these until the GCS emulator agrees, or add a `replace=` flag to `setattrs` to control whether we use PUT or PATCH?

shuuji3:
Thank you for tidying up my incomplete PR.

Issue #480, opened 2022-06-17
Default of `GCSFileSystem(token=None)` results in unauthenticated `cache` method credentials


noahgolmant:
When calling `GCSFileSystem(token=None)`, if `token=None`, then the credentials methods will [loop through](https://github.com/fsspec/gcsfs/blob/58323d27ef7bdae8d81201b86d8a196d0a9632c8/gcsfs/credentials.py#L218) the `['google_default', 'cache', 'cloud', 'anon']` authentication `method`s. If the `google_default` method fails and the underlying cache is empty (which is the default behavior of the top-level [`load_credentials()`](https://github.com/fsspec/gcsfs/blob/58323d27ef7bdae8d81201b86d8a196d0a9632c8/gcsfs/core.py#L1258) call), then the `_connect_cache()` method will succeed without setting the underlying `credentials` object.

This results in `gsutil ServiceException: 401 Anonymous caller does not have storage.objects.list access to bucket` errors when using `gcsfs` even though the machine may have correct IAM service accounts configured through `gcloud`, since both `gsutil` and `GCSFileSystem(token='cloud')` succeed.

A proposed solution is to raise an exception if `_connect_cache` is called with an empty cache, and pass that exception in the default `method=None` loop.

PR #481, opened 2022-06-17
Fix to also try `cloud` authentication method when `method=None`

noahgolmant:
This PR addresses #480 by raising an exception if we attempt to connect through the cache if the cache is empty. The cache is only available for the browser token-based access. It also modifies the `method=None` loop behavior to skip past the `cache` method and try the `cloud` method if the cache is empty, so the `gauth` IAM service account will also be checked with the compute engine API. Without this, the empty cache connection would result in `self.credentials=None`, which is equivalent to the `anon` connection method. So it would never try the `cloud` method  unless explicitly specified through `GCSFileSystem(token='cloud')`.

### DIFF
--- a/gcsfs/credentials.py
+++ b/gcsfs/credentials.py
@@ -94,6 +94,10 @@ class GoogleCredentials:
         self.credentials = gauth.compute_engine.Credentials()
 
     def _connect_cache(self):
+
+        if len(self.tokens) == 0:
+            raise ValueError("No cached tokens")
+
         project, access = self.project, self.access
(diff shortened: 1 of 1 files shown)


noahgolmant:
The remaining two failing tests are also failing on main with the same errors (`test_attrs`, `test_mkdir_with_path`)

martindurant:
> remaining two failing tests are also failing on main

Yes, we are aware of this

martindurant:
This all seems reasonable to me. Any comments,  @efiop perhaps?

PR #488, opened 2022-08-11
prevent find from corrupting dircache

dtrifiro:
Calling `find` corrupts the `dircache` by adding found objects to each directory, thus when calling `ls` after find, wrong results will be returned.

Here's a script to reproduce this:
```from pprint import pprint

fs = GCSFileSystem()

bucket_name = "dircache_corruption"
fs.rm(f"{bucket_name}", recursive=True)
fs.mkdir(f"{bucket_name}", exists_ok=True)
fs.touch(f"{bucket_name}/file")
fs.touch(f"{bucket_name}/anotherfile")
fs.touch(f"{bucket_name}/afolder/afile")
fs.clear_instance_cache()

print("Calling ls")
first_ls = fs.ls(bucket_name)
print(f"{first_ls=}\n")

print("Calling find")
find = fs.find(bucket_name)
print(f"{find=}\n")

print("Calling ls (again)")
second_ls = fs.ls(bucket_name)
print(f"{second_ls=}\n")

print(f"new items: {set(second_ls)-set(first_ls)}")
print(f"removed items: {set(first_ls)-set(second_ls)}")

dircache = {
    path: [obj["name"] for obj in objs] for path, objs in fs.dircache.items()
}

pprint(dircache)

assert sorted(first_ls) == sorted(second_ls)

```

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1086,54 +1086,55 @@ class GCSFileSystem(AsyncFileSystem):
     async def _find(self, path, withdirs=False, detail=False, prefix="", **kwargs):
         path = self._strip_protocol(path)
         bucket, key = self.split_path(path)
-        out, _ = await self._do_list_objects(
-            path,
-            delimiter=None,
-            prefix=prefix,
-        )
-        if not prefix and not out and key:
-            try:
-                out = [
-                    await self._get_object(
-                        path,
-                    )
-                ]
-            except FileNotFoundError:
-                out = []
-        dirs = []
-        sdirs = set()
+
+        if prefix:
+            _path = "" if not key else key.rstrip("/") + "/"
+            _prefix = f"{_path}{prefix}"
+        else:
+            _prefix = key
+
+        objects, _ = await self._do_list_objects(bucket, delimiter="", prefix=_prefix)
+
+        dirs = {}
         cache_entries = {}
-        for o in out:
-            par = o["name"]
-            while par:
-                par = self._parent(par)
-                if par not in sdirs:
-                    if len(par) < len(path):
-                        break
-                    sdirs.add(par)
-                    dirs.append(
-                        {
-                            "Key": self.split_path(par)[1],
-                            "Size": 0,
-                            "name": par,
-                            "StorageClass": "DIRECTORY",
-                            "type": "directory",
-                            "size": 0,
-                        }
-                    )
-                # Don't cache "folder-like" objects (ex: "Create Folder" in GCS console) to prevent
-                # masking subfiles in subsequent requests.
-                if not o["name"].endswith("/"):
-                    cache_entries.setdefault(par, []).append(o)
+
+        for obj in objects:
+            parent = self._parent(obj["name"])
+            previous = obj
+
+            while parent:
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -1086,54 +1086,70 @@ async def _isdir(self, path):
...
         bucket, key = self.split_path(path)
-        out, _ = await self._do_list_objects(
-            path,
-            delimiter=None,
-            prefix=prefix,
-        )
-        if not prefix and not out and key:
-            try:
-                out = [
-                    await self._get_object(
```
dtrifiro, line 1097: Got rid of the `_get_object` call in order to only use `_do_list_objects`: by properly providing `prefix`, we can list all objects and then process those

efiop (approved):
LGTM. Again, full disclosure: this is coming from dvc team, so I have a conflict of interest and would appreciate someone else taking a look as well 🙂 From our side we can confirm that this works for us.

gcsfs/core.py
```diff
@@ -1086,54 +1086,70 @@ async def _isdir(self, path):
...
+                    "StorageClass": "DIRECTORY",
+                    "type": "directory",
+                    "size": 0,
+                }
+
+                dirs[parent] = dir_info
+                pparent = self._parent(parent)
+                if len(pparent) < len(path):
+                    continue
+                cache_entries.setdefault(pparent, []).append(dir_info)
```
efiop: The number of parents is confusing here. `obj_parent`, `parent`, `pparent`. Also looks like the 
```
if len(pparent) < len(path):
                    continue
```

is equivalent to breaking, right? We've discussed this before, but maybe there is a way to extract a helper out of it, to break up the context and make it more digestable, since we are already getting into renaming stuff.

dtrifiro: Got rid of parents and tried to simplify even more

gcsfs/core.py
martindurant: Is there a chance that the user has get permission but not list? Is there a performance difference? I ask because similar things came up in s3fs.

dtrifiro: I guess that could happen as list requires `storage.objects.list` (https://cloud.google.com/storage/docs/access-control/iam-permissions#object_permissions, although I wouldn't expect find (or `ls` or anything that involves `recursive`) to work in this case, but I might have not understood what you mean here.

About performance: when running `find` on exact paths (e.g.: `gs://bucket/folder/file.txt`), we would be calling `self._do_list_objects`:
https://github.com/fsspec/gcsfs/blob/f169fddbb0ace94418d806ba3a5c0706c7dd08db/gcsfs/core.py#L513-L517

which would add a trailing slash to the provided path, thus querying for a non-existent prefix and resulted in having to do an extra call to `_get_object` in order to get the object, attempting to use dircache first and eventually falling back to the same list call that `_do_list_object` uses: overall I think we'll be making less calls.

martindurant:
Thanks for the work here

gcsfs/core.py
efiop: Looks much better now, thank you! 🔥

PR #491, opened 2022-08-24
docs: add a note regarding aiohttp and proxy envvars

ap--:
Hello everyone,

I just spent quite some time debugging why I couldn't use `gcsfs` to read from a bucket, allthough `gsutil` worked fine. I realized that `aiohttp` ignores `HTTP_PROXY` and `HTTPS_PROXY` by default.

This PR adds a small note to the docs to point other users to the solution.

Cheers,
Andreas 😃


martindurant:
Thanks for digging.

PR #492, opened 2022-08-28
For `_rm`, raise multiple exceptions instead of the first.

alxmrs:
I'm following this example to raise multiple exceptions, with the intention that this will help me better debug a GCS error: https://stackoverflow.com/a/50414672.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -984,7 +984,7 @@ class GCSFileSystem(AsyncFileSystem):
             and not isinstance(ex, FileNotFoundError)
         ]
         if exs:
-            raise exs[0]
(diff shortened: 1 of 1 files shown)


martindurant:
Do you think you can create a test showing this in operation?

I think the idea is reasonable.

ap--:
This seems to be related to [ExceptionGroups](https://docs.python.org/3.11/library/exceptions.html#exception-groups)
There's also a backport: https://github.com/agronholm/exceptiongroup

And here's a relevant discussion https://discuss.python.org/t/accepting-pep-654-exception-groups-and-except/10813/41

PR #493, opened 2022-08-29
Extend Retry logic to address OSErrors. 

alxmrs:
I've encountered a case where large bulk deletions throws an `OSError`. The GCS documentation points out that this is a general problem (https://cloud.google.com/storage/docs/deleting-objects#delete-objects-in-bulk). I did not expect, however, that this would throw an OSError instead of a 500 error.

In any case, this PR extends `gcsfs`'s retry logic to include internal errors that throw `OSError`s. My hope is that this fixes an associated issue in Pangeo Forge Recipes (pangeo-forge/pangeo-forge-recipes#406).

### DIFF
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -63,8 +63,8 @@ def is_retriable(exception):
     if isinstance(exception, HttpError):
         return exception.code in errs
 
-    if isinstance(exception, OSError):
-        return 'internal error' in str(exception)
+    if 'internal error' in str(exception):
+        return True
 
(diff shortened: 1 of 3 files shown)


martindurant:
Is it possible to ascertain what specific OSError happens? A blanket retry on all OSErrors seems like a bad idea, since something like "no internet" should not be retriable.

alxmrs:
It looks like my slow tests is, in fact, too slow. I'm happy to always skip this test.

martindurant:
Given #495 , is this still needed?

PR #499, opened 2022-09-27
don't include parent in find(withdirs=True)

### LINKED ISSUE #498: gcsfs 2022.8.2 does not work with pyarrow.fs.copy_files (earlier versions work)
When using gcsfs 2022.8.2 with pyarrow.fs.copy_files, we run into the following error:

```
  File "scratch2.py", line 6, in <module>
    pyarrow.fs.copy_files("gs://bucket/subdir", "/tmp/somewhere", source_filesystem=fs)
  File "/Users/kai/.pyenv/versions/3.7.7/lib/python3.7/site-packages/pyarrow/fs.py", line 246, in copy_files
    chunk_size, use_threads)
  File "pyarrow/_fs.pyx", line 1229, in pyarrow._fs._copy_files_selector
  File "pyarrow/error.pxi", line 99, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: GetFileInfo() yielded path 'bucket/nested', which is outside base dir 'gs://bucket/nested/subdir'
```

Reproduction script:

```
import pyarrow.fs

fs = pyarrow.fs.PyFileSystem(pyarrow.fs.FSSpecHandler(GCSFileSystem()))
pyarrow.fs.copy_files(
    "bucket/nested/subdir",
    "/tmp/somewhere",
    source_filesystem=fs)
```

The same code works fine with gcsfs 2022.7.1 and earlier.

Confirmed this does not work with pyarrow 6, 7, 8, and 9.

martindurant:
Fixes #498

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -398,7 +398,6 @@ class GCSFileSystem(AsyncFileSystem):
         self, method, path, *args, json_out=False, info_out=False, **kwargs
     ):
         logger.debug(f"{method.upper()}: {path}, {args}, {kwargs.get('headers')}")
-
         status, headers, info, contents = await self._request(
             method, path, *args, **kwargs
         )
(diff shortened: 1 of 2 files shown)


krfricke:
Thank you very much!

Issue #270, opened 2020-07-20
File paths with "#" characters cannot be opened.


zafercavdar:
**What happened**: Experienced `FileNotFoundError` while trying to open a file whose name contains `#` character.

**What you expected to happen**:  The expected case is that `GCSFileSystem` can properly open the specified file independent of its name.

**Minimal Complete Verifiable Example**:

```python
import gcsfs
fs = gcsfs.GCSFileSystem(project="my-project-id")
file_path = "my-bucket-name/data-folder/#my-data.json"
fp = fs.open(file_path, mode="r")
```

**Anything else we need to know?**: I tried to reproduce the issue with other special characters like `$` but I couldn't.

**Environment**:

- gcsfs version: 0.6.1
- Dask version: 2.14.0
- Python version: 3.6.9
- Operating System: Linux, Debian 9
- Install method (conda, pip, source): pip

martindurant:
This must be because of URL standards - the "#" character is supposed to denote a location within a webpage. [This line](https://github.com/dask/gcsfs/blob/master/gcsfs/core.py#L458) should do URL-encoding on the path, but please check to see what it actually does in your case.

Note that the S3 docs (and gcs is largely based on S3) says that a character like "#" is allowed but not recommended in a key name.

zafercavdar:
[`quote_plus`](https://github.com/dask/gcsfs/blob/b98549bbee6b810bc229cdbe947d5bc9ab7e464c/gcsfs/core.py#L78) method does a very simple URL encoding on `/` and white space characters. See  If I url-encode the path with `urllib.parse.quote` method, I can access the file.

martindurant:
If you succeed, feel free to propose changing the encoding function, as a PR.

jchang10:
I just spent the day hunting this issue down. I was using pandas, so it took a while to find the root of the issue.
```
gcfs = gcsfs.GCSFileSystem(project='mproject', token='mytoken')
# required. escape the '#' sign, too.
gcsfs.core.QUOTE_TABLE |= {35:'%23'}
```
Adding the above lines fixed all the problems with #'s in the path.

martindurant:
@jchang10 great! Would you like to make a PR adding that translation code? A test with a "#"-containing filename would be nice:
```
fs.pipe(fn, data)
assert fs.cat(fn) == data
```

PR #502, opened 2022-10-06
Escape all special characters in URL paths

### LINKED ISSUE #447: Doesn't work for a file path containing a hashtag (`#`)
The lib cannot read a file with the file path containing `#`. Weird thing is that it has no error too, just nothing is read. A silent bug.

```python
fs = gcsfs.GCSFileSystem()

# Nothing printed out
data = fs.open("gs://sample_project/my_folder/#this_is_file.json")
print(data.read())
data.close()

# Then I change the file name to remove `#`
# This works
data = fs.open("gs://sample_project/my_folder/this_is_file.json")
print(data.read())
data.close()
```

Not sure if there is an easy way to fix?

simonbohnen:
Filenames including `#` and possibly other special characters are currently not supported.
I would like to change the escaping which is done to use object names in URL paths to be closer to the implementation in `google-cloud-storage` [here](https://github.com/googleapis/python-storage/blob/928ebbccbe183666f3b35adb7226bd259d4e71c0/google/cloud/storage/blob.py#L4403).

Closes #447, #270

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,17 +26,10 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -c conda-forge pytest ujson requests decorator google-auth aiohttp google-auth-oauthlib flake8 black google-cloud-core google-api-core google-api-python-client -y
+          conda install -c conda-forge pytest ujson requests decorator google-auth aiohttp google-auth-oauthlib google-cloud-core google-api-core google-api-python-client -y
           pip install git+https://github.com/fsspec/filesystem_spec --no-deps
           conda list
           conda --version
-
-      - name: Install importlib-metadata for Python 3.7
-        if: matrix.python-version == '3.7'
-        shell: bash -l {0}
-        run: |
-          pip uninstall importlib-metadata -y
-          pip install -I 'importlib-metadata<4.3'
 
       - name: Install
         shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


simonbohnen:
I will add those 👌🏼 I will also check the naming rules for buckets.

Could you point me to where versioning is done in gcsfs?

martindurant:
> Could you point me to where versioning is done in gcsfs?

It seems I was thinking of s3fs, and gcsfs does not support versioning :) GCS does have this feature though ( https://cloud.google.com/storage/docs/object-versioning ), which I think can in theory be used with the current gcsfs interface by entering the generation number in some of the methods (untested).

simonbohnen:
Bucket names may not contain any characters which need to be URL-escaped, see [here](https://cloud.google.com/storage/docs/naming-buckets#requirements).

martindurant:
Please add `importlib-metadata<5.0` to the py37 environment, as the recent release breaks flake 8. We will likely drop py37 sometime soon, but for now...

simonbohnen:
> Please add `importlib-metadata<5.0` to the py37 environment, as the recent release breaks flake 8. We will likely drop py37 sometime soon, but for now...

Thanks for the hint! Could you point me to how to do that? I've tried [without success](https://github.com/simonbohnen/gcsfs/actions/runs/3205869347/jobs/5238923899) 😅

martindurant:
Oh, it's automatically installed by pre-commit. Please try to following:

```diff
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,8 +41,10 @@ jobs:
             export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/gcsfs/tests/fake-secret.json
             py.test -vv gcsfs

-      - name: Run pre-commit hooks
-        shell: bash -l {0}
-        run: |
-          pip install pre-commit
-          pre-commit run --all-file
+  lint:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.0
```

simonbohnen:
That worked, thanks!

martindurant:
Thanks for following this through.

simonbohnen:
Ofc! Could you turn this into a release? We could really use this 😅

martindurant:
I am behind on releases, will get to it :)

PR #503, opened 2022-10-19
deps: bump fsspec to 2022.10.0

skshetry:


martindurant:
You got to it before me :)
There should be an entry in the changelog, even if there were no changes.

skshetry:
Done. I added a changelog anyway, there’s still a bump. :)

PR #504, opened 2022-10-25
core: support object versioning

pmrowla:
Enables support for Cloud Storage [object versioning](https://cloud.google.com/storage/docs/object-versioning).

- Provides a consistent implementation with existing cloud versioning support in s3fs and adlfs
    - `generation` used instead of `version_id` to match the GCS terminology
    - `version_aware` filesystem attribute should be set when using cloud versioning functionality
- Supports specifying object generation in either path `bucket/key?generation=1234` querystring or `bucket/key#1234` fragment to remain consistent with GCS.
    - GCS CLI utilities (`gcloud`) accept and return versioned paths using the `#...` fragment
    - GCS rest API calls (and python SDK calls) use the `?generation=...` querystring (either provided through python SDK kwargs or directly in the URL)
- object `metageneration` is returned in `fs.info` & file`details`, but this implementation only supports returning the latest `metageneration` for a given object version.
- `find(..., versions=True)` will return versioned object names (in `path#generation` form) for all object versions in the specified path/prefix

~~TODO:~~
- [x] add tests (this PR has been tested manually on a real world GCS bucket, I'm still working out how to get versioning working with the fake-gcs-server docker container)

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -23,6 +23,7 @@ from .checkers import get_consistency_checker
 from .credentials import GoogleCredentials
 from . import __version__ as version
 from urllib.parse import quote as quote_urllib
+from urllib.parse import parse_qs, urlsplit
 
 logger = logging.getLogger("gcsfs")
 
@@ -85,7 +86,8 @@ def norm_path(path):
 
     Used by petastorm, do not remove.
     """
-    return "/".join(GCSFileSystem.split_path(path))
+    bucket, name, _ = GCSFileSystem._split_path(path)
+    return "/".join((bucket, name))
 
 
 async def _req_to_text(r):
@@ -117,6 +119,22 @@ def _chunks(lst, n):
     """
     for i in range(0, len(lst), n):
         yield lst[i : i + n]
+
+
+def _coalesce_generation(*args):
+    """Helper to coalesce a list of object generations down to one."""
+    generations = set(args)
+    if None in generations:
+        generations.remove(None)
+    if len(generations) > 1:
+        raise ValueError(
+            "Cannot coalesce generations where more than one are defined,"
+            " {}".format(generations)
+        )
+    elif len(generations) == 0:
+        return None
+    else:
+        return generations.pop()
 
 
 class GCSFileSystem(AsyncFileSystem):
@@ -221,6 +239,9 @@ class GCSFileSystem(AsyncFileSystem):
         Default location where buckets are created, like 'US' or 'EUROPE-WEST3'.
         You can find a list of all available locations here:
         https://cloud.google.com/storage/docs/locations#available-locations
+    version_aware: bool
+        Whether to support object versioning. If enabled this will require the
+        user to have the necessary permissions for dealing with versioned objects.
     """
 
     scopes = {"read_only", "read_write", "full_control"}
@@ -247,6 +268,7 @@ class GCSFileSystem(AsyncFileSystem):
         timeout=None,
         endpoint_url=None,
         default_location=None,
+        version_aware=False,
         **kwargs,
     ):
         super().__init__(
@@ -271,6 +293,7 @@ class GCSFileSystem(AsyncFileSystem):
         self._endpoint = endpoint_url
         self.session_kwargs = session_kwargs or {}
         self.default_location = default_location
+        self.version_aware = version_aware
 
         if check_connection:
             warnings.warn(
@@ -333,6 +356,13 @@ class GCSFileSystem(AsyncFileSystem):
                 path = path[len(protocol) + 2 :]
         # use of root_marker to make minimum required path, e.g., "/"
         return path or cls.root_marker
+
+    @classmethod
+    def _get_kwargs_from_urls(cls, path):
+        _, _, generation = cls._split_path(path, version_aware=True)
+        if generation is not None:
+            return {"version_aware": True}
+        return {}
 
     def _get_params(self, kwargs):
         params = {k: v for k, v in kwargs.items() if v is not None}
@@ -408,8 +438,7 @@ class GCSFileSystem(AsyncFileSystem):
             b["name"] for b in sync(self.loop, self._list_buckets, timeout=self.timeout)
         ]
 
-    @staticmethod
-    def _process_object(bucket, object_metadata):
+    def _process_object(self, bucket, object_metadata):
         """Process object resource into gcsfs object information format.
 
         Process GCS object resource via type casting and attribute updates to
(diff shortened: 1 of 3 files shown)


martindurant:
Thanks, this is looking good! Do ping me when you would like a review.

pmrowla:
@martindurant Question on preferred implementation:

GCS technically allows essentially any valid unicode character in object names. However, with versioning enabled, the gcloud CLI uses `#` URL fragments to denote generation (version ID)

The official [GCS docs](https://cloud.google.com/storage/docs/objects) state:

> It is strongly recommended that you avoid the following in your object names:
>
> - The "#" character: Google Cloud CLI commands interpret object names ending with #<numeric string> as [version identifiers](https://cloud.google.com/storage/docs/using-versioned-objects#access), so including "#" in object names can make it difficult or impossible to perform operations on such versioned objects using the gcloud CLI.
> - The "[", "]", "*", or "?" characters: gcloud storage and gsutil interpret these characters as wildcards, so including them in object names can make it difficult or impossible to perform [wildcard operations](https://cloud.google.com/storage/docs/wildcards) with those tools.

gcsfs currently allows both `#` and `?` to be used in object (path) names (and specifically tests them in `test_special_characters_filename()`).

Should gcsfs continue supporting `#` and `?` in object names? Supporting them complicates things with versioning, since we need to urlparse the object paths in order to get a generation ID.

The one alternative I can think of would be to use `fs.version_aware` to determine whether or not we do the urlparse, so `version_aware=False` would support `#` and `?` in object names, but `version_aware=True` would not. The drawback here is that in this case it's not straightforward to add a gcsfs implementation of `_get_kwargs_from_urls()` which would automatically set `version_aware` based on whether or not the URL includes a `#` or `?`.

We could do some additional validation and also try to check whether or not the URL fragment `#` or query string specify valid generation IDs (which according to the docs are only supposed to be integers), but this still isn't a strictly correct implementation either, since `foo#12345` is technically a valid GCS object name (which differs from object `foo` with `generation=12345`)

pmrowla:
This is ready for review, regarding the object name/generation special character issue, the current implementation uses:

> Using fs.version_aware and doing the valid generation ID verification seems like the best option to me

gcsfs/tests/conftest.py
```diff
@@ -63,7 +63,8 @@ def docker_gcs():
     container = "gcsfs_test"
     cmd = (
         "docker run -d -p 4443:4443 --name gcsfs_test fsouza/fake-gcs-server:latest -scheme "
-        "http -public-host http://localhost:4443 -external-url http://localhost:4443"
+        "http -public-host http://localhost:4443 -external-url http://localhost:4443 "
+        "-backend memory"
```
pmrowla, line 67: Versioning support is only implemented in the fake-gcs-server memory backend (previously we used the default local filesystem backend, but this switch should not matter for gcsfs CI purposes)

martindurant: Nice digging. This is a little odd but indeed doesn't affect us.

pmrowla:
CI issue should be resolved, I had missed that there are separate batch `_rm` implementations that get used depending on whether we are connecting to the actual google API or to an emulator

gcsfs/tests/conftest.py
```diff
@@ -115,5 +116,30 @@ def gcs(gcs_factory, populate=True):
     finally:
         try:
             gcs.rm(gcs.find(TEST_BUCKET))
+            gcs.rm(TEST_BUCKET)
```
pmrowla, line 119: We need to remove the empty bucket when we are done with it since depending on the next test, we may need to re-create the bucket with different configuration options (`versioning.enabled` = true/false)

martindurant: +1

gcsfs/core.py
```diff
@@ -455,23 +492,34 @@ async def _get_object(self, path):
         # Work around various permission settings. Prefer an object get (storage.objects.get), but
         # fall back to a bucket list + filter to object name (storage.objects.list).
         try:
-            res = await self._call("GET", "b/{}/o/{}", bucket, key, json_out=True)
+            print("GET", bucket, key, generation)
```
martindurant: Leftover debug - or should this be logging?

martindurant:
A quick related question: is it possible to know whether an existing bucket supports versioning?

gcsfs/core.py
pmrowla: This was leftover debugging, it's been removed now

pmrowla:
> A quick related question: is it possible to know whether an existing bucket supports versioning?

A `GET` request for the bucket (not for the root object/prefix in the bucket) will return metadata for the bucket that will have a `versioning.enabled` entry set to true if the bucket supports it.

https://cloud.google.com/storage/docs/json_api/v1/buckets/get

martindurant:
> Should gcsfs continue supporting # and ? in object names? 

Any further thoughts, community?

@pmrowla , is it a possible solution to require users to URL-encode "#" themselves if it appears in a path in a version-aware instance? Does that work?

pmrowla:
> @pmrowla , is it a possible solution to require users to URL-encode "#" themselves if it appears in a path in a version-aware instance? Does that work?

This could work, but one caveat would be that we already have to URL-encode (fsspec/gcsfs) paths before we make GCS API calls. So if users are providing already URL-encoded paths to gcsfs, we would need to do the extra decode and then re-encode before calling the API to avoid double-escaping `%` characters which GCS would then misinterpret. (Or we would need to have some flag where the user tells us they are doing the encoding themselves)

pmrowla:
Given that anyone trying to use versioning with Google's own tools cannot reasonably use `#` and `?` in paths, I don't think we need to go too far out of our way to try and support this edge case.

martindurant:
Yes, OK, let's stay with the simple.
I *suppose* it's also OK to assume versioning in url_to_filesystem if the URL contains a "#", since that's the most likely case.

gcsfs/core.py
```diff
@@ -1217,25 +1303,55 @@ def _open(
...
+                        generation = parsed["generation"][0]
+                # Sanity check whether this could be a valid generation ID. If
+                # it is not, assume that # or ? characters are supposed to be
+                # part of the object name.
+                if generation is not None:
+                    int(generation)
+                    key = parts.path
+            except ValueError:
+                generation = None
+        return (
```
martindurant, line 1347: I wonder if we can foresee any problems of having this method not fit the superclass in what it returns - a 3-tuple instead of a 2-tuple

martindurant: OK, actually never mind - I see that s3fs follows the same pattern, and we differentiate between strip_protocol and split_path. All good.

martindurant:
Thank you!

pmrowla:
@martindurant can we get a gcsfs release? This is needed downstream in DVC

pmrowla:
@martindurant Would appreciate getting this released whenever you have the time, thanks!

efiop:
Oh, I see that the releases are not automated here, so I can't help for now 🙁 I guess it is another reminder to me about a potential common fsspec template with a unified automated release process through gha. Need to get into it at some point.

martindurant:
Agreed, that would be nice! Also, to throw out tox...
I will release it in the next couple of days, though. Things got busy.

PR #508, opened 2022-11-11
Update pre-commit, add python versions

brl0:


.github/workflows/ci.yml
```diff
@@ -24,19 +28,16 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
-        shell: bash -l {0}
```
martindurant, line 27: Why are these not needed

.pre-commit-config.yaml
```diff
@@ -2,15 +2,22 @@
...
       - id: end-of-file-fixer
+      - id: fix-byte-order-marker
+      - id: requirements-txt-fixer
+      - id: trailing-whitespace
   - repo: https://github.com/ambv/black
-    rev: 22.3.0
+    rev: 22.10.0
     hooks:
     - id: black
   - repo: https://gitlab.com/pycqa/flake8
```
martindurant: I gather this one is now on github

martindurant (commented):
Looks good, two small questions

.github/workflows/ci.yml
brl0: Those aren't needed for each action now because I added the following as a default:
```yaml
defaults:
  run:
    shell: bash -l -eo pipefail {0}
```

brl0:
Hey Martin,

Thanks for taking a look at this.  The funny thing is this was an accident, I really only meant to open this on my own fork.  I was troubleshooting an issue I was running into that I eventually tracked down to a 2nd order dependency, which I finally got working by pinning `google-auth-oauthlib<0.6`.  

So my apologies if this is a bit of a spammy pull request, I think most of the changes were just me applying my OCD as I was digging into the issue, and honestly I had decided not to open a PR because I didn't feel like I made any meaningful changes. 
 I'd hate to create extra maintenance burden just to enforce my own preferences or opinions.

But since you've taken the time to review, I'll finalize by committing a couple of additional changes I had made before deciding to stop working on this, and will reply once that is done.

As always, thanks for all your community contributions and support.

.github/workflows/ci.yml
martindurant: good to know!

martindurant: good to know

martindurant:
It's always good to have other people poking at the CI! Funnily, I did a similar update in fsspec this week.

.pre-commit-config.yaml
brl0: Correct, also moved under the PSF org.

martindurant:
conda has been inexplicably strolling recently. fsspec switched to mamba: https://github.com/fsspec/filesystem_spec/blob/master/.github/workflows/main.yaml#L49

brl0:
@martindurant,  Finally got it to go green!  So I think this is ready for your review...
I added in dependencies for fuse, but haven't seen that test pass yet, so for now the test is skipped on python 3.7-3.8 due to hanging, and xfailed on 3.9+.  
Besides the changes to CI and pre-commit, I made some minor changes along the way, like converting read only `os.environ` calls to `os.getenv`, using a context manager for opening a file to stop an annoying python warning about an unclosed resource, and renaming a couple of things in tests to make them more obvious or consistent.
Since this was mostly an exercise in my OCD, I would be willing to undo any changes based on your preferences.

.pre-commit-config.yaml
```diff
@@ -1,16 +1,42 @@
...
-    rev: v3.4.0
+    rev: v4.4.0
     hooks:
-      - id: trailing-whitespace
+      - id: check-builtin-literals
+      - id: check-case-conflict
+      - id: check-docstring-first
+      - id: check-executables-have-shebangs
+      - id: check-json
+      - id: check-toml
```
martindurant: I would remove checks that aren't even relevant to this repo, like json, toml, shebang, yaml, private-key, byteorder.

martindurant:
I am happy with everything here, except I wold remove some unnecessary lint checks.
For FUSE, what you've done here is fine for this PR, but we should come back and fix it. Likely the low core count in CI is responsible.

.pre-commit-config.yaml
brl0: I will get those cleaned up, that's just some no-effort random copypasta, maybe from some dask or anaconda related repo, at some point I must have stopped giving them much scrutiny.

PR #513, opened 2022-12-29
Fixed issue in sign url function

tusharvaswanicoder:
There was an issue with sign function `gcsfs.core.GCSFileSystem.sign`. Whenever we were calling sign function it was giving the following error:-

Screenshot from 2022-12-29 17-05-00

Because `split_path` function was giving back 3 values `bucket, key, generation` but only two of them were being stored so I just added `, generation` to fix that issue.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1388,7 +1388,9 @@ class GCSFileSystem(AsyncFileSystem):
         )
         bucket = client.bucket(bucket)
         blob = bucket.blob(key)
-        return blob.generate_signed_url(expiration=expiration, generation=generation, **kwargs)
+        return blob.generate_signed_url(
+            expiration=expiration, generation=generation, **kwargs
+        )
 
 
 GoogleCredentials.load_tokens()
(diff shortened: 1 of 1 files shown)


martindurant:
Thank you, this is useful and could be merged as-is.

However, it would be nice to use the generation, if possible: where the bucket is versioned, [generate_signed_url](https://googleapis.dev/python/storage/latest/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url) seems to take a `generation=` argument we should use. Would you mind incorporating this?

tusharvaswanicoder:
Sure @martindurant I have added that to `generate_signed_url` method.

tusharvaswanicoder:
Hey @martindurant why its still not merged we need the updated package. So I would be thankful to you if you can merge it so that it would be much better for us as well as others. Thanks and if you need any other changes do tell me I would be happy to do that!

martindurant:
Sorry, I did not get a notification of the latest push.

PR #514, opened 2023-01-03
Don't immediately make seesion in init

martindurant:
This enables creating instances inside async contexts without needing asynchronous=True; useful for genericFS

Ref: https://github.com/fsspec/filesystem_spec/issues/1136

@bnaul , maybe this helps your deadlock issue?

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -303,12 +303,6 @@ class GCSFileSystem(AsyncFileSystem):
 
         self.credentials = GoogleCredentials(project, access, token)
 
-        if not self.asynchronous:
-            self._session = sync(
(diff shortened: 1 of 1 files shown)


bnaul:
Thanks @martindurant, I will give it a shot and let you know what we find!

martindurant:
I will merge this regardless as it simplifies the code.

tusharvaswanicoder:
@martindurant I think this caused a new issue I am not sure if this is the issue but I am getting the following problem now:
    async with self.session.get(
AttributeError: 'NoneType' object has no attribute 'get'

martindurant:
@tusharvaswanicoder , can you please try with #515? If it still fails, please leave your traceback, so I can see where the call is coming from.

PR #516, opened 2023-01-05
feat: support created and modified methods in GCSFileSystem

Mause:


gcsfs/core.py
```diff
@@ -741,6 +742,12 @@ async def _rmdir(self, bucket):
 
     rmdir = sync_wrapper(_rmdir)
 
+    def modified(self, path):
+        return datetime.fromisoformat(self.info(path)["updated"])
```
```suggestion
        return datetime.fromisoformat(self.info(path)["updated"].replace(".", ":"))
```

gcsfs/core.py
```diff
@@ -741,6 +742,12 @@ async def _rmdir(self, bucket):
...
+
+    def created(self, path):
+        return datetime.fromisoformat(self.info(path)["timeCreated"])
```
```suggestion
        return datetime.fromisoformat(self.info(path)["timeCreated"].replace(".", ":"))
```

Mause:
Oops, looks like pre 3.11 fromisoformat doesn't support timezones, that's annoying

Mause:
Thankyou!

PR #518, opened 2023-01-13
timestamps without enough zeros

martindurant:
Following #516

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -746,7 +746,9 @@ class GCSFileSystem(AsyncFileSystem):
     def _parse_timestamp(self, timestamp):
         assert timestamp.endswith("Z")
         UTC = timezone(timedelta(0))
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -746,7 +746,9 @@ def created(self, path):
     def _parse_timestamp(self, timestamp):
         assert timestamp.endswith("Z")
         UTC = timezone(timedelta(0))
-        return datetime.fromisoformat(timestamp[:-1]).astimezone(UTC)
+        timestamp = timestamp[:-1]
+        timestamp = timestamp + "0" * (6 - len(timestamp.rsplit(".", 1)[1]))
```
Mause, line 750:
```py
timestamp.rjust(desired_length_here, '0')
```
would have worked well here as well

martindurant: That might have been more elegant, but this works too :)

PR #528, opened 2023-02-09
Switch doc build to use Read the Docs version 2

ianthomas23:
This is the `gcsfs` equivalent of the recent `s3fs` PR fsspec/s3fs#697 so that both projects' doc build setups are as similar as possible. It uses the version 2 config of Read the Docs, fixes a couple of RTD-specific rendering issues (colons between attribute names and types, and bullet points in unordered lists), removes a few doc build warnings, and does a small amount of tidying up.

This is the workflow I use for local building of the docs:

1. `conda env create -n gcsfs_docs -f docs/environment.yml`
2. `conda activate gcsfs_docs`
3. `git clean -fxd`
4. `pip install -ve .`
5. `cd docs`
6. `make html`
7. Point web browser at `build/html/index.html`

For this to work with RTD we will possibly need these (as was done in fsspec/s3fs#697):

- Tick in the "build pull requests for this project" checkbox on the RTD project page.
- RTD oauth may need to be added to the github organisation/project/user (?) account.
- Resyncing the webhooks.
- Force-push to this PR to restart the doc build.


martindurant:
PR builds are enabled. This is the same org and same RTD user as before, web-hook appears to be live, I believe this might be enough.

ianthomas23:
Doc build worked, log is at https://readthedocs.org/projects/gcsfs/builds/19461675/ and generated docs at https://gcsfs--528.org.readthedocs.build/en/528/. It looks OK to me so this is ready for merging.

PR #531, opened 2023-02-24
Don't let find() mess up dircache

### LINKED ISSUE #530: `glob` of an ancestor dir causes `ls` to return multiple identical entries
version `gcsfs==2023.1.0` (but was seen on a late 2022 also) .   python `3.7.10`

Doing a `glob` can cause a subsequent `ls` to return lists with multiple identical entries like:
```
['<root>/t1/t2/t3', '<root>/t1/t2/t3', '<root>/t1/t2/t3']
```

Repro code is below.   The number of duplicates depends on the number of leaf nodes in the tree, whether matched by the glob or not.  Seems cache related but haven't figured out how.

```
import time

root = "<test-root>" # e.g. my-bucket/glob-test-dir

ggparent = root + "/t1"
gparent = ggparent + "/t2"
parent = gparent + "/t3"
leaf1 = parent + "/foo.txt"
leaf2 = parent + "/bar.txt"
leaf3 = parent + "/baz.txt"

fs = GCSFileSystem()

fs.touch(leaf1)
fs.touch(leaf2)
fs.touch(leaf3)
time.sleep(1)
fs.invalidate_cache()

print(fs.ls(gparent))  # -> ['<root>/t1/t2/t3']
print("----")
print(fs.glob(ggparent +"/"))   # trailing "/" is necessary to exibit behavior. More complex
                                # globs also do it , matching or not.
print("----")
print(fs.ls(gparent)) # -> ['<root>/t1/t2/t3', '<root>/t1/t2/t3', '<root>/t1/t2/t3']
```

martindurant:
Fixes #530

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,6 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout
+      - uses: actions/checkout@3
       - uses: actions/setup-python@v4
-      - uses: pre-commit/action
+      - uses: pre-commit/action@3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,6 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@3
+      - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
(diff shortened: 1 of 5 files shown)


isoflavone-r:
Thanks for the quick fix!

is that a stray "print" left at   core.py:1223  ?

martindurant:
Thanks for spotting that - fixed

PR #540, opened 2023-04-14
Wrap BytesIO to prevent accidental closure when retrying

martindurant:


tasansal:
@martindurant we were able to test this. Now we are getting this error:
```python
gcsfs.retry.HttpError: Invalid multipart request with 0 mime parts., 400
```

It looks like it is retrying now and hitting another error. Again, it happens occasionally, can't deterministically replicate it.

The exception raised is [here](https://github.com/fsspec/gcsfs/blob/b968445be614464dc14160fc6a85965947448d15/gcsfs/retry.py#L103)

Anything else we can get you to help debug?

I am looking at a log, and it seems like this is the flow:
1. retry_request gets called.
2. validate_response gets called.
3. it fails with HTTP 400.

martindurant:
Hm, OK - so it seems like aiohttp read to the end of the file, closes it (or tries to) and then on retry it reads from the end of the input, resulting in no data. That is annoying but fixable.

tasansal:
In your wrapper, when `close()` is called, we can call `.seek(0)` instead? Would that be an easy fix; or would it also be retired etc and more complicated?

martindurant:
Yes, that might work - commit pushed

tasansal:
Thanks! Will let you know how it goes. 

I also made a PR to your fork adding short docstrings to explain why this was added; for your review.

https://github.com/martindurant/gcsfs/pull/2

tasansal:
@martindurant, at first glance, seems to be working without any issues! Please feel free to go ahead and complete the PR for the next release. 

Since it is deterministic, we will re-run our tests to ensure we didn't just luck out on the first pass. If it happens again, we can open a new pr.

Thanks a lot for your quick fix and help!

ljstrnadiii:
We have been running into this issue for a while now and this helped a ton! Thank you 🙏

PR #541, opened 2023-04-21
Corrected typo  core.py

prakash12:
Corrected typo from "torage.googleapis" to "storage.googleapis"

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1051,7 +1051,7 @@ class GCSFileSystem(AsyncFileSystem):
 
     @property
     def on_google(self):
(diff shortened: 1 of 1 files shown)


martindurant:
I guess the previous version worked just fine for what was needed here. I wonder if the "s" can ever be capital?

prakash12:
> I guess the previous version worked just fine for what was needed here. I wonder if the "s" can ever be capital?

Agree on it, Change suggested to improve code readability .

martindurant:
Yeah, I think I might leave it as is, just in case something breaks - this is not something we can easily test in CI.

PR #548, opened 2023-05-03
Auto-prepend scheme prefix to STORAGE_EMULATOR_HOST (closes #547)

### LINKED ISSUE #547: Inconsistent `STORAGE_EMULATOR_HOST` format
`gcsfs` seems to treat `STORAGE_EMULATOR_HOST` as a URL, not as a host:port pair.

This is breaking the convention of other libraries (including official Google libs) that treat `STORAGE_EMULATOR_HOST` as a host, and thus makes it impossible to use `gcsfs` with other libraries, because:
- `gcsfs` expects `STORAGE_EMULATOR_HOST` to be `port://host:port`
- `gcloud.aio.storage` & `google.cloud.storage` expect `STORAGE_EMULATOR_HOST` to be `host:port`

STR:
```sh
$ STORAGE_EMULATOR_HOST=127.0.0.1:4443  # Running `fakegcs`
>>> import gcsfs
>>> fs = gcsfs.GCSFileSystem()
>>> fs.listdir('gs://fakebucket/foo/bar')
...
aiohttp.client_exceptions.InvalidURL: 127.0.0.1:4443/storage/v1/b/fakebucket/o?delimiter=/&prefix=foo/bar/
```

A backward-compatible fix would be to detect if the scheme prefix (`http://` or `https://`) is missing, and append it if it is. Probably something as simple as changing https://github.com/fsspec/gcsfs/blob/main/gcsfs/core.py#L117:

```py
_emulator_location = os.getenv("STORAGE_EMULATOR_HOST", None)
if not any(_emulator_location.startswith(scheme) for scheme in ('http://' ,'https://')):
    _emulator_location = f'http://{_emulator_location}'  # Default to http:// since most people use this locally with fake gcs server
```

and3rson:


gcsfs/core.py
```diff
@@ -115,9 +115,13 @@ def _location():
...
     """
     _emulator_location = os.getenv("STORAGE_EMULATOR_HOST", None)
-    return (
-        _emulator_location if _emulator_location else "https://storage.googleapis.com"
-    )
+    if _emulator_location:
+        if not any(
+            _emulator_location.startswith(scheme) for scheme in ("http://", "https://")
+        ):
+            _emulator_location = f"http://{_emulator_location}"
```
martindurant, line 122: I am fine to include this, but is there a reference somewhere saying that http:// should be assumed? Do gcloud CLI tools do this?

gcsfs/core.py
```diff
@@ -1763,3 +1766,5 @@ async def simple_upload(
     )
     checker.update(datain)
     checker.validate_json_response(j)
+    checker.update(datain)
```
martindurant: Why are these two lines replicated?

martindurant (commented):
Thanks for a sensible addition. I just have two questions.

gcsfs/core.py
and3rson: Good question - seems like `http` is assumed with third-party libraries only, since gcloud itself mostly uses grpc (I'm not too familiar with it myself). Here's an example from `gcloud.aio.pubsub`: https://github.com/talkiq/gcloud-aio/blob/master/pubsub/gcloud/aio/pubsub/publisher_client.py#L36

As for the format itself - the same convention seems to be followed across different projects, e.g.:
- Firestore: `FIRESTORE_EMULATOR_HOST` - https://firebase.google.com/docs/emulator-suite/connect_firestore#admin_sdks
  - Examples: https://github.com/googleapis/python-firestore/blob/a2bdb14768cfb366ca6ff8d80c66f3509134b5f0/owlbot.py, https://github.com/googleapis/python-firestore/blob/a8adf1eeaa361574c0d86918c8c87962034b42e6/noxfile.py#L272
- Pubsub: https://cloud.google.com/pubsub/docs/emulator#manually_setting_the_variables
- gcloud.aio.storage (third-party library): https://github.com/talkiq/gcloud-aio/blob/master/storage/gcloud/aio/storage/storage.py#L47

gcsfs/core.py
and3rson: Oops, seems like an accidental addition. Sorry!

gcsfs/core.py
and3rson: ...on the other hand, official library also seems to violate this convention:

https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/_helpers.py#L43

martindurant: 😃

PR #550, opened 2023-05-12
Expose fixed_key_metadata kwarg for put and pipe

### LINKED ISSUE #543: _put_file does not allow to specify content encoding
In order to utilize the [decompressive transcoding](https://cloud.google.com/storage/docs/transcoding) of GCS, I need to set the `content_encoding`. But the `_put_file` does not allow to set it. It allows to set `metadata`, but underneath it calls `simple_upload`, where the `content_encoding` must be set in `fixed_key_metadata` optional argument, and this is not exposed to the `_put_file`, so there is no way to set it.

martindurant:
Fixes #543

@racinmat , care to try this?

How would I write a test to assure that this is doing the right thing locally, perhaps with your gzip workflow?

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1105,6 +1105,7 @@ class GCSFileSystem(AsyncFileSystem):
         metadata=None,
         consistency=None,
         content_type="application/octet-stream",
+        fixed_key_metadata=None,
         chunksize=50 * 2**20,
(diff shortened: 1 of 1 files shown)


racinmat:
I just tried it, confirming it works

martindurant:
Thanks

PR #552, opened 2023-05-12
updated async docs

### LINKED ISSUE #542: Not working async, typo in docs
Docs in https://gcsfs.readthedocs.io/en/latest/#async mention `gcs.set_session()`, but there us only `gcs._set_session()`.
The demo itself is not working, I am running this code
```python
    import asyncio
    loop = asyncio.get_event_loop()

    async def run_program(loop):
        gcs = GCSFileSystem(project='ppt-ctodl-cto-data-40', asynchronous=True, loop=loop)
        await gcs._set_session()
        csv_files = gcs.glob('ds_hcs-nlp_test_by-date_csv/**/*.csv.gz')
        print('done')

    asyncio.run(run_program(loop))  # or call from your async code
```
and it crashes with
```
RuntimeError: Loop is not running
2023-04-25 15:47:11,021 [MainThread] - ERROR    asyncio: Unclosed client session
```

racinmat:
Closes #542

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -150,14 +150,11 @@ await the client creation before making any GCS call.
 
 .. code-block:: python
 
-    loop = ...  # however you create your loop
+    async def run_program():
+        gcs = GCSFileSystem(asynchronous=True)
+        print(await gcs._ls(""))
 
-    async def run_program(loop):
-        gcs = GCSFileSystem(..., asynchronous=True, loop=loop)
-        await gcs.set_session()
-        ...  # perform work
-
-    asyncio.run(run_program(loop))  # or call from your async code
+    asyncio.run(run_program())  # or call from your async code
 
 Concurrent async operations are also used internally for bulk operations
 such as ``pipe/cat``, ``get/put``, ``cp/mv/rm``. The async calls are
@@ -166,6 +163,10 @@ from normal code. If you are *not*
 using async-style programming, you do not need to know about how this
 works, but you might find the implementation interesting.
 
+For every synchronous function there is asynchronous one prefixed by ``_``, but
+the ``open`` operation does not support async operation. If you need it to open
+some file in async manner, it's better to asynchronously download it to
+temporary location and working with it from there.
 
 Proxy
 -----


docs/source/index.rst
```diff
@@ -166,6 +163,10 @@ from normal code. If you are *not*
 using async-style programming, you do not need to know about how this
 works, but you might find the implementation interesting.
 
+For every synchronous function there is asynchronous one prefixed by ``_``, but
+the ``open`` operation does not support async operation. If you need it to open
+some file in async manner, it's better to asynchronously download it to
```
martindurant, line 168: Note that we *may* implement open_async at some point, which will not be like a normal file, supporting only streaming read/write and no seek or random access.

martindurant:
Perfect, thank you.

docs/source/index.rst
alessio-locatelli: @racinmat @martindurant The docs still do not provide a working async example. So far, my best attempt is still raising `RuntimeError: Loop is not running`:

```py
import asyncio
import gcsfs

async def main():
    gcs = gcsfs.GCSFileSystem(asynchronous=True)
    print(await gcs._ls(""))
    with gcs.open('mybucket/new-file', 'wb') as f:
        pass


asyncio.run(main())
```

Are you trying to say that I **cannot** open a file if I use asyncio? Can someone show a working example that opens a file?

<details>
<summary>Versions</summary>

- gcsfs 2025.3.2
- Python 3.13.3

Run my code via `PYTHONASYNCIODEBUG=1 STORAGE_EMULATOR_HOST=` where `STORAGE_EMULATOR_HOST` must point to a running https://github.com/fsouza/fake-gcs-server

</details>

The same non-async code complains about `RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one`:

```py
import gcsfs

def main():
    gcs = gcsfs.GCSFileSystem(asynchronous=False)
    print(gcs.ls(""))
    with gcs.open('mybucket/new-file', 'wb') as f:
        pass

main()
```

martindurant: > I cannot open a file if I use asyncio

Currect - the file-like API derived from io.FileIO, and only supports blocking calls. There is an open_async method, which gives you a _streaming_ asynchronous object. However, this does not support random access by design, because of the internal state that would need to be maintained.

Elsewhere, we have discussed whether we could support something like aiofiles, which is like the file-like API but with async read/write methods. With the current code, it would not be hard to implement, but we would need some design around how best to expose this to users.

PR #554, opened 2023-05-20
GCSFileSystem can accept a raw token

j-bennet:
Related to https://github.com/fsspec/gcsfs/issues/553.

`Credentials` will happily be created from an invalid token - google library doesn't seem to do any checks when the object is instantiated. However, when using such an instance of `Credentials`, an `HttpError` will be raised.

### DIFF
--- a/gcsfs/credentials.py
+++ b/gcsfs/credentials.py
@@ -132,12 +132,6 @@ class GoogleCredentials:
             )
         return token
 
-    def _raw_token_to_credentials(self, token: str):
-        try:
-            return Credentials(token)
-        except Exception:
-            raise ValueError(f"Unable to create Credentials from token: {token}")
-
     def _connect_token(self, token):
         """
         Connect using a concrete token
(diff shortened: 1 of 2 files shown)


martindurant:
> things will break down the line.

Something other than permission errors?

gcsfs/credentials.py
```diff
@@ -132,29 +132,37 @@ def _dict_to_credentials(self, token):
...
     def _connect_token(self, token):
         """
         Connect using a concrete token
 
         Parameters
         ----------
         token: str, dict or Credentials
-            If a str, try to load as a Service file, or next as a JSON; if
+            If a str and a valid file name, try to load as a Service file, or next as a JSON;
+            if not a valid file name, assume it's a valid raw token, and pass to Credentials. If
```
martindurant:
```suggestion
            if not a valid file name, assume it's a valid raw (non-renewable/session) token, and pass to Credentials. If
```

PR #556, opened 2023-05-30
added refresh flag to ls

SchadtJ:
This change is intended to implement the refresh functionality defined on AbstractFileSystem for GCSFileSystem (https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L331).

GCSFileSystem inherits from AsyncFileSystem, which inherits from AbstractFileSystem.

I believe this works as my understanding is "_list_objects" will generate the cache for a path, should one not be found. "_list_objects" will be called after "invalidate_cache".

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -840,7 +840,7 @@ class GCSFileSystem(AsyncFileSystem):
         path = self._strip_protocol(path).rstrip("/")
 
         if refresh:
-            self.invalidate_cache()
+            self.invalidate_cache(path)
         if path in ["/", ""]:
             out = await self._list_buckets()
         else:
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -835,10 +835,12 @@ class GCSFileSystem(AsyncFileSystem):
             prefix = path[:ind].split("/")[-1]
         return await super()._glob(path, prefix=prefix, **kwargs)
 
-    async def _ls(self, path, detail=False, prefix="", versions=False, **kwargs):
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -835,10 +835,12 @@ async def _glob(self, path, prefix="", **kwargs):
             prefix = path[:ind].split("/")[-1]
         return await super()._glob(path, prefix=prefix, **kwargs)
 
-    async def _ls(self, path, detail=False, prefix="", versions=False, **kwargs):
+    async def _ls(self, path, detail=False, refresh=False, prefix="", versions=False, **kwargs):
         """List objects under the given '/{bucket}/{prefix} path."""
         path = self._strip_protocol(path).rstrip("/")
 
+        if refresh:
+            self.invalidate_cache()
```
martindurant: This removes all cached paths, but probably we only wanted to remove this specific one? 
```suggestion
            self.invalidate_cache(path)
```

gcsfs/core.py
```diff
@@ -835,10 +835,12 @@ async def _glob(self, path, prefix="", **kwargs):
             prefix = path[:ind].split("/")[-1]
         return await super()._glob(path, prefix=prefix, **kwargs)
 
-    async def _ls(self, path, detail=False, prefix="", versions=False, **kwargs):
+    async def _ls(self, path, detail=False, refresh=False, prefix="", versions=False, **kwargs):
```
martindurant: Let's put the new kwarg last, so that if anyone is relying on the order, they don't get broken
```suggestion
    async def _ls(self, path, detail=False, prefix="", versions=False, refresh=False, **kwargs):
```

Issue #559, opened 2023-06-22
The info object in GCSFS does not have mtime/ctime


cvm-a:
This is expected by the PyArrow FSSpecHandler _create_file_info method.

This is the GCSFS version of https://github.com/fsspec/filesystem_spec/issues/526

martindurant:
It appears that GCS provides fields 'timeCreated' and 'updated'. Would you like to make PR aliasing these to the format/names expected by pyarrow?

cvm-a:
Sure, I can give that a shot

PR #563, opened 2023-06-26
Translate time metadata from GCS names to fsspec standard names.

cvm-a:
To fix issue #559.


martindurant:
Thank you, looks good!
Any chance of adding a test to ensure we don't loose this behaviour in the future?

cvm-a:
> Thank you, looks good! Any chance of adding a test to ensure we don't loose this behaviour in the future?

The added lines in test_core.py should fail if the two new fields are not present. Is there anything else we should test for?

Also, should we remove the old GCS-specific names? It doesn't seem to be a documented part of the API, but I don't know what is your policy if someone is depending on them despite that (Hyrum's law).

martindurant:
> Also, should we remove the old GCS-specific names?

I'd say it's fine to keep the original names too

PR #573, opened 2023-08-09
Asyncio Listing and Inventory Report Integration

hanseaston:
@martindurant 

I've completed the integration of the inventory report service and asyncio listing logic based on our discussions.

The overview of the logic is as follows:

For listing, the user can optionally pass in a `inventory_report_info` argument as part of the `kwargs`.

If the argument is not passed in, then the listing logic is exactly the same as previously, thus it is backward compatible with existing implementation.

If the argument is passed in, the high-level is as follows:

1. Validate whether `inventory_report_info` has the correct fields included to fetch the inventory report.
2. Fetch the inventory report configuration.
3. Parse the inventory report configuration to validate and extract the inventory report location.
4. Does listing on the inventory report metadata.
5. Use the metadata to fetch the latest inventory report(s).
6. Parse the inventory report content and extract objects metadata.
7. If the user wants to directly use the inventory report for listing, then return the objects metadata.
8. If the user wants to retrieve the most up-to-date listing, then use the object names from the inventory report to divide up workload for each coroutine fast listing.

The listing performance is ~10x from my own experiments. 

Latency for 1 - 6 should be trivial since: 
1. Fetching the inventory report is very fast, a single call.
2. Performing listing on inventory is very fast, most likely a single call (5000 inventory reports on a single page).
3. Downloading inventory report is relatively past (inventory reports with 1M objects is ~40MB).

In addition, if the user chooses to use the inventory report content directly for listing results, the speedup can be >100x.

Comments have been added throughout the implementation, and there are comprehensive test cases covering each method in the `InventoryReport` class.

Would really appreciate it if any of you has any feedback, so that we can have this awesome feature integrated!

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -24,6 +24,7 @@ from fsspec.utils import setup_logging, stringify_path
 from . import __version__ as version
 from .checkers import get_consistency_checker
 from .credentials import GoogleCredentials
+from .inventory_report import InventoryReport
 from .retry import retry_request, validate_response
 
 logger = logging.getLogger("gcsfs")
@@ -537,25 +538,31 @@ class GCSFileSystem(AsyncFileSystem):
                 raise FileNotFoundError(path)
         return self._process_object(bucket, res)
 
-    async def _list_objects(self, path, prefix="", versions=False):
+    async def _list_objects(self, path, prefix="", versions=False, **kwargs):
+
         bucket, key, generation = self.split_path(path)
         path = path.rstrip("/")
 
-        try:
-            clisting = self._ls_from_cache(path)
-            hassubdirs = clisting and any(
-                c["name"].rstrip("/") == path and c["type"] == "directory"
-                for c in clisting
-            )
-            if clisting and not hassubdirs:
-                return clisting
-        except FileNotFoundError:
-            # not finding a bucket in list of "my" buckets is OK
-            if key:
-                raise
+        inventory_report_info = kwargs.get("inventory_report_info", None)
+
+        # Only attempt to list from the cache when the user does not use
+        # the inventory report service.
+        if not inventory_report_info:
+            try:
+                clisting = self._ls_from_cache(path)
+                hassubdirs = clisting and any(
+                    c["name"].rstrip("/") == path and c["type"] == "directory"
+                    for c in clisting
+                )
+                if clisting and not hassubdirs:
+                    return clisting
+            except FileNotFoundError:
+                # not finding a bucket in list of "my" buckets is OK
+                if key:
+                    raise
 
         items, prefixes = await self._do_list_objects(
-            path, prefix=prefix, versions=versions
+            path, prefix=prefix, versions=versions, **kwargs
         )
 
         pseudodirs = [
@@ -574,28 +581,149 @@ class GCSFileSystem(AsyncFileSystem):
             else:
                 return []
         out = pseudodirs + items
-        # Don't cache prefixed/partial listings
-        if not prefix:
+        
+        use_snapshot_listing = False if not inventory_report_info \
+            else inventory_report_info.get("use_snapshot_listing")
+
+        # Don't cache prefixed/partial listings, in addition to 
+        # not using the inventory report service to do listing directly.
+        if not prefix and use_snapshot_listing is False:
             self.dircache[path] = out
         return out
 
     async def _do_list_objects(
-        self, path, max_results=None, delimiter="/", prefix="", versions=False
-    ):
+        self, path, max_results=None, delimiter="/", 
+                    prefix="", versions=False, **kwargs):
+        
         """Object listing for the given {bucket}/{prefix}/ path."""
         bucket, _path, generation = self.split_path(path)
         _path = "" if not _path else _path.rstrip("/") + "/"
         prefix = f"{_path}{prefix}" or None
 
+        # Page size of 5000 is officially supported across GCS.
+        default_page_size = 5000
+
+        inventory_report_info = kwargs.get("inventory_report_info", None)      
+
+        # Check if the user has configured inventory report option.
+        if inventory_report_info is not None:
+            items, prefixes = await \
+                InventoryReport.fetch_snapshot(
+                    gcs_file_system=self,
+                    inventory_report_info=inventory_report_info,
+                    prefix=prefix)
+            
+            use_snapshot_listing = inventory_report_info.get("use_snapshot_listing")
+
+            # If the user wants to rely on the snapshot from the inventory report
+            # for listing, directly return the results.
+            if use_snapshot_listing:
+                return items, prefixes
+            
+            # Otherwise, use the snapshot to initate concurrent listing.
+            return await self._concurrent_list_objects_helper(
+                items=items,
+                bucket=bucket,
+                delimiter=delimiter,
+                prefix=prefix,
+                versions=versions,
+                generation=generation,
+                page_size=default_page_size
+            )
+        
+        # If the user has not configured inventory report, proceed to use
+        # sequential listing.
+        else:
+            return await self._sequential_list_objects_helper(
+                bucket=bucket,
+                delimiter=delimiter,
+                start_offset=None,
+                end_offset=None,
+                prefix=prefix,
+                versions=versions,
+                generation=generation,
+                page_size=default_page_size)
+        
+    async def _concurrent_list_objects_helper(self, items, bucket, delimiter,
+                                            prefix, versions, generation, page_size):
+        """
+        Lists objects using coroutines, using the object names from the inventory
+        report to split up the ranges.
+        """
+        
+        # Extract out the names of the objects fetched from the inventory report.
+        snapshot_object_names = [item["name"] for item in items]
+        snapshot_object_names = sorted(snapshot_object_names)
+
+        # Determine the number of coroutines needed to concurrent listing. 
+        # Ideally, want each coroutine to fetch a single page of objects.
+        num_coroutines = len(snapshot_object_names) // page_size + 1
+        num_objects_per_coroutine = len(snapshot_object_names) // num_coroutines
+
+        start_offsets = []
+        end_offsets = []
+
+        # Calculate the split splits of each coroutine (start offset and end offset).
+        for i in range(num_coroutines):
+            range_start = i * num_objects_per_coroutine
+            if i == num_coroutines - 1:
+                range_end = len(snapshot_object_names)
+            else:
+                range_end = range_start + num_objects_per_coroutine
+
+            if range_start == 0:
+                prefix_start = None
+            else:
+                prefix_start = snapshot_object_names[range_start]
+
+            if range_end == len(snapshot_object_names):
+                prefix_end = None
+            else:
+                prefix_end = snapshot_object_names[range_end]
+
+            start_offsets.append(prefix_start)
+            end_offsets.append(prefix_end)
+
+        # Assign the coroutine all at once, and wait for them to finish listing.
+        results = await asyncio.gather(*[self._sequential_list_objects_helper(
+            bucket=bucket,
+            delimiter=delimiter,
+            start_offset=start_offsets[i],
+            end_offset=end_offsets[i],
+            prefix=prefix,
+            versions=versions,
+            generation=generation,
+            page_size=page_size) for i in range(0, len(start_offsets))])
+
+        items = []
+        prefixes = []
+
+        # Concatenate the items and prefixes from each coroutine for final results.
+        for i in range(len(results)):
+            items_from_process, prefixes_from_process = results[i]
+            items.extend(items_from_process)
+            prefixes.extend(prefixes_from_process)
+  
+        return items, prefixes
+    
+    async def _sequential_list_objects_helper(self, bucket, delimiter,
+            start_offset, end_offset, prefix, versions, generation, page_size):
+        """
+        Sequaltial list objects within the start and end offset range.
+        """
+
         prefixes = []
         items = []
+
         page = await self._call(
             "GET",
             "b/{}/o",
             bucket,
             delimiter=delimiter,
             prefix=prefix,
-            maxResults=max_results,
+            startOffset=start_offset,
+            endOffset=end_offset,
+            maxResults=page_size,
             json_out=True,
             versions="true" if versions or generation else None,
         )
@@ -605,13 +733,16 @@ class GCSFileSystem(AsyncFileSystem):
         next_page_token = page.get("nextPageToken", None)
 
         while next_page_token is not None:
+
             page = await self._call(
                 "GET",
                 "b/{}/o",
                 bucket,
                 delimiter=delimiter,
                 prefix=prefix,
-                maxResults=max_results,
+                startOffset=start_offset,
+                endOffset=end_offset,
+                maxResults=page_size,
                 pageToken=next_page_token,
                 json_out=True,
                 versions="true" if generation else None,
@@ -623,6 +754,7 @@ class GCSFileSystem(AsyncFileSystem):
             next_page_token = page.get("nextPageToken", None)
 
         items = [self._process_object(bucket, i) for i in items]
+
         return items, prefixes
 
     async def _list_buckets(self):
@@ -856,7 +988,7 @@ class GCSFileSystem(AsyncFileSystem):
         else:
             out = []
             for entry in await self._list_objects(
-                path, prefix=prefix, versions=versions
+                path, prefix=prefix, versions=versions, **kwargs
             ):
                 if versions and "generation" in entry:
                     entry = entry.copy()
--- a/gcsfs/inventory_report.py
+++ b/gcsfs/inventory_report.py
@@ -466,7 +466,53 @@ class InventoryReport:
             'prefix' will also be empty, and 'items' will contains the object 
             names in the snapshot.
         """
-        pass
+        if prefix is None:
+            prefix = ""
+
+        # Filter the prefix and returns the list if the user does not want to use
+        # the snapshot for listing.
+        if use_snapshot_listing is False:
+
+            return [obj for obj in objects \
+                if obj.get("name").startswith(prefix)], []
+        
+        else:
+            
+            # If the user wants to use the snapshot, generate both the items and 
+            # prefixes manually.
+            items = []
+            prefixes = set()
+
+            for obj in objects:
+            
+                
+                # Fetch the name of the object.
+                obj_name = obj.get("name")
+
+                # If the object name doesn't start with the prefix, continue.
+                # In the case where prefix is empty, it will always return
+                # true (which is the expected behavior).
+                if not obj_name.startswith(prefix):
+                    continue
+
+                # Remove the prefix.
+                object_name_no_prefix = obj_name.removeprefix(prefix)
+
+                # Determine whether the object name is a directory.
+                first_delimiter_idx = object_name_no_prefix.find("/")
+
+                # If not, then append it to items.
+                if first_delimiter_idx == -1:
+                    items.append(obj)
+                    continue
+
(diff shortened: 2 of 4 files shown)


gcsfs/inventory_report.py
```diff
@@ -0,0 +1,307 @@
...
+        """
+        Validates the inventory report information dictionary that user
+        passes in.
+
+        Parameters:
+            inventory_report_info (dict): A dictionary containing the inventory
+            report information with the following keys:
+                - "use_snapshot_listing" (bool): A flag indicating whether
+                  to use snapshot listing in the inventory report.
+                - "location" (str): The location of the inventory report in GCS.
```
bernardhan33, line 155: Would "region" be a better name than "location" here, since "location" in the context of inventory reports refers to Asia, Australia, etc while "region" refers to "asia-east1", "us-west1" which are the potential values of this parameter?

gcsfs/tests/test_inventory_report.py
```diff
@@ -0,0 +1,35 @@
...
+    methods include a setup method just above them.
+    """
+
+    @pytest.mark.parametrize("inventory_report_info, expected_error", [
+        # Check whether missing inventory report info will raise exception.
+        ({"location": "us-west", "id": "123"}, \
+         "Use snapshot listing is not configured."),
+        ({"use_snapshot_listing": True, "id": "123"}, \
+         "Inventory report location is not configured."),
+        # Check complete inventory report infor will not raise exception.
```
bernardhan33: 1. typo on `info`.
2. should this comment be immediately above line 22?

gcsfs/inventory_report.py
```diff
@@ -127,7 +127,20 @@ class (see 'core.py').
...
+            project,
+            location,
+            id
+        )
+        try:
+            raw_inventory_report_config = await gcs_file_system \
+                ._call("GET", url, json_out=True)
+            return raw_inventory_report_config
+        except Exception as e:
+            print(f"Error countered when fetching inventory report config: {e}.")
```
bernardhan33: typo: `encountered`?

gcsfs/inventory_report.py
```diff
@@ -165,7 +167,55 @@ class representing the parsed inventory report configuration.
...
+        metadata_fields = \
+            object_metadata_report_options.get("metadataFields")
+        storage_destination_options = \
+            object_metadata_report_options.get("storageDestinationOptions")
+        
+        # Save relevant report config properties.
+        csv_options = raw_inventory_report_config.get("csvOptions")
+        bucket = storage_destination_options.get("bucket")
+        destination_path = storage_destination_options.get("destinationPath")
+        metadata_fields = metadata_fields
```
bernardhan33: this might be a silly question -- why is this line needed?

gcsfs/tests/test_inventory_report.py
```diff
@@ -184,6 +186,61 @@ def test_parse_raw_inventory_report_config_returns_correct_config(self):
...
+        # and then a page with one item and without next page token.
+        gcs_file_system._call.side_effect = [{"items": [], "nextPageToken": None}]
+
+        # Create a mock for InventoryReportConfig.
+        inventory_report_config = mocker.MagicMock(spec=InventoryReportConfig)
+        inventory_report_config.bucket = "bucket_name"
+        inventory_report_config.destination_path = "destination_path"
+
+        # If no inventory report metadata is fetched, an exception should be raised.
+        with pytest.raises(ValueError) as e_info:
```
bernardhan33: nit: seems that `pytest.raises()` supports `match` flag which you can save the line of `asserting`. [Example](https://github.com/fsspec/gcsfs/blame/616fdd447ab1427f5e2a26b1af6c6c33a0844ed9/gcsfs/tests/test_core.py#L950C8-L950C8).

On the other hand, do we really want to test the error message info though? That sounds a little like change detector. Is it possible to make a constant in the `inventory_report.py` and reference it in multiple places?

gcsfs/inventory_report.py
```diff
@@ -318,7 +318,41 @@ class (see 'core.py').
...
+        # Get the most recent inventory report date.
+        most_recent_inventory_report = inventory_report_metadata[0]
+        most_recent_date = InventoryReport._convert_str_to_datetime(
+            most_recent_inventory_report.get("timeCreated")).date()
+        
+        inventory_report_content = []
+
+        # Run a for loop here, since there might be multiple inventory reports
+        # generated on the same day. For reference, 1 million objects will be
+        # split into only 2 inventory reports, so it is very rare that there
```
bernardhan33, line 396: It would be great to link a reference here on the cap of 500k objects per single report but I can't find it on our official doc anywhere. feel free to ignore.

gcsfs/inventory_report.py
```diff
@@ -466,7 +466,53 @@ def _construct_final_snapshot(objects, prefix, use_snapshot_listing):
...
+        
+        else:
+            
+            # If the user wants to use the snapshot, generate both the items and 
+            # prefixes manually.
+            items = []
+            prefixes = set()
+
+            for obj in objects:
+
```
bernardhan33: nit: remove double line?

gcsfs/inventory_report.py
```diff
@@ -466,7 +466,53 @@ def _construct_final_snapshot(objects, prefix, use_snapshot_listing):
...
+                # Determine whether the object name is a directory.
+                first_delimiter_idx = object_name_no_prefix.find("/")
+
+                # If not, then append it to items.
+                if first_delimiter_idx == -1:
+                    items.append(obj)
+                    continue
+
+                # If it is, recompose the directory and add to the prefix set.
+                dir = object_name_no_prefix[0:first_delimiter_idx]
```
bernardhan33: nit: `[:first_delimiter_idx]`?

gcsfs/core.py
```diff
@@ -574,28 +581,149 @@ async def _list_objects(self, path, prefix="", versions=False):
...
-    ):
+        self, path, max_results=None, delimiter="/", 
+                    prefix="", versions=False, **kwargs):
+        
         """Object listing for the given {bucket}/{prefix}/ path."""
         bucket, _path, generation = self.split_path(path)
         _path = "" if not _path else _path.rstrip("/") + "/"
         prefix = f"{_path}{prefix}" or None
 
+        # Page size of 5000 is officially supported across GCS.
```
bernardhan33, line 607: nit: worth a reference here?

gcsfs/core.py
```diff
@@ -574,28 +581,149 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+                    prefix=prefix)
+            
+            use_snapshot_listing = inventory_report_info.get("use_snapshot_listing")
+
+            # If the user wants to rely on the snapshot from the inventory report
+            # for listing, directly return the results.
+            if use_snapshot_listing:
+                return items, prefixes
+            
+            # Otherwise, use the snapshot to initate concurrent listing.
```
bernardhan33: typo: initiate

gcsfs/core.py
```diff
@@ -574,28 +581,149 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+            items_from_process, prefixes_from_process = results[i]
+            items.extend(items_from_process)
+            prefixes.extend(prefixes_from_process)
+  
+        return items, prefixes
+    
+    async def _sequential_list_objects_helper(self, bucket, delimiter,
+            start_offset, end_offset, prefix, versions, generation, page_size):
+        """
+        Sequaltial list objects within the start and end offset range.
```
bernardhan33: typo: sequential

gcsfs/core.py
```diff
@@ -574,28 +581,149 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+                bucket=bucket,
+                delimiter=delimiter,
+                start_offset=None,
+                end_offset=None,
+                prefix=prefix,
+                versions=versions,
+                generation=generation,
+                page_size=default_page_size)
+        
+    async def _concurrent_list_objects_helper(self, items, bucket, delimiter,
```
bernardhan33: Sorry if I've missed it, could you point me to where the tests for `_concurrent_list_objects_helper` and `_sequential_list_objects_helper` are added?

Magichan33 (commented):
Hi all, this is Bernard Han from Google GCS Team. I'm Hans' co-host this summer. Leaving some general comments on this public PR.

Magichan33:
Also, is README changes in a separate commit or PR?

martindurant:
Everything passes, so that's a good start :)

I will start looking at this later today - there is quite a lot! Sorry for being slow getting here, got hit by covid.

hanseaston:
> Everything passes, so that's a good start :)
> 
> I will start looking at this later today - there is quite a lot! Sorry for being slow getting here, got hit by covid.

gcsfs/inventory_report.py
hanseaston: The naming convention is to conform to the JSON API here: https://cloud.google.com/storage/docs/json_api/v1/reportConfig/get, and it uses `location` as the naming.

gcsfs/inventory_report.py
hanseaston: fixed!

gcsfs/tests/test_inventory_report.py
hanseaston: fixed, thank you!

gcsfs/inventory_report.py
hanseaston: no, that's not a silly question, it's that I made a silly mistake, thank you!

gcsfs/tests/test_inventory_report.py
hanseaston: 1. Changed to using the `match` flag.
2. I guess we could do that, but since there is only 1 place so far doing this, setting a constant might stand out too much. I personally think it'd be better to conform with how the other tests are written ;)

gcsfs/inventory_report.py
hanseaston: done!

gcsfs/core.py
hanseaston: done!

gcsfs/core.py
hanseaston: done!

gcsfs/core.py
hanseaston: > nit: worth a reference here?

I believe this is already GA'ed. Users can verify by themselves that the HTTP  response is indeed returning 5000 objects per page, rather than 1000 objects per page. However, I think the public documentation is slightly outdated and have not updated this yet.

gcsfs/core.py
hanseaston: since `_concurrent_list_objects_helper` and `_sequential_list_objects_helper` are both helper methods in `core.py`, there is currently no explicit testing for the two methods. I was looking at `core.py`, and I see that the helper methods are not being unit tested specifically. However, the main methods are all integration tested in `test_core.py`.

There is the `test_inventory_report_listing` that does a simple integration test with the inventory report passed in. As for  `_sequential_list_objects_helper`, the function is basically the original `do_list_objects`, so I personally felt there is no need to test it again.

Let me know your thoughts @Magichan33.

hanseaston:
> Also, are README changes in a separate commit or PR?

I am thinking of adding a README change, and maybe a blog post on how to use the inventory report, once the PR is officially merged (in order to finalize all the details).

gcsfs/core.py
bernardhan33: I see. Sounds good to me! I'll let Martin have the final say on this though.

gcsfs/tests/test_inventory_report.py
martindurant: It's fine if we need to fix the test in the future, it forces us to consider why we are making that future change

gcsfs/core.py
martindurant: Where the higher function simply dispatches to two helper functions, it is fine not to write separate tests of the helper functions. The exception would be to make sure that invalid sets of arguments raise correctly, where the higher function should never send such arguments. I don't think this applies here.

gcsfs/core.py
martindurant: Maybe our own documentation (i.e., the prose) should have these facts and links to wherever they are mentioned, even if it's not official released user docs.

martindurant:
Yes, we certainly need prose "howto" documentation around this, in our own docs source. I would recommend any blog focus on the process of development and speedup for your example case rather that a how-to-use piece. Of course, the two can link to one-another.

I don't mind if the docs lag behind this PR, effectively making the new feature experimental. In fact, calling it an experimental feature isn't a bad idea, since we don't have a good feel of the set of use cases it might be pulled into, and there may be some that behave badly.

gcsfs/core.py
hanseaston: Sounds good to me!

hanseaston:
> Yes, we certainly need prose "howto" documentation around this, in our own docs source. I would recommend any blog focus on the process of development and speedup for your example case rather that a how-to-use piece. Of course, the two can link to one-another.

Sounds good, I will write up a prose (internally) and a blog (shared externally) once all the changes are finalized for this PR. Meanwhile, should I add a couple comments on top of the inventory report logic, indicating this is currently experimental?

hanseaston:
Update: I fixed the typos that Bernard has pointed out in the latest commit @Magichan33.

martindurant:
> should I add a couple comments on top of the inventory report logic, indicating this is currently experimental?

Yes, I think so

hanseaston:
> > should I add a couple comments on top of the inventory report logic, indicating this is currently experimental?
> 
> Yes, I think so

Sounds good, just updated that.

gcsfs/core.py
```diff
@@ -574,28 +582,170 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+                gcs_file_system=self,
+                inventory_report_info=inventory_report_info,
+                prefix=prefix,
+            )
+
+            use_snapshot_listing = inventory_report_info.get("use_snapshot_listing")
+
+            # If the user wants to rely on the snapshot from the inventory report
+            # for listing, directly return the results.
+            if use_snapshot_listing:
```
martindurant, line 625: I seem to recall that we talked about a config around using the listing if it is less than X seconds old. Here we have only a bool condition. This is fine, but wondering whether you plan on taking it further.

gcsfs/core.py
```diff
@@ -537,25 +538,32 @@ async def _get_object(self, path):
...
-                for c in clisting
-            )
-            if clisting and not hassubdirs:
-                return clisting
-        except FileNotFoundError:
-            # not finding a bucket in list of "my" buckets is OK
-            if key:
-                raise
+        # NOTE: the inventory report logic is experimental.
+        inventory_report_info = kwargs.get("inventory_report_info", None)
```
martindurant, line 547: In the future, these kwargs will have default values in instance attributes, so the user doesn't have to pass kwargs every time.

gcsfs/core.py
```diff
@@ -537,25 +538,32 @@ async def _get_object(self, path):
...
+
+        # Only attempt to list from the cache when the user does not use
```
martindurant, line 549: Does an inventory based listing get cached, if a later call is made without inventory-specific kwargs? 

Can we not use the normal expiry time allowed on the dircache to control whether we do another inventory lookup?

gcsfs/core.py
```diff
@@ -574,28 +582,170 @@ async def _list_objects(self, path, prefix="", versions=False):
             else:
                 return []
         out = pseudodirs + items
-        # Don't cache prefixed/partial listings
-        if not prefix:
+
+        use_snapshot_listing = (
+            False
+            if not inventory_report_info
+            else inventory_report_info.get("use_snapshot_listing")
```
martindurant, line 589:
```suggestion
        use_snapshot_listing = inventory_report_info and inventory_report_info.get("use_snapshot_listing")
```

gcsfs/core.py
```diff
@@ -574,28 +582,170 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+        )
+
+        # Don't cache prefixed/partial listings, in addition to
+        # not using the inventory report service to do listing directly.
+        if not prefix and use_snapshot_listing is False:
```
martindurant, line 594: Why not cache the snapshot listing?

gcsfs/core.py
```diff
@@ -574,28 +582,170 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+        self, items, bucket, delimiter, prefix, versions, generation, page_size
+    ):
+        """
+        Lists objects using coroutines, using the object names from the inventory
+        report to split up the ranges.
+        """
+
+        # Extract out the names of the objects fetched from the inventory report.
+        snapshot_object_names = [item["name"] for item in items]
+        snapshot_object_names = sorted(snapshot_object_names)
```
martindurant, line 663: Join these two lines

gcsfs/core.py
```diff
@@ -574,28 +582,170 @@ async def _list_objects(self, path, prefix="", versions=False):
...
+        # Determine the number of coroutines needed to concurrent listing.
+        # Ideally, want each coroutine to fetch a single page of objects.
+        num_coroutines = len(snapshot_object_names) // page_size + 1
+        num_objects_per_coroutine = len(snapshot_object_names) // num_coroutines
+
+        start_offsets = []
+        end_offsets = []
+
+        # Calculate the split splits of each coroutine (start offset and end offset).
+        for i in range(num_coroutines):
```
martindurant, line 674: I'm sure there's a simpler functools/itertools way to do this!

martindurant (commented):
Just small things left

gcsfs/core.py
hanseaston: Yes, the intention is to keep it simpler and more straightforward in this first PR. And after making the blog post, we can gather additional features that the user might want to support (including adding this config if a certain threshold has been met, x number of days, for example).

gcsfs/core.py
hanseaston: Okay it sounds good. Just making sure, nothing you recommend changing for now?

gcsfs/core.py
hanseaston: Yes, the inventory report listing results (if `use_snapshot_listing` is set to `False`, and thus it actually uses coroutines to fetch most up-to-date result) is cached, refer to the end of `list_objects` method.

When `use_snapshot_listing` is set to `True`, it is slightly complicated. Since the inventory reports might not contain all, but only potentially **a subset of** the object metadata (the user decides what object metadata they want to include the inventory report when they set up the inventory report in cloud console), I've decided not to add it to the cache, since that subset does not fully represent everything in the directory.

But maybe what we could do is - instead of not listing from the cache whenever the inventory report option is configured - we only use the cache in two cases 1) if inventory report option is not passed in 2) inventory report is passed in, but the user wants to get most up-to-date listing result.

So instead of the `if not inventory_report_info:` check, we do `if not inventory_report_info or not inventory_report_info["use_snapshot_listing"]. 

Thoughts?

gcsfs/core.py
hanseaston: Explained in the comment above. The snapshot listing might not contain all object metadata information, but only whatever subset of metadata the user wants to include.

gcsfs/core.py
hanseaston: will do!

gcsfs/core.py
hanseaston: will do!

gcsfs/core.py
hanseaston: will do!

gcsfs/core.py
martindurant: correct

gcsfs/core.py
martindurant: Leave it for now. I think there may be a use case for caching the manifest inventory and relying on it entirely, but we can see if anyone asks for that.

martindurant:
👍

PR #575, opened 2023-08-18
Add test_move_recursive

### LINKED ISSUE #489: mv and rename not working as expected
I create a folder in a bucket  "/my-bucket/folder/" which contains some text files. I then run

`fs.rename("/my-bucket/folder/", "/my-bucket/new_folder", recursive=True)`

Instead of seeing "/my-bucket/new_folder", I see "/my-bucket/new_foldermy-bucket/folder"

That doesn't seem correct. mv does the same thing.

Testing with fsspec and the localfilesystem give the expected result of there being just a "/folder/new_folder" that you would expect

ianthomas23:
Closes #489.

Issue #489 has been fixed by the recent improvements to recursive path expansions. This PR adds an explicit test to confirm this and ensure that any future regressions are caught.

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -399,6 +399,19 @@ def test_move(gcs):
     assert not gcs.exists(fn)
 
 
+@pytest.mark.parametrize("slash_from", ([False, True]))
+def test_move_recursive(gcs, slash_from):
+    # See issue #489
+    dir_from = TEST_BUCKET + "/nested"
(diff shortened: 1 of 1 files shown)


martindurant:
gcsfs/tests/test_core.py::test_gcs_glob now fails - any thoughts?

martindurant:
never mind, I think this was fixed in the follow-on ** PR

martindurant:
nope, still fails :|

PR #572, opened 2023-08-09
Better double asterisk support

john-jam:
This PR updates the `gcsfs` implementation to support updates made in the [Better double asterisks ** support](https://github.com/fsspec/filesystem_spec/pull/1329) pull request in the `filesystem_spec` repo.

Successful run: https://github.com/fsspec/filesystem_spec/actions/runs/5847462518/job/15915309794

@ianthomas23 Let me know if you prefer to merge your PR with `derived test` and I'll remove my updates regarding this.

PR #576, opened 2023-08-24
Avoid copying files with same name prefix when copying folder

john-jam:
As described [here](https://github.com/fsspec/filesystem_spec/pull/1340), when copying a folder, the files with the same name prefix are also copied.

This PR fixes this behavior. I guess there's probably a better way to fix this though.

TODO:
- [x] Merge `main` once [this PR](https://github.com/fsspec/gcsfs/pull/572) is merged so the derived tests are executed in this repo.
- [ ] Rerun the CI once [this PR](https://github.com/fsspec/filesystem_spec/pull/1340) is merged to highlight this fix.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -835,18 +835,6 @@ class GCSFileSystem(AsyncFileSystem):
         else:
             raise FileNotFoundError(path)
 
-    async def _glob(self, path, prefix="", **kwargs):
-        if not prefix:
-            # Identify pattern prefixes. Ripped from fsspec.spec.AbstractFileSystem.glob and matches
-            # the glob.has_magic patterns.
-            indstar = path.find("*") if path.find("*") >= 0 else len(path)
-            indques = path.find("?") if path.find("?") >= 0 else len(path)
-            indbrace = path.find("[") if path.find("[") >= 0 else len(path)
-
-            ind = min(indstar, indques, indbrace)
-            prefix = path[:ind].split("/")[-1]
-        return await super()._glob(path, prefix=prefix, **kwargs)
-
     async def _ls(self, path, detail=False, prefix="", versions=False, **kwargs):
         """List objects under the given '/{bucket}/{prefix} path."""
         path = self._strip_protocol(path).rstrip("/")
@@ -1251,6 +1239,9 @@ class GCSFileSystem(AsyncFileSystem):
             _prefix = f"{_path}{prefix}"
         else:
             _prefix = key
+
+        if _prefix != "" and await self._isdir(f"{bucket}/{_prefix}"):
+            _prefix = _prefix.rstrip("/") + "/"
 
         objects, _ = await self._do_list_objects(
             bucket, delimiter="", prefix=_prefix, versions=versions
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1387,23 +1387,26 @@ class GCSFileSystem(AsyncFileSystem):
         **kwargs,
     ):
         path = self._strip_protocol(path)
-        bucket, key, generation = self.split_path(path)
 
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
-        if prefix:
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -835,18 +835,6 @@ async def _info(self, path, generation=None, **kwargs):
...
-        if not prefix:
-            # Identify pattern prefixes. Ripped from fsspec.spec.AbstractFileSystem.glob and matches
-            # the glob.has_magic patterns.
-            indstar = path.find("*") if path.find("*") >= 0 else len(path)
-            indques = path.find("?") if path.find("?") >= 0 else len(path)
-            indbrace = path.find("[") if path.find("[") >= 0 else len(path)
-
-            ind = min(indstar, indques, indbrace)
-            prefix = path[:ind].split("/")[-1]
-        return await super()._glob(path, prefix=prefix, **kwargs)
```
jlanglois-jam, line 1002: If we apply the fix in the `_find` method below then this glob prefix construction no longer work for the edge case highlighted [here](https://github.com/fsspec/filesystem_spec/pull/1340/files#diff-1b7970a1cf6d8e169b2eab44ba722468be130bfeeef9deaed64e0d89882b88edR516).

This is because the the glob `subdir*` will create a `subdir` prefix and will be handled like a folder in the `_find` method (and add a trailing slash and miss the `subdir.txt` file).

Also, not sure to understand the initial benefit of this `_glob` method override. Was this just for caching?
Anyway, if another fix exists maybe this could be left as-is.

gcsfs/tests/test_core.py
```diff
@@ -285,10 +285,11 @@ def test_gcs_glob(gcs):
...
         if gcs.isfile(f)
     )
+    # the following is no longer true since the glob method list the root path
     # Ensure the glob only fetches prefixed folders
-    gcs.dircache.clear()
-    gcs.glob(TEST_BUCKET + "/nested**1")
-    assert all(d.startswith(TEST_BUCKET + "/nested") for d in gcs.dircache)
+    # gcs.dircache.clear()
+    # gcs.glob(TEST_BUCKET + "/nested**1")
+    # assert all(d.startswith(TEST_BUCKET + "/nested") for d in gcs.dircache)
```
jlanglois-jam, line 299: Without the `_glob` method override, `_find` is now called for the parent folder (`TEST_BUCKET`) so other folder like `test` are now in the dircache.

martindurant:
@ianthomas23 , if you have time

gcsfs/core.py
```diff
@@ -1410,6 +1398,9 @@ async def _find(
         else:
             _prefix = key
 
+        if _prefix != "" and await self._isdir(f"{bucket}/{_prefix}"):
```
martindurant: Is there no way to avoid isdir here? This doubles the number of calls for a simple list - and isdir surely is doing a list op anyway!

jlanglois-jam: You're right, it's not useful to double the calls there. That's why I initially indicated that a better solution should probably exist ;)

I updated the code with [this commit](https://github.com/fsspec/gcsfs/pull/576/commits/d1ff25f9870a5a09f287a68bd52183a699c41dfd) to try out another solution inspired by [the `_find` method in s3fs](https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L818:L823) which first fetches the path as if it is a directory then calls the `_info` method on the path to check if it's a file and returns it accordingly. In gcsfs, we can't call `_info` because some tests are expecting versioned files (`_info` does not support versioned files) or a prefixed search. So I simply moved the initial implementation into a block executed if the first search (when we treat path as a directory) does not return anything. It still doubles the calls when we are searching a single file though.

Let me know if that works for you. This commit is tested against [this PR](https://github.com/fsspec/filesystem_spec/pull/1340) to fix [this test](https://github.com/fsspec/filesystem_spec/actions/runs/5973337483/job/16205407770?pr=1340#step:6:484).

martindurant:
Will be releasing tomorrow :)

PR #577, opened 2023-08-24
Use utcnow time to test ctime info

john-jam:
When running tests from the future (in a zone at day+1 compared to UTC time), the `test_info` test fails with:
```
>       assert gcs.created(a).isoformat().startswith(today)
E       AssertionError: assert False
```

This fix uses the UTC date with `datetime.datetime.utcnow().date()` instead of `datetime.date.today()` since it has to be compared with `ctime` info that seems to be UTC as well.


martindurant:
How many times has this been run without hitting midnight :)

PR #586, opened 2023-10-06
Fix small transaction

### LINKED ISSUE #389: Uploading files fails when using transactions
**What happened**:
When trying to upload to upload a file with the following code and gcsfs version 2021.4.0 I get the following error:
```
gcsfs.utils.HttpError: Invalid request.  The number of bytes uploaded is required to be equal or greater than 262144, except for the final request (it's recommended to be the exact multiple of 262144).  The received request contained 21 bytes, which does not meet this requirement., 400
```

I would expect the file to be written to the bucket and no exception to be thrown.

```python

def main():
    file_system = GCSFileSystem(
            project="<our_project_name>", access="read_write"
        )
    path = "<our_test_bucket>/v15-test/test_file"

    file_system.start_transaction()


    with file_system.open(path, mode="wb") as f:
            f.write(b"This is a test string")

    file_system.end_transaction()


if __name__ == "__main__":
    main()
```

The upload succeeds when not using transactions. However we can't just remove the transaction in our production code as we run preemptible instances and want to be really sure that we have uploaded the files.

**Environment**:

- Dask version: None
- Python version: Python 3.9.2
- Operating System: Linux

martindurant:
Fixes #389 

It seems return core 499 is new behaviour on GCS upon upload cancel, and not supported on the emulator - so the test does not actually run in CI, but works against real GCS.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1736,7 +1736,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
             head = {}
             l = len(data)
 
-            if (l < GCS_MIN_BLOCK_SIZE) and not final:
+            if (l < GCS_MIN_BLOCK_SIZE) and (not final or not self.autocommit):
                 # either flush() was called, but we don't have enough to
                 # push, or we split a big upload, and have less left than one
                 # block.  If this is the final part, OK to violate those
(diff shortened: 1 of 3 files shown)


jonb377 (approved):
LGTM - I was able to verify the fix locally. Thanks Martin!

PR #589, opened 2023-11-02
Fix double asterisk glob in tests

mariosasko:
To avoid an error after https://github.com/fsspec/filesystem_spec/pull/1382


martindurant:
Thank you!

PR #591, opened 2023-11-27
Ensure consistent use of versions during pagination

### LINKED ISSUE #590: Error when listing large directory with versions=True
In some cases, listing a large directory with `versions=True` fails with a cryptic error. This only happens when results are paginated.

Here's an example:
```
>>> import gcsfs
>>> fs = gcsfs.GCSFileSystem(version_aware=True)
>>> ll = fs.ls("gcs://dql-50k-laion-files/000003/", versions=True)
_request non-retriable exception: Invalid argument., 400
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/retry.py", line 123, in retry_request
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 430, in _request
    validate_response(status, contents, path, args)
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/retry.py", line 110, in validate_response
    raise HttpError(error)
gcsfs.retry.HttpError: Invalid argument., 400
  File "<stdin>", line 1, in <module>
  File "/Users/ronan/.pyenv/versions/dql-3.10/lib/python3.10/site-packages/fsspec/asyn.py", line 118, in wrapper
  File "/Users/ronan/.pyenv/versions/dql-3.10/lib/python3.10/site-packages/fsspec/asyn.py", line 103, in sync
    raise return_result
  File "/Users/ronan/.pyenv/versions/dql-3.10/lib/python3.10/site-packages/fsspec/asyn.py", line 56, in _runner
    result[0] = await coro
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 998, in _ls
    for entry in await self._list_objects(
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 564, in _list_objects
    items, prefixes = await self._do_list_objects(
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 638, in _do_list_objects
    return await self._sequential_list_objects_helper(
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 753, in _sequential_list_objects_helper
    page = await self._call(
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 437, in _call
  File "/Users/ronan/.pyenv/versions/dql-3.10/lib/python3.10/site-packages/decorator.py", line 221, in fun
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/retry.py", line 158, in retry_request
    raise e
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/retry.py", line 123, in retry_request
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/core.py", line 430, in _request
    validate_response(status, contents, path, args)
  File "/Users/ronan/devel/iterative/gcsfs/gcsfs/retry.py", line 110, in validate_response
    raise HttpError(error)
gcsfs.retry.HttpError: Invalid argument., 400
```

rlamy:
Fixes #590.

I didn't add a test since I don't know how to write a sensible unit-test for paginated results.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -599,6 +599,7 @@ class GCSFileSystem(AsyncFileSystem):
         bucket, _path, generation = self.split_path(path)
         _path = "" if not _path else _path.rstrip("/") + "/"
         prefix = f"{_path}{prefix}" or None
+        versions = bool(versions or generation)
 
         # Page size of 5000 is officially supported across GCS.
         default_page_size = 5000
@@ -628,7 +629,6 @@ class GCSFileSystem(AsyncFileSystem):
                 delimiter=delimiter,
                 prefix=prefix,
                 versions=versions,
-                generation=generation,
                 page_size=default_page_size,
             )
 
@@ -642,12 +642,11 @@ class GCSFileSystem(AsyncFileSystem):
                 end_offset=None,
                 prefix=prefix,
                 versions=versions,
-                generation=generation,
                 page_size=default_page_size,
             )
 
     async def _concurrent_list_objects_helper(
-        self, items, bucket, delimiter, prefix, versions, generation, page_size
+        self, items, bucket, delimiter, prefix, versions, page_size
     ):
         """
         Lists objects using coroutines, using the object names from the inventory
@@ -696,7 +695,6 @@ class GCSFileSystem(AsyncFileSystem):
                     end_offset=end_offsets[i],
                     prefix=prefix,
                     versions=versions,
-                    generation=generation,
                     page_size=page_size,
                 )
                 for i in range(0, len(start_offsets))
@@ -722,7 +720,6 @@ class GCSFileSystem(AsyncFileSystem):
         end_offset,
         prefix,
         versions,
-        generation,
         page_size,
     ):
         """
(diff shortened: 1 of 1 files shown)


martindurant:
So that I understand what's going on here: listing of files should either be versioned or not, because it's not anticipated to be able to "list" a specific file generation? We usually do allow list of a specific path, though (would return a list of one item, essentially the same as info() ), so we need to make sure that that still works. Of course, such a listing with a specified generation would never be paginated.

To test, the [call](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax) accepts `max-keys`, so that could be set to a small number for the same of a test. I'm not immediately sure if that would need to be a monkeypatch. (having said that, pushing >1000 files to a local moto server shouldn't take that long at all)

rlamy:
The only use of `generation` is to decide whether to ask for versions. I guess it's indeed to support single-file `ls`. The issue seems to be that the page token may be rejected (depending on bucket settings, auth method, ...) if `versions` isn't set in the same way as in the first call.

I tried your suggestion for testing, but it looks like gcs-fake-server (this is GCS, not S3 :wink:) doesn't ever return paginated results.

rlamy:
Thank you! Any idea of when the next release will be?

martindurant:
> when the next release will be

As soon as we've sorted out the aiobotocore version pinning issue in s3fs (because they are always released together with fsspec)

PR #598, opened 2023-12-20
Add test to agree on desired aligned behavior of ls for empty directo…

hmeyer:
…ries.


martindurant:
You are very welcome to try this - thank you.
I believe it will take much more changes through the codebase to "treat zero-"/" files as directories" and also cope with the case where they do not exist.

PR #606, opened 2024-02-08
close_session: clean up loop handling

pgcamus:
Previously we were potentially running cleanup on a different
event loop, even if the session's associated event loop was
active. This manifested in RuntimeError instances being raised
when synchronous cleanup was attempted. Change cleanup to
always use the session's associated event loop and add some
logging when things have gone wrong.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -330,16 +330,24 @@ class GCSFileSystem(AsyncFileSystem):
     def project(self):
         return self.credentials.project
 
+    # Clean up the aiohttp session
+    #
+    # This can run from the main thread if invoked via the weakref callbcak.
+    # This can happen even if the `loop` parameter belongs to another thread
+    # (e.g. the fsspec IO worker). The control flow here is intended to attempt
+    # in-thread asynchronous cleanup first, then fallback to synchronous
+    # cleanup (which can handle cross-thread calls).
     @staticmethod
     def close_session(loop, session):
         if loop is not None and session is not None:
             if loop.is_running():
                 try:
-                    loop = asyncio.get_event_loop()
-                    loop.create_task(session.close())
+                    current_loop = asyncio.get_event_loop()
+                    current_loop.create_task(session.close())
                     return
                 except RuntimeError:
                     pass
+
                 try:
                     sync(loop, session.close, timeout=0.1)
                 except fsspec.FSTimeoutError:
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -342,7 +342,7 @@ class GCSFileSystem(AsyncFileSystem):
         if loop is not None and session is not None:
             if loop.is_running():
                 try:
-                    current_loop = asyncio.get_event_loop()
+                    current_loop = asyncio.get_running_loop()
                     current_loop.create_task(session.close())
                     return
                 except RuntimeError:


gcsfs/core.py
```diff
@@ -335,15 +335,18 @@ def close_session(loop, session):
         if loop is not None and session is not None:
             if loop.is_running():
                 try:
-                    loop = asyncio.get_event_loop()
```
martindurant, line 338: This is essentially "are we in the same thread as the loop".
I don't believe create_task should be done from other threads (this is what loop.call_soon_threadsafe or asyncio.run_coroutine_threadsafe are for, used by sync() just below)

pgcamus: A bit more detail on the symptom that I'm seeing

* We hit a `RuntimeError` in both `create_task()` and `sync()` because the `loop` (as returned by `asyncio.get_event_loop()` is closed.
* The `loop` parameter to the function is not the same as `asyncio.get_event_loop()` (calling `id()` on them yields different results).

Sorry, I'm not quite following the intent here. Is the goal to run `session.close` on the event loop of whatever the current thread is? If so, maybe we should guard things by checking running and closed status on the loop returned by `asyncio.get_event_loop()` instead of the function parameter. It's weird that we're checking one loop and then using a different one to do the work.

martindurant: The gcsfs instance should be used in async mode os sync mode, not both. If the former, we will always be within the event loop, it will be running and we will be on its thread. If sync, the loop should never be in the current thread.
Do you actually have multiple loops in your application?

martindurant: The exception is garbage collection, which happens on the main thread, and the original loop/thread may or may not be live

martindurant: ... and garbage collection maybe well happen during interpreter shutdown, when any given object attribute might have disappeared. I am not certain why the session really need to be shut down in such a case, but not doing so produces warnings.

pgcamus: Yes, this is happening in the weakref callback which is presumably invoked from GC.

This is what I see with fsspec 2023.9.2 and gcsfs 2023.9.2 (i.e. without my PR applied):

```
site-packages/gcsfs/core.py:342: RuntimeWarning: coroutine 'ClientSession.close' was never awaited

  pass

RuntimeWarning: Enable tracemalloc to get the object allocation traceback

  File "lib/python3.10/weakref.py", line 667, in _exitfunc

    f()

  File "lib/python3.10/weakref.py", line 591, in __call__

    return info.func(*info.args, **(info.kwargs or {}))

  File "site-packages/gcsfs/core.py", line 344, in close_session

    sync(loop, session.close, timeout=0.1)

  File "site-packages/fsspec/asyn.py", line 76, in sync

    raise RuntimeError("Loop is not running")

RuntimeError: Loop is not running
```

pgcamus: I am using the synchronous interface to fsspec

```
    def my_function():
        fs = fsspec.filesystem("gcs")
        fs.invalidate_cache()

        if not fs.exists(f"gs://..."):
            return

        for path in fs.glob(
            f"gs://.../*/*/*"
        ):
            # Do some stuff
```

pgcamus: Ok, how about this

Leave the `create_task` section running against the current thread's event loop (which can `RuntimeError` if there is none).

Change the `sync` section to run against the `loop` parameter as passed to this function.

martindurant: That sounds reasonable. It may also be worth digging how the two values end up different.

pgcamus: My guess --

* The `loop` parameter is the loop owned by the fsspec IO thread (our `GCSFileSystem` instance is created w/o an explicit `loop`, so we get the default).
* Because we're running in the weakref GC callback, we're in the main thread.

gcsfs/core.py
```diff
@@ -330,16 +330,24 @@ def base(self):
...
+    # cleanup (which can handle cross-thread calls).
     @staticmethod
     def close_session(loop, session):
         if loop is not None and session is not None:
             if loop.is_running():
                 try:
-                    loop = asyncio.get_event_loop()
-                    loop.create_task(session.close())
+                    current_loop = asyncio.get_event_loop()
+                    current_loop.create_task(session.close())
```
martindurant: So the difference is, that `loop` in `sync()` below is still the original. I don't see how this makes any difference, because the only way to avoid the first `return` is if `get_event_loop()` fails, and so no assign happens. Do you find there could be a RuntimeError in `create_task()` ?

pgcamus: Yeah. From what I can tell

1. The GC weakref callback runs in the main thread.
1. The main thread has no event loop, so when we call `get_event_loop()` it fails because it creates a new one which has never been started.
1. `create_task()` then fails because it refuses to create a task on a closed loop, and `BaseEventLoop._check_closed()` raises the exception.

With this fix, we don't see the exception anymore, but we do see the following, which is related but I don't think is caused by this change.

```
RuntimeWarning: coroutine 'ClientSession.close' was never awaited
```

```suggestion
                    current_loop = asyncio.get_event_loop()
                    if not current_loop.is_running():
                        raise RuntimeError
                    current_loop.create_task(session.close())
```

or

```suggestion
                    current_loop = asyncio.get_running_loop()
                    current_loop.create_task(session.close())
```

pgcamus: Ah, and I think this will fix the `RuntimeWarning` as well since we'll avoid creating the awaitable in the first place.

pgcamus: Yup, my test shows that this address both issues. PR update incoming.

PR #608, opened 2024-02-27
retry bulk rm

### LINKED ISSUE #558: Errors when deleting a directory with huge number of files
I'm trying to delete a path witth `fs.rm(some_path, recursive=True)` that has a huge number of files. gcsfs in general can deal with this well and finishes the delete in like 30 secs while `gsutil -m rm -r` takes almost an hour! However, I do this delete ocassionally in a long loop and eventually I see error below:

```
  File "/home/milad/micromamba/envs/foo/lib/python3.8/site-packages/fsspec/asyn.py", line 111, in wrapper
  File "/home/milad/micromamba/envs/foo/lib/python3.8/site-packages/fsspec/asyn.py", line 96, in sync
    raise return_result
  File "/home/milad/micromamba/envs/foo/lib/python3.8/site-packages/fsspec/asyn.py", line 53, in _runner
    result[0] = await coro
  File "/home/milad/micromamba/envs/foo/lib/python3.8/site-packages/gcsfs/core.py", line 987, in _rm
    raise exs[0]
  File "/home/milad/micromamba/envs/foo/lib/python3.8/site-packages/gcsfs/core.py", line 952, in _rm_files
    raise OSError(out)
OSError: {'We encountered an internal error. Please try again.'}
```

I also got this error once:

```
OSError: {"You have exceeded your bucket's IO capacity. Please follow the request rate and access distribution guidelines at https://cloud.google.com/storage/docs/hotspotting to scale your workload."}
```

Probably related to the exponential backoff that Google recommends? Is that implemented in gcsfs? I also tried changing the `batchsize` argument of `rm` but it didn't make a difference.

---
I'm using version `2022.7.1` if that's relevant

martindurant:
Fixes #558


slevang (approved):
This looks good - I tried it on my example in #558 and added some additional logging:

1) only a few random files end up in `remaining`
2) all the errors I'm seeing are 429
3) the failed files all go through on first retry, and the overall request then succeeds

Thanks!

martindurant:
I did some cleanup here, if people wouldn't mind trying again.

mil-ad:
Thanks @martindurant ! I'll try reproducing tomorrow

slevang:
On the latest commit I'm getting failures like this now:

```
  File "/opt/miniconda3/envs/salient/lib/python3.11/site-packages/zarr/storage.py", line 1531, in rmdir
    self.fs.rm(store_path, recursive=True)
  File "/opt/miniconda3/envs/salient/lib/python3.11/site-packages/fsspec/asyn.py", line 118, in wrapper
  File "/opt/miniconda3/envs/salient/lib/python3.11/site-packages/fsspec/asyn.py", line 103, in sync
    raise return_result
  File "/opt/miniconda3/envs/salient/lib/python3.11/site-packages/fsspec/asyn.py", line 56, in _runner
    result[0] = await coro
                ^^^^^^^^^^
  File "/opt/miniconda3/envs/salient/lib/python3.11/site-packages/gcsfs/core.py", line 1286, in _rm
    raise errors[0]
OSError: {    "code": 503,    "message": "Server is not responding",    "errors": [      {        "message": "Server is not responding",        "domain": "global",        "reason": "backendError"      }    ]  }

```

When I use the previous commit, today I am also getting 503 instead of 429 but the failed files are correctly filtered out and retried.

martindurant:
I made a small change to batch the leftovers, if any, instead of repeating for each batch.

I don't see how 503 is escaping, unless you are running out of retries (could add logging here):

```python
                    if code in [200, 204]:
                        out.append(path)
                    elif code in errs and i < 5:
                        remaining.append(path)
                    else:
                        ...
```
should mean 503 either gets the path back in `remaining` and is retried, so long as retries are left. Maybe the server really was flaky? `errs` contains [500, 501, 502, 503, 504, 408, 429].

slevang:
Of course I can't trip any errors now that I try again (regardless of commit). I'll try again tomorrow.

When I hit the 503 earlier, I did have a print statement that should have logged if anything went into `remaining` but I never saw that, suggesting it was not getting flagged as retriable somehow.

gcsfs/core.py
```diff
@@ -1177,47 +1180,63 @@ async def _rm_files(self, paths):
...
-                query = f"?generation={generation}" if generation else ""
-                parts.append(
-                    template.format(
-                        i=i + 1,
-                        bucket=quote(bucket),
-                        key=quote(key),
-                        query=query,
+        for retry in range(1, 6):
+            remaining = []
+            for chunk in _chunks(paths, 20):
```
slevang:
```suggestion
            for chunk in _chunks(paths, batchsize):
```

gcsfs/core.py
```diff
@@ -1177,47 +1180,63 @@ async def _rm_files(self, paths):
...
+                parents = set(self._parent(p) for p in paths) | set(paths)
+                [self.invalidate_cache(parent) for parent in parents]
+                txt = content.decode()
+                responses = txt.split(boundary)[1:-1]
+                for path, response in zip(paths, responses):
+                    m = re.search("HTTP/[0-9.]+ ([0-9]+)", response)
+                    code = int(m.groups()[0]) if m else None
+                    if code in [200, 204]:
+                        out.append(path)
+                    elif code in errs and i < 5:
```
slevang: Here's the problem:
```suggestion
                    elif code in errs and retry < 5:
```

slevang (changes requested):
Just had the wrong enumerator for the current retry state. With this change it seems to work well.

Any particular reason for dropping the `batchsize` to 20? You can apparently go as high as 2000, although I expect failures may become more likely if this is too big. Don't know if it makes any difference to speed, and this is way faster than `gsutil` already.

slevang:
On batch size, Google says this:
> You should not include more than 100 calls in a single batch request. If you need to make more calls than that, use multiple batch requests. The total batch request payload must be less than 10MB.

martindurant:
I was thinking that with the requests concurrent, smaller batches would be better. Maybe that's wrong, since sending the requests should take about the same bandwidth regardless.

We have two batch sizes here: in the outer _rm and the inner _rm_files. Should they be the same?

slevang:
I'm not sure but that seems ok. Then the outer `_rm` ends up handling all the batching and each call to `_rm_files` only runs a single chunk. So the code could actually be simplified to drop a loop from `_rm_files` right?

Also should a user be able to configure `batchsize` from the public API? [AbstractFileSystem.rm()](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L1190) doesn't pass through additional `**kwargs`.

martindurant:
Actually, it's this one: https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L316

And you are right, if _rm_files is not designed to be called from outside, there's no reason it should do its own internal looping.

slevang:
Thanks!

PR #609, opened 2024-02-28
Update docs about credentials

jdangerx:
Hi! I was trying to use a service account key with GCSFS, which wasn't mentioned in the docs - so I tried it, and it worked, and I think we can make it slightly clearer that it works.

Also, we ran into a silly issue where we had an unexpanded `~` in a path, and assumed that would work - adding some clarification there, too.

Finally, it took a surprising amount of googling to find the Credentials class (which I *knew* I had worked with before...) so I added a link as well.

I'm not sure if y'all have any RST style requirements but I'm happy to edit any of this to suit your needs!

FWIW - I didn't *actually* build the docs since a quick `make html` errored out with:

```
> make html 
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v7.2.6

Theme error:
no theme named 'sphinx_rtd_theme' found (missing theme.conf?)
make: *** [html] Error 2
```

And I decided that I should get back to my normal work ;)


martindurant:
Thanks for this. The range of possibilities allowed increased slowly with time, and the docs doesn't always keep up :|

PR #610, opened 2024-03-03
Add options to `mkdir`

slevang:
Noticed there was no way to specify a few common options on bucket creation via the IAM API.

I can modify the test if there is a better way, or remove entirely.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -837,8 +837,8 @@ class GCSFileSystem(AsyncFileSystem):
         location=None,
         create_parents=True,
         enable_versioning=False,
-        uniform_access=False,
-        public_access_prevention=True,
+        enable_object_retention=False,
+        iam_configuration=None,
         **kwargs,
     ):
         """
@@ -846,7 +846,7 @@ class GCSFileSystem(AsyncFileSystem):
 
         If path is more than just a bucket, will create bucket if create_parents=True;
         otherwise is a noop. If create_parents is False and bucket does not exist,
-        will produce FileNotFFoundError.
+        will produce FileNotFoundError.
 
         Parameters
         ----------
@@ -854,7 +854,8 @@ class GCSFileSystem(AsyncFileSystem):
             bucket name. If contains '/' (i.e., looks like subdir), will
             have no effect because GCS doesn't have real directories.
         acl: string, one of bACLs
-            access for the bucket itself
+            access for the bucket itself. See:
+            https://cloud.google.com/storage/docs/access-control/lists#predefined-acl
         default_acl: str, one of ACLs
             default ACL for objects created in this bucket
         location: Optional[str]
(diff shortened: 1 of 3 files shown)


gcsfs/tests/test_core.py
```diff
@@ -1478,3 +1479,16 @@ def test_find_maxdepth(gcs):
 
     with pytest.raises(ValueError, match="maxdepth must be at least 1"):
         gcs.find(f"{TEST_BUCKET}/nested", maxdepth=0)
+
+
+def test_mkdir_options(gcs):
+    gcs = GCSFileSystem(endpoint_url=gcs._endpoint)
+    if not gcs.on_google:
+        pytest.skip("emulator doesn't support IAM policies.")
```
martindurant: Instead, we should mock to make sure that at least the right IAM config dict is being sent

gcsfs/tests/test_core.py
```diff
@@ -1478,3 +1479,16 @@ def test_find_maxdepth(gcs):
 
     with pytest.raises(ValueError, match="maxdepth must be at least 1"):
         gcs.find(f"{TEST_BUCKET}/nested", maxdepth=0)
+
+
+def test_mkdir_options(gcs):
+    gcs = GCSFileSystem(endpoint_url=gcs._endpoint)
```
martindurant: Why make a new instance here?

gcsfs/tests/test_core.py
```diff
@@ -1478,3 +1479,16 @@ def test_find_maxdepth(gcs):
...
+    if not gcs.on_google:
+        pytest.skip("emulator doesn't support IAM policies.")
+
+    gcs.mkdir(TEST_CUSTOM_BUCKET, uniform_access=True, public_access_prevention=True)
+    info = gcs.info(TEST_CUSTOM_BUCKET)
+    gcs.rm(TEST_CUSTOM_BUCKET, recursive=True)
```
martindurant: This should be a fixture to make sure the bucket is cleaned up even on failure

gcsfs/core.py
```diff
@@ -865,6 +867,12 @@ async def _mkdir(
         enable_versioning: bool
             If True, creates the bucket in question with object versioning
             enabled.
+        uniform_access: bool
+            If True, creates the bucket in question with uniform access
```
martindurant: I don't know if most people will know what "uniform access" means

gcsfs/core.py
```diff
@@ -837,6 +837,8 @@ async def _mkdir(
         location=None,
         create_parents=True,
         enable_versioning=False,
+        uniform_access=False,
```
martindurant: I wonder how many kwargs we should pass before we start to make some Options dataclass

slevang: Fair. Would you be in favor of replicating the entire [Buckets:Insert](https://cloud.google.com/storage/docs/json_api/v1/buckets/insert) api? Or at least the [gcloud storage buckets create](https://cloud.google.com/sdk/gcloud/reference/storage/buckets/create) options? I wanted to add these in particular because they are the default (and recommended) options for buckets created through the console UI.

Looks like some others we are missing are class/autoclass, retention policy and requester pays. Plus some more niche options.

martindurant: How about we allow "iamConfiguration" as a kwarg, so the caller can put whatever they like in it? We could also other arguments that the API has - then it's just a JSON blob we pass along.

gcsfs/core.py
```diff
@@ -837,22 +837,25 @@ async def _mkdir(
         location=None,
         create_parents=True,
         enable_versioning=False,
+        enable_object_retention=False,
```
slevang, line 840: Added explicit handling for `enableObjectRetention` because it is the only other arg not passed through the request body, and `iamConfiguration` because it requires removing `acl` args to prevent a 400. Anything else we want can now be passed through `kwargs` to the request body.

slevang (commented):
I think this is a cleaner approach with the kwargs. `location` and `enable_versioning` also get passed to the json body but I didn't want to break these.

I got rid of the previous test but lmk if you have a particular way you would like these tested. I found the emulator and on_google split a little confusing to work with, especially for bucket creation.

Also added some more helpful info and links in the docstring.

gcsfs/core.py
martindurant: +1

PR #612, opened 2024-03-08
Ensure dircache on ls

### LINKED ISSUE #611: `fs.isdir` latency 200x slower beginning with version 2023.09.01
Hello,

I'm running the following code to measure the latency of `isdir`:

```
fs = GCSFileSystem(
    token="google_default", projectstring="..."
)

datadir = "<path to GCS directory with thousands of files>"
filenames = ["gs://" + x["name"] for x in fs.listdir(datadir)][:500]
times = []
for f in tqdm(filenames):
    begin = time.time()

    fs.isdir(f)

    end = time.time()
    times.append(end - begin)

print("Average time: ", sum(times) / len(times))
```

In version 2023.09.01, the average time per fs.isdir() call is 0.05 seconds. In version 2023.09.00, the average time is 0.0001 seconds. This causes a significant slowdown (from 2 seconds to several minutes) when multiplied by the thousands of files in our GCS directory. 

Thank you for your help,
Devin

martindurant:
Fixes #611 

I would still like to add a test to make sure this doesn't happen again, and also to double check what happens within find()

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -583,7 +583,6 @@ class GCSFileSystem(AsyncFileSystem):
             }
             for prefix in prefixes
         ]
-        breakpoint()
         if not (items + pseudodirs):
             if key:
                 return [await self._get_object(path)]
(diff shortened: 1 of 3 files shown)


martindurant:
@chotzen , I believe this is already enough to sort out your situation

PR #624, opened 2024-05-23
Add seek(0) to request data to prevent issues on retries

### LINKED ISSUE #621: Invalid multipart request on retry
~~_NOTE: This issue is eerily similar to #290 , which was fixed at the time by @nbren12 in #380 . I have the creeping suspicion that the same issue might have arisen again, though I have been unable to suss it out just yet._~~

### Problem description

Every now and then, a request stemming from `simple_upload()` will fail on its first attempt, for some retriable reason, but then the retry will receive a non-retriable HTTP 400 error: `Invalid multipart request with 0 mime parts.`. ~~As the error message suggests, it appears that the `Content-Type` header ([as required by the docs](https://developers.google.com/drive/api/guides/manage-uploads#http_1)) is missing (or empty?) on the retry, though it was evidently present on the initial attempt.~~ _**EDIT**: It seems I have misinterpreted the error message, and the issue is instead that the data stream is empty on the retry. I will keep remainder of the original post as-is for historical accuracy._

### Investigation

In my use case, the call originates from using the `xarray.Dataset.to_zarr()` function from `xarray` (which, interestingly, seems to be the use case for more people reporting similar issues. I suspect because the hard failure on the retry causes the upload to fail halfway though, leading to a corrupted Zarr store; but I digress). I have included the Traceback below, starting from the call to `to_zarr` for the sake of completion. The more relevant bit starts when we enter `fsspec` code with a call to `fsspec.mapping.setitems()` which takes a dict of `{path_as_str: data_as_bytes}` as only argument; This means it's a fairly basic call without potentially weird kwargs bleeding through and impacting the behaviour of fsspec/gcsfs, and so we can pretty much disregard anything related to xarray/zarr as a root cause. 

The most relevant code starts when we enter gcsfs with a call to `_pipe_file()` with only a `path` and `data` parameter, meaning the rest of the arguments are the defaults (most notably: `content_type="application/octet-stream"`). This leads to a call to `simple_upload()`, where the core of the request is created. The `content_type` parameter is used to generate a part of the body of the POST request, while the actual header of the call is set to `headers={"Content-Type": 'multipart/related; boundary="==0=="'}`. 

This leads to a call of `_request()` (via `_call`), where we see the first attempt failed (for some unknown but retriable reason), and the `retry_request` decorator from `gcsfs/retry.py` _supposedly_ performs the exact same `_request` call again, however this time it fails with the non-retriable `gcsfs.retry.HttpError: Invalid multipart request with 0 mime parts., 400`, suggesting that the retried `_request` did not have the same headers. However, I have failed to spot where this change might occur.

### Traceback:
```
(...)
  File "/usr/local/lib/python3.10/site-packages/xarray/backends/api.py", line 1832, in to_zarr
    dump_to_store(dataset, zstore, writer, encoding=encoding)
  File "/usr/local/lib/python3.10/site-packages/xarray/backends/api.py", line 1362, in dump_to_store
    store.store(variables, attrs, check_encoding, writer, unlimited_dims=unlimited_dims)
  File "/usr/local/lib/python3.10/site-packages/xarray/backends/zarr.py", line 657, in store
    self.set_variables(
  File "/usr/local/lib/python3.10/site-packages/xarray/backends/zarr.py", line 779, in set_variables
    writer.add(v.data, zarr_array, region)
  File "/usr/local/lib/python3.10/site-packages/xarray/backends/common.py", line 241, in add
    target[region] = source
  File "/usr/local/lib/python3.10/site-packages/zarr/core.py", line 1495, in __setitem__
    self.set_orthogonal_selection(pure_selection, value, fields=fields)
  File "/usr/local/lib/python3.10/site-packages/zarr/core.py", line 1684, in set_orthogonal_selection
    self._set_selection(indexer, value, fields=fields)
  File "/usr/local/lib/python3.10/site-packages/zarr/core.py", line 2058, in _set_selection
    self._chunk_setitems(lchunk_coords, lchunk_selection, chunk_values, fields=fields)
  File "/usr/local/lib/python3.10/site-packages/zarr/core.py", line 2261, in _chunk_setitems
    self.chunk_store.setitems(to_store)
  File "/usr/local/lib/python3.10/site-packages/zarr/storage.py", line 1441, in setitems
    self.map.setitems(values)
  File "/usr/local/lib/python3.10/site-packages/fsspec/mapping.py", line 124, in setitems
    self.fs.pipe(values)
  File "/usr/local/lib/python3.10/site-packages/fsspec/asyn.py", line 118, in wrapper
  File "/usr/local/lib/python3.10/site-packages/fsspec/asyn.py", line 103, in sync
    raise return_result
  File "/usr/local/lib/python3.10/site-packages/fsspec/asyn.py", line 56, in _runner
    result[0] = await coro
  File "/usr/local/lib/python3.10/site-packages/fsspec/asyn.py", line 399, in _pipe
    return await _run_coros_in_chunks(
  File "/usr/local/lib/python3.10/site-packages/fsspec/asyn.py", line 254, in _run_coros_in_chunks
    await asyncio.gather(*chunk, return_exceptions=return_exceptions),
  File "/usr/local/lib/python3.10/asyncio/tasks.py", line 408, in wait_for
    return await fut
  File "/usr/local/lib/python3.10/site-packages/gcsfs/core.py", line 1268, in _pipe_file
    location = await simple_upload(
  File "/usr/local/lib/python3.10/site-packages/gcsfs/core.py", line 1954, in simple_upload
    j = await fs._call(
  File "/usr/local/lib/python3.10/site-packages/gcsfs/core.py", line 437, in _call
  File "/usr/local/lib/python3.10/site-packages/decorator.py", line 221, in fun
  File "/usr/local/lib/python3.10/site-packages/gcsfs/retry.py", line 158, in retry_request
    raise e
  File "/usr/local/lib/python3.10/site-packages/gcsfs/retry.py", line 123, in retry_request
  File "/usr/local/lib/python3.10/site-packages/gcsfs/core.py", line 430, in _request
    validate_response(status, contents, path, args)
  File "/usr/local/lib/python3.10/site-packages/gcsfs/retry.py", line 112, in validate_response
    raise HttpError({"code": status, "message": msg})  # text-like
gcsfs.retry.HttpError: Invalid multipart request with 0 mime parts., 400
```

### Environment
- Python 3.10.13
- fsspec==2023.12.2
- gcsfs==2023.12.2.post1

Metamess:
Closes #621 

Credits go to @martindurant for the code.

When a simple_upload request fails but the `data` BytesIO stream was fully consumed, the retry attempt may fail because it does not send any data in the retried request. This PR attempt to fix that by always performing a `seek(0)` before starting the request.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -421,6 +421,8 @@ class GCSFileSystem(asyn.AsyncFileSystem):
         self, method, path, *args, headers=None, json=None, data=None, **kwargs
     ):
         await self._set_session()
+        if hasattr(data, "seek"):
+            data.seek(0)
         async with self.session.request(
             method=method,
             url=self._format_path(path, args),


rabernat:
Thanks for this fix @Metamess! We have hit the same bug! 🙏 

@martindurant would be wonderful to get a release asap. 🙏

Issue #627, opened 2024-06-27
Token refresh does not work


dbalabka:
The problem is related to #32. Unfortunately, it is not obvious to implement token refresh when a library is used by another library (dask -> pyarrow -> fsspec -> gcsfs). It would be amazing to implement more prone token refresh.

Here is an exception that we get when the job runs longer than an hour:
gcsfs version is `2023.12.2.post1`
`HttpError: Invalid Credentials, 401`
```
File /opt/conda/lib/python3.10/site-packages/dask_expr/_expr.py:3727, in _execute_task()

File /opt/conda/lib/python3.10/site-packages/dask/dataframe/io/parquet/core.py:97, in __call__()

File /opt/conda/lib/python3.10/site-packages/dask/dataframe/io/parquet/core.py:645, in read_parquet_part()

File /opt/conda/lib/python3.10/site-packages/dask/dataframe/io/parquet/core.py:646, in <listcomp>()

File /opt/conda/lib/python3.10/site-packages/dask/dataframe/io/parquet/arrow.py:641, in read_partition()

File /opt/conda/lib/python3.10/site-packages/dask/dataframe/io/parquet/arrow.py:1774, in _read_table()

File /opt/conda/lib/python3.10/site-packages/dask/dataframe/io/parquet/arrow.py:264, in _read_table_from_path()

File /opt/conda/lib/python3.10/site-packages/pyarrow/parquet/core.py:341, in __init__()

File ~/.../.venv/lib/python3.10/site-packages/pyarrow/_parquet.pyx:1250, in pyarrow._parquet.ParquetReader.open()
   1248 
   1249         with nogil:
-> 1250             check_status(builder.Open(self.rd_handle, properties, c_metadata))
   1251 
   1252         # Set up metadata

File ~/.../.venv/lib/python3.10/site-packages/pyarrow/types.pxi:88, in pyarrow.lib._datatype_to_pep3118()
     86 """
     87 try:
---> 88     char = _pep3118_type_map[type.id()]
     89 except KeyError:
     90     return None

File /opt/conda/lib/python3.10/site-packages/fsspec/spec.py:1844, in read()

File /opt/conda/lib/python3.10/site-packages/fsspec/caching.py:69, in _fetch()

File /opt/conda/lib/python3.10/site-packages/gcsfs/core.py:1850, in _fetch_range()

File /opt/conda/lib/python3.10/site-packages/fsspec/asyn.py:118, in wrapper()

[...]

PR #628, opened 2024-06-27
Remove deprecated poperty usage

dbalabka:
#627

### DIFF
--- a/gcsfs/credentials.py
+++ b/gcsfs/credentials.py
@@ -164,19 +164,19 @@ class GoogleCredentials:
         else:
             raise ValueError("Token format not understood")
         self.credentials = credentials
-        if self.credentials.valid:
+        if self.credentials.token_state == google.auth.credentials.TokenState.FRESH:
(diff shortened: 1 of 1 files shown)


martindurant:
> This change requires v2.26.0

Since that version is pretty new, we should support both ways rather than force everyone to upgrade. Can you please link where the new behaviour was introduced upstream, please (or a link to TokenState's doc) ?

dbalabka:
@martindurant , here is a PR: https://github.com/googleapis/google-auth-library-python/pull/1368

I'm not sure they have documentation yet. Also, let's skip this change for now unless it is required. I assumed it would fix my issue with a token refresh (see #627), but it seems there is no connection. Hence, we can close this PR.

martindurant:
>  let's skip this change for now unless it is required

I'm happy to leave it open for others to see, and we'll probably need it eventually.

martindurant:
A year has passed... what is the status of upstream use of this type now?

Issue #630, opened 2024-07-10
`AttributeError` when error is a string in `retry.py`


keunhong:
It seems like in rare cases GCS returns a string in the `"error"` field rather than a JSON object here: https://github.com/fsspec/gcsfs/blob/c7cf66d4b2ba1b5574813476283dace278f309c3/gcsfs/retry.py#L101

This causes an attribute error here: https://github.com/fsspec/gcsfs/blob/c7cf66d4b2ba1b5574813476283dace278f309c3/gcsfs/retry.py#L22 

```
File "[...]/lib/python3.10/site-packages/gcsfs/retry.py", line 21, in __init__
    self.code = error_response.get("code", None)
AttributeError: 'str' object has no attribute 'get'
```

This happens very rarely and I haven't been able to log exactly what is returned by Google in this case.

martindurant:
Would you like to make a PR with an `isinstance` on this line?

keunhong:
> Would you like to make a PR with an `isinstance` on this line?

Will do!

PR #631, opened 2024-07-18
Handle case when error message is a string.

keunhong:
Fixes https://github.com/fsspec/gcsfs/issues/630.


martindurant:
Thanks for getting this in.

Issue #233, opened 2020-01-22
read() returns incomplete file?


fernandobrito:
Hello,

This is my first time using the library, so please apologize me if I'm missing something obvious.

I'm trying to read a `.csv.gz` file (quite small, less than 100kb) following the documentation (https://gcsfs.readthedocs.io/en/latest/#examples), by doing:
```
fs = gcsfs.GCSFileSystem(project=project_id, token=service_key_path)
uri = bucket_name + '/' + filename
with fs.open(uri, 'rb') as f:
    content = f.read()
```

Here, I expect `content` to have the entire file, but it has only 168 rows (out of 946 in total). Reading the same file with `google-cloud-storage` works fine.

```
storage_client = storage.Client(credentials=credentials, project=project_id)
bucket = storage_client.bucket(bucket_name)
blob = bucket.blob(filename)
content = blob.download_as_string()
```

What am I missing? 
Thanks

martindurant:
That's certainly not what's supposed to happen, and not what happens in other people's experience. I cannot think of what might be happening, so any debugging and poking around the API you might do would be appreciated. Can you make the file for which this is happening public (or other similar file with random data) ?

(cc @TomAugspurger )

fernandobrito:
@martindurant Hey, thanks for the quick response! I had time to do some investigation this morning, and the issue that I described happens when I upload files to GCS bucket using `gsutil cp -Z` (for compression: https://cloud.google.com/storage/docs/gsutil/commands/cp#options_1)

Somehow, `gcsfs` is able to decompress the file correctly, as I can see its content, but it stops before the end of the file.

How to reproduce:

Run the script below (public bucket) and then change to `filename = dummy.csv`.
It's the same file in my computer, but one was uploaded with `gsutil cp -Z` and the one without the `-Z`.


```
from google.cloud import storage
import gcsfs

bucket_name = 'fernando-playground-temp'
filename = 'dummy.compressed.csv'

storage_client = storage.Client.create_anonymous_client()
bucket = storage_client.bucket(bucket_name)
blob = bucket.blob(filename)

content = blob.download_as_string()
print(len(content))
print(content[0:20])

print('--------')

fs = gcsfs.GCSFileSystem(token='anon')
uri = bucket_name + '/' + filename
with fs.open(uri, 'rb') as f:
    content = f.read()
    print(len(content))
    print(content[0:20])
```

TomAugspurger:
[...]

PR #635, opened 2024-08-08
Full support for decompressive transcoding (e.g. Content-Encoding: gzip ) during reading

### LINKED ISSUE #461: gcsfs doesn't properly handle gzipped files, ignoring content-encoding
I have a file "foo.txt.gz" that has been uploaded with the following metadata:
> Content-Type: text/plain
> Content-Encoding: gzip

I'm trying to copy its contents to a new file in cloud storage that is uncompressed to workaround a bug where my tooling (gcloud) can't properly handle gzip input.

If I try to pass the compression flag on read, it complains about the file not being a gzip file, implying that transcoding is occurring:
```python
with fs.open('gcs://jw-sandbox/uploads.txt.gz', 'rb', compression='gzip') as read_file:
...     with fs.open('gcs://jw-sandbox/uploads.txt', 'wb') as write_file:
...             shutil.copyfileobj(read_file, write_file)

gzip.BadGzipFile: Not a gzipped file (b'gs')
```

If I try to read the file without the compression flag and just dump contents to stdout, I only get the first N bytes of the decompressed contents where N is the compressed size:
```python
with fs.open('gcs://jw-sandbox/uploads.txt.gz', 'rb', compression='gzip') as read_file:
...     for f in read_file:
...             print(f)
```

```
>>> print(gcsfs.__version__)
2022.02.0
```

the-xentropy:
This fixes #461 and #233  without needing users to change their existing code.

The bugs were caused by assumptions in fsspec about the veracity and non-ambiguity of 'size' information returned by AbstractBufferedFile subclasses like GCSFile and AbstractFileSystem subclasses like GCSFileSystem  (e.g. `self.size = self.details["size"]` in `AbstractBufferedFile`, which is used by all base caches to truncate requests and responses). Since in GCS if compression-at-rest/compression transcoding is used there's no way to retrieve the real size of the object's *content* without decompressing the whole thing either server or client side, fixing these issues required overriding some behaviors in the underlying base classes. Care was taken to preserve behavior for storage objects not using compression at rest, however.

The fix keeps the data handling for GCS files which do not use compression at rest/compressive transcoding mostly identical by adding new control flow to detect when transcoding is done and adding some branch logic for handling those edge-cases. This did unfortunately mean implementing implementing variant methods to base classes with only minor changes to how they perform underlying operations (e.g. read() in GCSFile) which were previously just inherited from AbstractBufferedFile.

It does introduce one new semantic to GCSFs. [In line with fsspec's ArchiveFileSystem](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.archive.AbstractArchiveFileSystem.info) semantics, GCSFile will return size = None when the file can not be determined fully in advance. This allows us to distinguish known zero size and unknown size.

The only new performance overhead seen by non-users of compressive decoding is a single info() call resulting in a HEAD request done before the point where we create the GCSFile object in GCSFilesystem on file open()s, because we need to swap out the cache to one compatible with the lack of concrete file size but do not yet have the information to make that control flow decision. We can probably switch to the new transcoding cache wholesale in the future when we have faith it holds up to eliminate this call though, but I made it work this way to keep the data and control flow the same for the base case where users are not using compressive transcoding. Since the compression-at-rest case was completely broken, doing it this way means that even if these changes end up totally disastrous (it shouldn't though) it should only break something which is already broken.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -64,6 +64,85 @@ SUPPORTED_FIXED_KEY_METADATA = {
     "custom_time": "customTime",
 }
 
+class GCSTranscodingReadAheadCache(fsspec.caching.BaseCache):
+    """Cache which reads only when we get beyond a block of data, but with additional support for transcoding.
+
+    The default fsspec caches all *require* knowing the full file size ahead of time and will either error or truncate
+    aggressively if you don't provide meaningful values, so we implement our own cache which is mostly the same
+    but with special handling for the case where we can't know the file size ahead of time.
+
+    In practice, this is done by allowing end=None to mean "read to end". Although that may look strange,
+    In fsspec ArchiveFileSystem, size = None means the file size is undeterminable, so re-using those semantics and
+    making end mean 'read to end' for reading/caching it allows for clearly semantically meaningful syntax like
+    `some_file.read(length=some_file.size)` to work as expected without throwing errors, but still error if you try
+    something like some_file.size-10, since that's concretely an undeterminable value.
+
+    This is a much simpler version of BytesCache, and does not attempt to
+    fill holes in the cache or keep fragments alive. It is best suited to
+    many small reads in a sequential order (e.g., reading lines from a file).
+    """
+
+    name = "gcstranscodingreadahead"
+
+    def __init__(self, blocksize: int, fetcher: fsspec.caching.Fetcher, size: int) -> None:
+        super().__init__(blocksize, fetcher, size)
+        self.cache = b""
+        self.start = 0
+        self.end = 0
+
+    def _fetch(self, start: int | None, end: int | None) -> bytes:
+        """Fetches data from a remote source, rounding up to a minimum block size for the amount of data to fetch
+        at once. When requesting the full object, caching isn't performed"""
+        if start is None:
+            start = 0
+        # when we use the this cacher, end == None means "read to end"
+        # whenever end is not None, it means we're requesting a specific range, so we should honor it
+        if end is not None:
+            if start is None:
+                start = 0
+            if end is None or end > self.size:
+                end = self.size
+            if start >= self.size or start >= end:
+                return b""
+            l = end - start
+            if start >= self.start and end <= self.end:
+                # cache hit
+                self.hit_count += 1
+                return self.cache[start - self.start: end - self.start]
+            elif self.start <= start < self.end:
+                # partial hit
+                self.miss_count += 1
+                part = self.cache[start - self.start:]
+                l -= len(part)
+                start = self.end
+            else:
+                # miss
+                self.miss_count += 1
+                part = b""
+            end = min(self.size, end + self.blocksize)
+            self.total_requested_bytes += end - start
+            self.cache = self.fetcher(start, end)  # new block replaces old
+            self.start = start
+            self.end = self.start + len(self.cache)
+            return part + self.cache[:l]
+        else:
+            # end=None -> read to end
+            # this may look wasteful, but it's equivalent to the above
+            # in the case where end = file.size, which is very common (e.g. open() followed by read())
+            part = b""
+            if start >= self.start and self.cache:
+                # cache hit, we have at least some prefix we don't have to re-download
+                self.hit_count += 1
+                part = self.cache[start - self.start:]
+                start = self.end
+            self.cache = self.fetcher(start, None)
+            self.start = start
+            # since we're discarding previous data, we have to use the length of the cache to update self.end
+            self.end = self.start + len(self.cache)
+            return part + self.cache
+
+
+fsspec.caching.register_cache(GCSTranscodingReadAheadCache)
 
 def quote(s):
     """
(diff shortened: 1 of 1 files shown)


the-xentropy:
As an aside, this is my first pull request on GitHub to a broader community project, so let me know if there's anything I can do to polish this up.

the-xentropy:
Before we merge I think we should add some unit tests for the transcoding case and maybe beef up some test cases in the non-transcoding case (e.g. normally uploaded gzip files) to ensure we don't introduce some new issue or cause regressions in the future, but I could use some help in understanding how to develop for the test suite here, test them locally, test them against whatever dev buckets there are, etc.

gcsfs/core.py
```diff
@@ -1914,6 +2003,41 @@ def _simple_upload(self):
...
+        length: int (-1)
+            Number of bytes to read; if <0, all remaining bytes.
+        """
+        length = -1 if length is None else int(length)
+        if self.mode != "rb":
+            raise ValueError("File not in read mode")
+        if length < 0 and self.size is not None:
+            length = self.size - self.loc
+        if self.closed:
+            raise ValueError("I/O operation on closed file.")
```
manish, line 2023: Shouldn't this be at the top? If the file is closed, nothing else should be done

gcsfs/core.py
```diff
@@ -64,6 +64,85 @@
...
+        """Fetches data from a remote source, rounding up to a minimum block size for the amount of data to fetch
+        at once. When requesting the full object, caching isn't performed"""
+        if start is None:
+            start = 0
+        # when we use the this cacher, end == None means "read to end"
+        # whenever end is not None, it means we're requesting a specific range, so we should honor it
+        if end is not None:
+            if start is None:
+                start = 0
+            if end is None or end > self.size:
```
manish, line 103: Not sure if `end is None` will ever be true since in the outer block we mentioned `if end is not None`

gcsfs/core.py
```diff
@@ -64,6 +64,85 @@
...
+    something like some_file.size-10, since that's concretely an undeterminable value.
+
+    This is a much simpler version of BytesCache, and does not attempt to
+    fill holes in the cache or keep fragments alive. It is best suited to
+    many small reads in a sequential order (e.g., reading lines from a file).
+    """
+
+    name = "gcstranscodingreadahead"
+
+    def __init__(self, blocksize: int, fetcher: fsspec.caching.Fetcher, size: int) -> None:
```
manish, line 87: Can you provide some examples of how I can pass `fetcher`? I see it is of type.

```python
Fetcher = Callable[[int, int], bytes]  # Maps (start, end) to bytes
```

what implementation can one pass? I agree that having unit tests would be good.

gcsfs/core.py
```diff
@@ -64,6 +64,85 @@
...
+                start = 0
+            if end is None or end > self.size:
+                end = self.size
+            if start >= self.size or start >= end:
+                return b""
+            l = end - start
+            if start >= self.start and end <= self.end:
+                # cache hit
+                self.hit_count += 1
+                return self.cache[start - self.start: end - self.start]
```
manish, line 111: Okay, this took me some time to understand. So the cache only contains one contiguous segment and cannot hold more than one segment, as evidenced by `self.start` and `self.end`. Maybe it should be named `self.cache_start` and `self.cache_end`?

The cache holds a subset of data as a byte array and stores the cursor location within the actual content. Will it make sense to have multi-segment cache?

gcsfs/core.py
```diff
@@ -1914,6 +2003,41 @@ def _simple_upload(self):
...
+        handling for the case where we can not know the file size in advance (e.g. decompressive transcoding).
+
+        Parameters
+        ----------
+        length: int (-1)
+            Number of bytes to read; if <0, all remaining bytes.
+        """
+        length = -1 if length is None else int(length)
+        if self.mode != "rb":
+            raise ValueError("File not in read mode")
```
manish, line 2019: Good idea to provide more helpful error message

```python
raise ValueError(f"File '{self.path}' is not in read mode")
```

martindurant:
There is quite a lot of code here, and I haven't yet had a chance to look through it, sorry. 

> As an aside, this is my first pull request on GitHub to a broader community project, so let me know if there's anything I can do to polish this up.

Welcome, and thanks for getting involved.

One comment on caching: only readahead and none caching make sense, since we can only really stream from the start when using gzip.

Question: 
- when you issue bytes= range requests for compression encoded files, is this evaluated against the on-disk file? I assume, then, that anything not starting at zero would fail to decompress within aiohttp.
- is it possible to turn decompression off in the client, so that fsspec can do it (e.g., `fsspec.open("gcs://...", compression="gzip")` works fine for files that happen to be compressed, but without explicit transcoding set)?

the-xentropy:
I gotta look back into this a bit more extensively. I've noticed many subtle and not so subtle issues with my attempt at solving this and have to go back to the drawing board.

Thanks for your patience

PR #642, opened 2024-10-03
if-doesnt-exist condition on pipe()

martindurant:
Implements overwrite=False option in pipe/pipe_file (one-shot only, not multipart)

cc @TomAugspurger


martindurant:
ref: https://github.com/fsspec/filesystem_spec/issues/1693

PR #643, opened 2024-10-07
Remove race condition in credentials.py

flinz:
See https://github.com/fsspec/filesystem_spec/issues/565#issuecomment-2396993568

### DIFF
--- a/gcsfs/credentials.py
+++ b/gcsfs/credentials.py
@@ -180,7 +180,8 @@ class GoogleCredentials:
                     return  # repeat to avoid race (but don't want lock in common case)
                 logger.debug("GCS refresh")
                 self.credentials.refresh(req)
-                self.apply(self.heads)
+                # https://github.com/fsspec/filesystem_spec/issues/565
+                self.credentials.apply(self.heads)
 
     def apply(self, out):
         """Insert credential headers in-place to a dictionary"""


flinz:
@martindurant here is a branch with [your proposed fix](https://github.com/fsspec/filesystem_spec/issues/565#issuecomment-2396993568). 

The tests here pass, but I do see weird behavior when running an actual workload. Still double checking i am not messing things up on my side.

```
│   File "/usr/local/lib/python3.11/site-packages/fsspec/asyn.py", line 118, in wrapper                                                                                           │
│     return sync(self.loop, func, *args, **kwargs)                                                                                                                               │
│            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                               │
│   File "/usr/local/lib/python3.11/site-packages/fsspec/asyn.py", line 103, in sync                                                                                              │
│     raise return_result                                                                                                                                                         │
│   File "/usr/local/lib/python3.11/site-packages/fsspec/asyn.py", line 56, in _runner                                                                                            │
│     result[0] = await coro                                                                                                                                                      │
│                 ^^^^^^^^^^                                                                                                                                                      │
│   File "/usr/local/lib/python3.11/site-packages/gcsfs/core.py", line 1030, in _ls                                                                                               │
│     for entry in await self._list_objects(                                                                                                                                      │
│                  ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                      │
│   File "/usr/local/lib/python3.11/site-packages/gcsfs/core.py", line 563, in _list_objects                                                                                      │
│     clisting = self._ls_from_cache(path)                                                                                                                                        │
│                ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                        │
│   File "/usr/local/lib/python3.11/site-packages/fsspec/spec.py", line 379, in _ls_from_cache                                                                                    │
│     raise FileNotFoundError(path)
```

reineking:
I applied the fix in my run (writing hundreds of Parquet files with ~1b rows in total), and it works 🙂 Before the job was hanging consistently after 1h.

martindurant:
> I do see weird behavior when running an actual workload.

This causes the workflow to fail? Are you running multiple threads too? Missing credentials should not cause FileNotFound but PermissionError, and the traceback indicates this is coming from the directory listing cache - doing reading?

flinz:
> > I do see weird behavior when running an actual workload.
> 
> This causes the workflow to fail? Are you running multiple threads too? Missing credentials should not cause FileNotFound but PermissionError, and the traceback indicates this is coming from the directory listing cache - doing reading?

Yes this causes a failure. We're not running with multiple threads, it's the first (reading) call in the process, which uses `GCSFilesystem().ls(path)`.

flinz:
@martindurant: I now call `ls(path, refresh=True)` and this works as intended. Preliminary runs confirm that I do not get the race condition/hanging anymore. So, as to this bugfix, I think we're good.

The only slight doubt that remains, is why this caching behavior didn't appear before.

> So I understand clearly: `ls` (alone and with no other operations) works with current main, but not with this patch?

What I do exactly (in pseudocode) is:

- `gcsfs.GCSFileSystem().exists(parent_path)` is called earlier
-  extra code creates new paths `path/subdir1/subdir2` with content in `subdir1` and `subdir2`
- `gcsfs.GCSFileSystem().ls(path/subdir1/subdir2)` is called later, and this is the one that fails, only after the patch.

One important point: I did notice that pre-patch I ran `2023.12.2.post1`, which didn't contain https://github.com/fsspec/gcsfs/pull/612. So, if the cache persists somehow between instantiations of `GCSFileSystem`, then that could be an explanation for the failure of the `ls`.

martindurant:
`gcsfs.GCSFileSystem()` will get the *same instance* each time, by design, since creating credentials and connections is expensive. It will also persist the directory cache. You can avoid this by passing `skip_instance_cache=True`, but it would be worth finding out what's going on.

> extra code creates new paths

If this is from a different instance/process/machine, then it is to be expected that the directory cache of the instance above has become stale, so `refresh` is the right thing to do (or `.invalidate_cache()`).

flinz:
> gcsfs.GCSFileSystem() will get the same instance each time, by design, since creating credentials and connections is expensive. It will also persist the directory cache. You can avoid this by passing skip_instance_cache=True, but it would be worth finding out what's going on.

Thats very useful information @martindurant ! I was trying to find something along these lines in the documentation, but failed to. If this is not in there yet, would it make sense adding it?

martindurant:
It is an inherited behaviour from fsspec, mentioned here: https://filesystem-spec.readthedocs.io/en/latest/features.html#instance-caching

I totally agree that the whole project and set of repos could do with a docs reorganise, but I think it's beyond my personal effort level right now.

Issue #645, opened 2024-10-24
GCSFileSystem doesn't properly handle cache_timeout = 0


amyewang:
When using `GCSFileSystem`, I would expect passing in `cache_timeout = 0` to result in no caching, since according to the docs: `Set cache_timeout <= 0 for no caching, None for no cache expiration.` Yet when `cache_timeout = 0`, `self.cache_timeout` is set to `None` and caching is being used. 

When `cache_timeout = 0`, [`self.cache_timeout = cache_timeout or kwargs.pop("listings_expiry_time", None)`](https://github.com/fsspec/gcsfs/blob/358ba237b874ac1d03602019407133a9a30daa77/gcsfs/core.py#L304) is evaluating to `None`, since the value of `0` is falsy. Currently circumventing this by passing in `cache_timeout = -1`, but docs state that a value of `0` should be supported and indicate no caching.

martindurant:
OK, so the docstring and the functionality need to be brought into line - can you please make the PR? Ideally we should have the same behaviour between implementations, so the change might be in fsspec.dircache . I don't mind which sentinel value means "no caching", especially since we also have `use_listings_cache`. I suppose an expiry time of 0 could be understood as no caching, or as no expiry (i.e., use cache and never make stale).

PR #646, opened 2024-10-30
Handle cache_timeout = 0 in GCSFileSystem

amyewang:
Fixes https://github.com/fsspec/gcsfs/issues/645

### DIFF
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -1428,6 +1428,8 @@ def test_expiry_keyword():
     assert gcs.dircache.listings_expiry_time == 1
     gcs = GCSFileSystem(cache_timeout=1, token="anon")
     assert gcs.dircache.listings_expiry_time == 1
+    gcs = GCSFileSystem(cache_timeout=0, token="anon")
+    assert gcs.dircache.listings_expiry_time == 0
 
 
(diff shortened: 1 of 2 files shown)


martindurant:
A test for this would be nice.

amyewang:
@martindurant added a test to `test_expiry_keyword` & passes as expected. let me know if you think it should live somewhere else!

gcsfs/tests/test_core.py
```diff
@@ -1428,6 +1428,8 @@ def test_expiry_keyword():
     assert gcs.dircache.listings_expiry_time == 1
     gcs = GCSFileSystem(cache_timeout=1, token="anon")
     assert gcs.dircache.listings_expiry_time == 1
+    gcs = GCSFileSystem(cache_timeout=0, token="anon")
+    assert gcs.dircache.listings_expiry_time == 0
```
martindurant, line 1432: Does this imply that the cache is disabled? What was the previous value before this PR, None?

amyewang: Yes, based on the docstring the `listings_expiry_time` should be set to 0, which means cache has a timeout of 0. Before this change the `gcs.dircache.listings_expiry_time` evaluated to `None`, so the cache was enabled

Issue #654, opened 2024-12-31
Guess correct mimetype instead of defaulting to application/octet-stream


kujenga:
Currently this implementation has a hard-coded default content type of `application/octet-stream` if the caller does not specify a content type: https://github.com/fsspec/gcsfs/blob/e3ebfb9531aa8c5486b468e5c90e39aca9d80f95/gcsfs/core.py#L1336

The S3 implementation takes a more intelligent approach and guesses the mime type when it is not specified: https://github.com/fsspec/s3fs/blob/51e3c80ef380a82081a171de652e2b699753be2b/s3fs/core.py#L1237-L1240

The idea behind this issue is to do the same thing in this client. This should provide more correct default behavior without requiring more work on the part of users of this library, e.g. it will improve the ability to view files directly in the browser from GCS.

PR #655, opened 2025-01-02
Guess file content type when not specified

kujenga:
This commit improves the default content type behavior for this library by guessing the correct mime type when one is not provided. This improves the accuracy of the default behavior in representing the files and matches other fsspec implementations.

Closes https://github.com/fsspec/gcsfs/issues/654


martindurant:
I must admit to not being particularly fond of this kind of guessing, but since the user can always override, I suppose it's OK.

PR #673, opened 2025-04-09
add userProject to batch deletion query

### LINKED ISSUE #672: Error on batch version of rm when in requester pays mode
When a `rm` operation is performed, including indirectly through a `mv`,  the optimized batched version of the command is executed (property `on_google=True`).
However, possibly due to updates on google sides, in case of `requester_pays=True` mode Google complains that the userProject is not provided for billing. I debugged and checked it was there, but the batch version does weird stuff with the header so I don't really know what's going on. 
Anyway I get this error:
```
OSError: {    "code": 400,    "message": "Bucket is a requester pays bucket but no user project provided.",    "errors": [      {        "message": "Bucket is a requester pays bucket but no user project provided.",        "domain": "global",        "reason": "required"      }    ]  }
```
If I change the code to avoid the `_rm_files` method (setting the property `on_google=False` by hand) it works.

Seems like a bug, due to changes on google sides. Here to let you know.
Thanks for your work!

GuidoBallabio:
Fixes #672

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1245,7 +1245,14 @@ class GCSFileSystem(asyn.AsyncFileSystem):
             parts = []
             for i, p in enumerate(chunk):
                 bucket, key, generation = self.split_path(p)
-                query = f"?generation={generation}" if generation else ""
+                query_params = self._get_params(
+                    {"generation": generation} if generation else {}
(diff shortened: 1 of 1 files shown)


martindurant:
Please apply the linting suggestion

Issue #533, opened 2023-03-08
trigger a warning when trying to append to a file on google cloud storage


Courvoisier13:
When trying to append to a file on GCS using pandas, the file was being overwritten. That is because GCS files are immutable. But there was no warning and it took a while to debug.

The `mode = 'a'` in `pandas.to_csv` is passed on to `fsspec.open`. It would be nice if we have a warning or error in this case. 
See https://github.com/pandas-dev/pandas/issues/51821

```
import pandas as pd
df = pd.DataFrame({
    'account-start': ['2017-02-03', '2017-03-03', '2017-01-01'],
    'client': ['Alice Anders', 'Bob Baker', 'Charlie Chaplin'],
    'balance': [-1432.32, 10.43, 30000.00],
    'db-id': [1234, 2424, 251],
    'proxy-id': [525, 1525, 2542],
    'rank': [52, 525, 32],
    ...
})
header = True
to_csv_mode = 'w'
with pd.read_csv(gs_path, chunksize=1) as reader:
    for r in reader:
        r.to_csv(temp_gs_path, index=False, header=header, mode=to_csv_mode)
        header = False
        to_csv_mode = 'a'
```

martindurant:
(transferred to gcsfs)

martindurant:
Append in GCS is possible, via [compose](https://cloud.google.com/storage/docs/json_api/v1/objects/compose) (currently used by the `merge()` method), where we upload the extra data to some scratch file, and when close/committing, do a merge to rewrite the final destination. S3fs does something similar, except that in that case there is no need to write the new data to an actual key which then needs to be removed when done.

@Courvoisier13 , are you interested in working on this?

Courvoisier13:
I've built indeed my own solution using `compose`. I'm a bit swamped right now. Maybe I can come back when i have more time.

martindurant:
It would be very helpful!

PR #675, opened 2025-04-22
Create warning for appending mode 'a' operations

tassilucas:
Add specific warning for files in appending mode since gcsfs ignores `mode='a'` as stated in issue #533.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1825,7 +1825,9 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
         self.checker = get_consistency_checker(consistency)
 
         if "a" in self.mode:
-            warnings.warn("Append mode 'a' is not supported in GCS. Using overwrite mode instead.")
+            warnings.warn(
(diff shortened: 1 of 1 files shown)


gcsfs/core.py
```diff
@@ -1824,6 +1824,9 @@ def __init__(
         self.acl = acl
         self.checker = get_consistency_checker(consistency)
 
+        if "a" in self.mode:
+            warnings.warn("Append mode 'a' is not supported in GCS. Using overwrite mode instead.")
```
martindurant: Should actually replace the mode, no?
```suggestion
            warnings.warn("Append mode 'a' is not supported in GCS. Using overwrite mode instead.")
            self.mode = self.mode.replace("a", "w")
```

martindurant (commented):
Thanks for the suggestion - it is reasonable.

I wonder whether someone would like to try to implement append mode via [compose](https://cloud.google.com/storage/docs/json_api/v1/objects/compose) , i.e., write new data to a temporary file and then join the parts.
For small initial files (< block_size), they can simply be read and re-written.

Issue #312, opened 2020-11-27
Pseudo folders break directory detection code in fs.info 


dynamix:
Various tools/libraries create pseudo folders in GS, for example Tensorflow or using the "Create folder" function in the Google Storage Browser UI.   

The code that identifies directories does not handle this properly: https://github.com/remerge/gcsfs/blob/be50a9a8d23a04529c40540a1cc123ce5b3512fc/gcsfs/core.py#L819-L820

The folder name will be an exact match and assumed to be a file on the first invocation. This breaks various other methods like `isdir`. 

How to reproduce:

```Python
import gcsfs
import tensorflow as tf

path = "YOUR-BUCKET/foo"

tf.io.gfile.mkdir(f"gs://{path}")

fs = gcsfs.GCSFileSystem()
print(fs.isdir(path)) # => False
print(fs.isdir(path)) # => True

```

I'll create a PR tomorrow to fix this.

martindurant:
I would argue that `path` is indeed a file because ... it is a file, albeit empty. Directories are only ever implied by their contents, since their names are returned as "common prefixes", i.e., a different mechanism (and the names will not contain "/" at the end). 
What were to happen if of the folder placeholders were to contain data?

dynamix:
Yes that is what I meant by pseudo folder. The issue here is that what `info` returns is not consistent due to trying to support the idea of folders. In the example `isdir` should be idempotent but it does returns `false` on the first invocation and `true` on the second. 

The pseudo folders are usually indicated by a zero size objects ending with `/`.  Detecting this would be my suggested fix: https://github.com/dask/gcsfs/pull/313

martindurant:
> In the example isdir should be idempotent but it does returns false on the first invocation and true on the second.

Totally agree that this is wrong, and there should be a test that ensures that repeated calls get the same result.

isidentical:
I also agree consistency is the important part.

bilelomrani1:
Just wanted to check if there is any news on this issue?

martindurant:
This is likely fixed in the current release. Are you still experiencing a failing case, @bilelomrani1 ?

bilelomrani1:
I apologize for the confusion caused by my previous comment. After further investigation, I realized that the issue I encountered was not related to what I initially thought. Instead, I discovered that the problem stemmed from the fact that `fs.makedirs` behaves as a no-op with `gcsfs`, which is a documented behavior. Sorry for the inconvenience.

[...]

PR #676, opened 2025-04-24
Yet another fix for isdir

Bjoern-Rapp:
This is yet another stab at a fix for #312. 
Empty files that ends with "/" are treated as empty directories in "ls" and "info".
Test requsted in #313 is added.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1058,6 +1058,7 @@ class GCSFileSystem(asyn.AsyncFileSystem):
             out = await self._list_buckets()
         else:
             out = []
+            dir_names = set()
             for entry in await self._list_objects(
                 path, prefix=prefix, versions=versions, **kwargs
             ):
@@ -1069,8 +1070,11 @@ class GCSFileSystem(asyn.AsyncFileSystem):
                         "storageClass": "DIRECTORY",
                         "type": "directory",
                     }
-                    if entry in out:
+
+                if entry["type"] == "directory":
+                    if entry["name"] in dir_names:
                         continue
+                    dir_names.add(entry["name"])
 
                 if versions and "generation" in entry:
                     entry = entry.copy()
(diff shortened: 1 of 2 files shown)


martindurant:
This seems to break the batch copy tests. It is plausible, but probably unlikely, that the tests are incorrect - see https://filesystem-spec.readthedocs.io/en/latest/copying.html

Bjoern-Rapp:
Sorry, I was not able to run any of the "derived" tests on Windows. Will investegate.

Bjoern-Rapp:
The tests are correct, and should all pass now.

gcsfs/core.py
```diff
@@ -1060,13 +1061,32 @@ async def _ls(
...
             ):
+                if entry["size"] == 0 and entry["name"].endswith("/"):
+                    entry = {
+                        "bucket": entry["bucket"],
+                        "name": path.rstrip("/"),
+                        "size": 0,
+                        "storageClass": "DIRECTORY",
+                        "type": "directory",
+                    }
+                    if entry in out:
```
martindurant: Might this comparison of a dict to all the dicts so far be expensive?

gcsfs/core.py
```diff
@@ -1060,13 +1061,32 @@ async def _ls(
...
                     entry = entry.copy()
                     entry["name"] = f"{entry['name']}#{entry['generation']}"
+
                 out.append(entry)
 
         if detail:
-            return out
+            return sorted(
+                out,
+                key=lambda e: (
```
martindurant: You want to show directories first? I think that's not the case in any other fsspec ls, it should be simple sorting.

martindurant (commented):
OK, looks good. I just have a couple of questions.

gcsfs/core.py
Bjoern-Rapp: You previously commented that ´info(path)´ and ´ls(path)[0]´ should be equal.
Showing the directories first was an attempt at that insuring that.
But that’s not the case anyway, when a file shadows a “directory". If you have the files foo/bas and foo/bas/ ´info(path)´ shows type=file and´ls(path)[0]´ shows type=directory
Ordering by name is perhaps the simple and correct solution here.

gcsfs/core.py
Bjoern-Rapp: Yes, that is expensive. The check is necessary, directories should not be listed twice when there exists both a placeholder file and a file under that prefix. Directories is not hashable. I could make a temporary set of directory paths to check against?

martindurant: > I could make a temporary set of directory paths to check against?

Yes, I'm afraid we probably need that

gcsfs/core.py
martindurant: +1

Bjoern-Rapp:
Does this look good now?

PR #677, opened 2025-04-28
Moved config and dependency to pyproject.toml

Bjoern-Rapp:
Set up a PEP 621 compliant pyproject.toml file, and move the project and build config from setup.py, setup.cfg and isort.cfg, and move dependencies from requirements.txt.

This PR does not attempt to change any other aspect around the build system. Modern versions of Setuptools is able to read config from the pyproject.toml file. The resulting build should have minimale differences.

This PR does add "dev" dependency group, so that contributes can choose to set a virtual environment with a PEP 735 compatible package manger, in addition to Conda.


martindurant:
Thanks for doing this work.

I notice some back and forth around versioneer. If you would like to remove it in favour of setuptools-scm, I would be OK with that.

martindurant:
(would be happy to review a more up to date version of this)

PR #679, opened 2025-05-04
Add support for specifying Cloud KMS keys when creating files

kaiju:
Allows for creating files with a specific Cloud KMS key by passing a fully qualified resource path to the `kms_key_name` keyword argument.

I've attempted to add some basic tests for this functionality. `fake-gcs-server` does not seem to support KMS, so they're skipped unless running against the live API.

This is my first contribution to this project and I'm happy to address any concerns :)

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1791,6 +1791,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
         content_type=None,
         timeout=None,
         fixed_key_metadata=None,
+        kms_key_name=None,
         generation=None,
         **kwargs,
     ):
@@ -1830,6 +1831,9 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
                 - custom_time
             More info:
             https://cloud.google.com/storage/docs/metadata#mutable
+        kms_key_name: str
+            Resource name of the Cloud KMS key that will be used to encrypt
+            the object.
         timeout: int
             Timeout seconds for the asynchronous callback.
         generation: str
@@ -1866,6 +1870,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
         self.metadata = metadata or det.get("metadata", {})
         self.fixed_key_metadata = _convert_fixed_key_metadata(det, from_google=True)
         self.fixed_key_metadata.update(fixed_key_metadata or {})
+        self.kms_key_name = kms_key_name
         self.timeout = timeout
         if mode in {"wb", "xb"}:
             if self.blocksize < GCS_MIN_BLOCK_SIZE:
@@ -1973,6 +1978,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
             self.content_type,
             self.metadata,
             self.fixed_key_metadata,
+            self.kms_key_name,
             mode="create" if "x" in self.mode else "overwrite",
             timeout=self.timeout,
         )
@@ -2006,6 +2012,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
             self.consistency,
             self.content_type,
             self.fixed_key_metadata,
+            self.kms_key_name,
             mode="create" if "x" in self.mode else "overwrite",
             timeout=self.timeout,
         )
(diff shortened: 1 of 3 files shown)


gcsfs/core.py
```diff
@@ -1791,6 +1791,7 @@ def __init__(
         content_type=None,
         timeout=None,
         fixed_key_metadata=None,
+        kms_key_name=None,
```
martindurant: Here and in the other signatures, it is probably better to add the extra key in the last place, just in case someone is depending on the order. We should probably enforce these to be keyword only at some point, but until then...

gcsfs/core.py
```diff
@@ -1830,6 +1831,9 @@ def __init__(
                 - custom_time
             More info:
             https://cloud.google.com/storage/docs/metadata#mutable
+        kms_key_name: str
+            Resource name of the Cloud KMS key that will be used to encrypt
+            the object.
```
martindurant, line 1836: Perhaps a lint to how to make these?

gcsfs/tests/settings.py
```diff
@@ -3,3 +3,7 @@
 TEST_BUCKET = os.getenv("GCSFS_TEST_BUCKET", "gcsfs_test")
 TEST_PROJECT = os.getenv("GCSFS_TEST_PROJECT", "project")
 TEST_REQUESTER_PAYS_BUCKET = "gcsfs_test_req_pay"
+TEST_KMS_KEY = os.getenv(
+    "GCSFS_TEST_KMS_KEY",
+    f"projects/{TEST_PROJECT}/locations/us/keyRings/gcsfs_test/cryptKeys/gcsfs_test_key",
```
martindurant, line 8: "gcsfs_test" is from the TEST_BUCKET, or is this unrelated?

gcsfs/tests/test_core.py
```diff
@@ -87,6 +88,16 @@ def test_simple_upload(gcs):
     assert gcs.cat(fn) == b"zz"
 
 
+def test_simple_upload_with_kms(gcs):
+    if not gcs.on_google:
+        pytest.skip("emulator does not support kmsKeyName")
```
martindurant, line 93: Feel free to ask on their repo whether this is planned

gcsfs/tests/test_core.py
```diff
@@ -100,6 +111,24 @@ def test_large_upload(gcs):
         gcsfs.core.GCS_MAX_BLOCK_SIZE = orig
 
 
+def test_large_upload_with_kms(gcs):
+    if not gcs.on_google:
+        pytest.skip("emulator does not support kmsKeyName")
+    orig = gcsfs.core.GCS_MAX_BLOCK_SIZE
+    gcsfs.core.GCS_MAX_BLOCK_SIZE = 262144  # minimum block size
```
martindurant, line 118: This is like a monkeypatch or fixture - should the test fail, it might trigger other failures as the value will not have been reset. 
Can the same be achieved by using block_size in open() (that seems to be what happens int he following test)?

gcsfs/tests/test_core.py
```diff
@@ -136,6 +165,53 @@ def test_multi_upload(gcs):
...
+    d = b"01234567" * 2**15
+
+    # something to write on close
+    with gcs.open(fn, "wb", block_size=2**18) as f:
+        f.write(d)
+        f.write(b"xx")
+    assert gcs.cat(fn) == d + b"xx"
+    assert gcs.info(fn)["contentType"] == "application/octet-stream"
+    assert "kmsKeyName" not in gcs.info(fn)
+    # empty buffer on close
```
martindurant, line 206: This is different only in the content-type?

martindurant (commented):
Code looks good - some small comments.

I haven't actually tried this yet, since I have no keys defined.

gcsfs/tests/settings.py
kaiju: It's unrelated.

gcsfs/tests/test_core.py
kaiju: Will do :+1:

gcsfs/tests/test_core.py
kaiju: I copied the existing `test_simple_upload`, `test_large_upload`, and `test_multi_upload` upload tests just to cover passing `kms_key_name` in those contexts. I'm not clear on some of the finer details around the implementation of the original tests.

Perhaps dropping the new tests and parametrizing the existing tests to cover `kms_key_name` would be a better idea? Alternatively I'm happy to just simplify the new tests instead of wholesale copying the existing ones.

PR #760, opened 2026-02-13
Implement cat_ranges to optimize multi-range reads in Zonal bucket

suni72:
### **Key Changes**

* **Optimized Multi-Range Reads:** Implemented `_cat_ranges` in ExtendedGcsFileSystem to group and batch multi-range requests based on bucket type and use one mrd per unique zonal object path.
* **Utility:** Added `download_ranges` utility in `zb_hns_utils.py` to encapsulate batched download logic and enforce the `MRD_MAX_RANGES` limit in `AsyncMultiRangeDownloader`.
* **Refined Slicing Semantics:** Updated `_process_limits_to_offset_and_length` and `core._cat_file` to align with standard Python slicing behavior, correctly handling negative indices, zero-length, and invalid ranges without throwing errors.
* **Code Refactoring:** Updated `fetch_range_split` to use new download_ranges utility method for improved readability and added documentation for the `supports_append` argument in `GCSFile`.
* **Testing & Validation:** Added unit and integration tests for `_cat_ranges`, sync `cat_ranges`, `_fetch_zonal_batch`, `_group_requests_by_bucket_type`, and `download_ranges`, including specific validation for negative batch sizes and handling of mixed Zonal/Regional buckets in `_cat_ranges`.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1942,6 +1942,9 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
             Timeout seconds for the asynchronous callback.
         generation: str
             Object generation.
+        supports_append: bool
+            If True, allows opening file in append mode. This is generally not supported
+            by GCS, but may be supported by subclasses (e.g. ZonalFile).
         """
         bucket, key, path_generation = gcsfs.split_path(path)
         if not key:
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1167,8 +1167,20 @@ class GCSFileSystem(asyn.AsyncFileSystem):
 
     async def _cat_file(self, path, start=None, end=None, **kwargs):
         """Simple one-shot get of file data"""
+        # if start and end are both provided and valid, but start >= end, return empty bytes
+        # Otherwise, _process_limits would generate an invalid HTTP range (e.g. "bytes=5-4"
+        # for start=5, end=5), causing the server to return the whole file instead of nothing.
+        if (
+            start is not None
+            and end is not None
+            and start >= 0
+            and end >= 0
+            and start >= end
+        ):
+            return b""
         u2 = self.url(path)
-        if start or end:
+        # 'if start or end' fails when start=0 or end=0 because 0 is Falsey.
+        if start is not None or end is not None:
             head = {"Range": await self._process_limits(path, start, end)}
         else:
             head = {}
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1967,7 +1967,9 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
             Object generation.
         supports_append: bool
             If True, allows opening file in append mode. This is generally not supported
-            by GCS, but may be supported by subclasses (e.g. ZonalFile).
+            by GCS, but may be supported by subclasses (e.g. ZonalFile). This flag
+            should be set by subclasses that support append operations. Otherwise,
+            the mode will be overwritten to "wb" mode with a warning.
         """
         bucket, key, path_generation = gcsfs.split_path(path)
         if not key:
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -25,6 +25,7 @@ from google.cloud.storage.asyncio.async_multi_range_downloader import (
 from gcsfs import __version__ as version
 from gcsfs import zb_hns_utils
 from gcsfs.core import GCSFile, GCSFileSystem
+from gcsfs.zb_hns_utils import MRD_MAX_RANGES
 from gcsfs.zonal_file import ZonalFile
 
 logger = logging.getLogger("gcsfs")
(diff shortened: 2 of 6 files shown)


gcsfs/core.py
```diff
@@ -1167,8 +1167,20 @@ def url(self, path):
...
         """Simple one-shot get of file data"""
+        # if start and end are both provided and valid, but start >= end, return empty bytes
+        # Otherwise, _process_limits would generate an invalid HTTP range (e.g. "bytes=5-4"
+        # for start=5, end=5), causing the server to return the whole file instead of nothing.
+        if (
+            start is not None
+            and end is not None
+            and start >= 0
+            and end >= 0
+            and start >= end
```
ankitaluthra1, line 1178: Do we know any scenarios in which these invalid values would be passed ?

gcsfs/core.py
```diff
@@ -1953,6 +1965,9 @@ def __init__(
             Timeout seconds for the asynchronous callback.
         generation: str
             Object generation.
+        supports_append: bool
+            If True, allows opening file in append mode. This is generally not supported
```
ankitaluthra1, line 1969: Consider making the purpose a bit more clearer, when it should be set

gcsfs/extended_gcsfs.py
```diff
@@ -203,17 +205,15 @@ async def _process_limits_to_offset_and_length(
...
         """
         size = file_size
 
         if start is None:
             offset = 0
         elif start < 0:
             size = (await self._info(path))["size"] if size is None else size
-            offset = size + start
+            # If start is negative and larger than the file size, we should start from 0.
+            offset = max(0, size + start)
```
ankitaluthra1, line 216: Again, what are scenarios in which start can be larger than file size and actully the user wanted start as 0

gcsfs/extended_gcsfs.py
```diff
@@ -390,6 +375,179 @@ async def _is_bucket_hns_enabled(self, bucket):
...
+        try:
+            await self._get_grpc_client()
+            mrd = await AsyncMultiRangeDownloader.create_mrd(
+                self.grpc_client, bucket, object_name, generation
+            )
+
+            file_size = mrd.persisted_size
+            if file_size is None:
+                # set file_size here to avoid network call in process_limits_to_offset_and_length
+                file_size = (await self._info(f"{bucket}/{object_name}"))["size"]
```
ankitaluthra1, line 393: when do we need this fallback ? are there any scenarios in which this field is not present, my assumption was field would always be present

ankitaluthra1 (changes requested):

gcsfs/extended_gcsfs.py
```diff
@@ -390,6 +375,179 @@ async def _is_bucket_hns_enabled(self, bucket):
...
+            )
+
+        # Execute concurrently and stitch Results
+        all_results = await asyncio.gather(*tasks)
+        results = [None] * len(paths)
+        for batch_res in all_results:
+            for idx, val in batch_res:
+                results[idx] = val
+
+        return results
```
ankitaluthra1, line 554: the method is too long, consider smaller modular extractions

gcsfs/extended_gcsfs.py
```diff
@@ -390,6 +375,179 @@ async def _is_bucket_hns_enabled(self, bucket):
...
+                if sem:
+                    async with sem:
+                        result = await coro
+                else:
+                    result = await coro
+
+                # Zonal returns a zip/list of multiple items.
+                # Regional (super._cat_file) returns a single bytes object.
+                # We normalize everything to: [(index, data), (index, data)...]
+                if isinstance(result, bytes):
```
ankitaluthra1: This check seems a bit fragile based on return types, consider having wrapper method with more dependable check of bucket being regional or zonal

ankitaluthra1 (changes requested):

gcsfs/core.py
suni72: This is more like a safety measure to accurately handle all cases in a similar way python slices work, since we don't raise error for invalid inputs. It's just a speculation but..when methods like open_parquet dynamically calculate which ranges to read for requested data, it might calculate the start or end values as negative for edge cases/corrupt files. code ref for open parquet range calculation: https://github.com/fsspec/filesystem_spec/blob/90bcbba391bddef400dde62e03c2eea9a2bdbd3d/fsspec/parquet.py#L173

gcsfs/extended_gcsfs.py
suni72: This is to ensure we mimic python slicing behavior. For example, if a file is 100 bytes long, and a user (or a library) requests the last 150 bytes (file[-150:]), standard Python behavior is to cap at the beginning of the file and return the whole 100 bytes, rather than throwing an invalid input error.

gcsfs/core.py
suni72: Updated comment to: 
```
If True, allows opening file in append mode. This is generally not supported
            by GCS, but may be supported by subclasses (e.g. ZonalFile). This flag
            should be set by subclasses that support append operations. Otherwise,
            the mode will be overwritten to "wb" mode with a warning.
```

gcsfs/extended_gcsfs.py
suni72: Added a helper method for creating the async tasks for zonal and regional batches

gcsfs/extended_gcsfs.py
suni72: Added a wrapper for normalizing the results

gcsfs/extended_gcsfs.py
suni72: I added this fallback since persisted_size is an optional field and only set when stream is opened without read_handle: https://github.com/googleapis/python-storage/blob/d8dd1e074d2431de9b45e0103181dce749a447a0/google/cloud/storage/asyncio/async_read_object_stream.py#L127. Confirming this with Chandra if the field will always be present

PR #780, opened 2026-03-19
Parallelize bucket get and listing in _info for bucket paths

yuxin00j:
This PR optimizes the GCSFileSystem._info method for bucket roots (gs://bucket) by running the bucket metadata retrieval (GET) and the bucket contents listing (ls) concurrently. It introduces a generic parallel_tasks_first_completed async context manager to ensure clean resource management and early returns.

### Refactor GCSFileSystem._info (gcsfs/core.py)
- Uses the new utility for bucket root paths.
- If GET completes first, it immediately returns the metadata.
- If GET fails (e.g., due to permissions), it falls back to the ls result to determine if the bucket exists.


gcsfs/tests/perf/microbenchmarks/info/configs.yaml
```diff
@@ -26,3 +26,12 @@ scenarios:
     pattern: "info"
     depth: 10
     processes: [4, 8]
+
+  # Tests the fallback logic in _info() when GET bucket fails (e.g., restricted permissions).
```
jasha26: While the parallelization of GET and ls is a great optimization, I suggest moving the service account (SA) impersonation and the "restricted bucket" scenario to the integration test suite instead of the microbenchmarks:

- Functional vs. Performance: Testing "the fallback logic in _info() when GET bucket fails" is primarily a functional correctness check. Latency is unlikely to vary significantly based on the service account used.
- E2E Complexity: Using subprocess to call gcloud for tokens adds external dependencies and overhead that are outside the scope of a microbenchmark.

**Recommendation**: Keep microbenchmarks focused on the performance delta of the parallelization. Verification of the IAM fallback path is better suited for test_extended_gcsfs.py.

gcsfs/core.py
```diff
@@ -1048,17 +1048,28 @@ async def _info(self, path, generation=None, **kwargs):
         """File information about this path."""
         path = self._strip_protocol(path).rstrip("/")
         if "/" not in path:
-            try:
-                out = await self._call("GET", f"b/{path}", json_out=True)
+            # Parallelize the call and ls
+            res_call, res_ls = await asyncio.gather(
```
jasha26: If the primary goal is to save time, there is one subtle issue with asyncio.gather: it waits for both tasks to complete before returning. If the bucket GET call is fast but the ls call is slow, gather will make the user wait for the slower ls call even if the metadata is already available.

Can we try to use independent tasks:

        call_task = asyncio.create_task(self._call("GET", f"b/{path}", json_out=True))
        ls_task = asyncio.create_task(self._ls(path, max_results=1))

        try:
            # If the direct call succeeds, return immediately without waiting for ls
            out = await call_task
            out.update(size=0, type="directory")
            return out
        except OSError:
            # Fallback to the ls check which is already running in the background
            try:
                if await ls_task:
                    return {"name": path, "size": 0, "type": "directory"}
                raise FileNotFoundError(path)
            except Exception:
                raise

martindurant: > gather will make the user wait for the slower ls call

I was wondering about that. So you need asyncio.wait(return_when=FIRST_COMPLETED) actually, in case the ls finishes before the get (even if this is unlikely).

Note that you do not need to explicitly create tasks in general, you can pass instantiated coroutine objects to these functions.

gcsfs/tests/perf/microbenchmarks/info/configs.yaml
yuxin00j: Done.

gcsfs/core.py
yuxin00j: Done.

gcsfs/tests/perf/microbenchmarks/conftest.py
```diff
@@ -147,7 +147,8 @@ def _benchmark_listing_fixture_helper(
     create_folders=False,
     require_file_paths=False,
 ):
-    gcs = extended_gcs_factory()
+    gcs_admin = extended_gcs_factory()
+    gcs = gcs_admin
```
jasha26: I noticed here that gcs is being assigned to gcs_admin, but gcs is never used after that. All setup and teardown calls have been moved to gcs_admin .

Since we are recommending the removal of the service-account-specific benchmarking scenarios, we should avoid this renaming entirely. I suggest reverting gcs_admin back to gcs and removing the redundant assignment to keep the helper function clean and consistent with existing patterns

jasha26 (approved):

gcsfs/tests/perf/microbenchmarks/conftest.py
yuxin00j: Done.

gcsfs/core.py
```diff
@@ -1048,17 +1048,25 @@ async def _info(self, path, generation=None, **kwargs):
...
+                except OSError:
+                    if await ls_task:
+                        return {"name": path, "size": 0, "type": "directory"}
                     raise FileNotFoundError(path)
-            return out
+            finally:
+                if not get_task.done():
+                    get_task.cancel()
+                if not ls_task.done():
+                    ls_task.cancel()
```
googlyrahman: I don't quite understand the reasoning behind this change.

From a first look, I think this could backfire on our customers. I don't see a reason why we should concurrently execute these two requests. The second `_ls` call only happens when the backend returns a 403 (Forbidden/Permission Error), meaning the user lacks `bucket.get` permission. In that case, we try the list operation in the hope that the user has `bucket.list` permission, which makes sense.

If you look at gcsfs/retry.py, the `OSError` is only raised when the status code is 403. In all other cases, the exception is different, meaning we would never need to execute the `_ls` call.

However, with this change, no matter the scenario, we end up making two calls where one would usually suffice. This might not be noticeable for a few `_info` calls, but it will backfire significantly when there is a high volume of them.

I see that you're cancelling the task once we have the result of get_task, but that doesn't really help. The asyncio event loop will still have initiated both requests concurrently, wasting resources just to speed things up for users who lack `bucket.get` permission while degrading the experience for those who have both permissions, or just `bucket.get`.

Hence, if my understanding is correct, this PR would only improve performance for the failure case where a user lacks `bucket.get` but has `bucket.list` permissions (rare customer) while degrading performance for the happy path, where the user has `bucket.get` or both permissions.

martindurant: You make good points here. Can we measure?

yuxin00j: Yes, you are right. This change aims to make the case when it falls back to `bucket.list` faster.

yuxin00j: I updated the code to use the same implementation as #786 to handle parallel tasks.
For happy path, I ran the micro benchmark for main and this branch: [result](https://docs.google.com/spreadsheets/d/1AfmFXI2JAzLMLWWYLW6CTx0HGLAFOtpgbpTkybfsVC0/edit?usp=sharing). The percentage of change are between -29% and 15%. Half of the cases are above 0 and half are below 0. The absolute values of the latencies are small. So I believe these are reasonable variance and this change does not degrade the happy path performance.

martindurant: What is the number of directories for this benchmark case - does it matter?

yuxin00j: The number of directories does not matter here.

PR #787, opened 2026-03-25
Enhance Retry Logic and Configuration for Storage Control API

Mahalaxmibejugam:
Implement a robust, configurable retry mechanism for the Storage Control API (used for HNS and Zonal bucket operations) within ExtendedGcsFileSystem. It replaces the default SDK retry behavior with a custom AsyncRetry configuration providing granular control over timeouts and retriable errors.

 Key Changes
   * Custom Retries: Integrated AsyncRetry into all Storage Control RPCs (rename, mkdir, ls, etc.) with a 60s
     default budget.
   * Explicit Timeouts: Decoupled attempts from the global budget using a fixed 30s per-attempt RPC timeout.
   * Error Predicate: Expanded retriable errors to include standard transient failures and specific
     unauthenticated scenarios.
   * Configuration: Added support for customizing retry parameters (initial, timeout, etc.) via the FileSystem
     constructor or fsspec conf files.
   * Testing & Docs: Added comprehensive unit tests for retry resolution and updated the retries.rst
     documentation.
     
     
Also updated CloudBuild cleanup scripts to correctly handle billing projects for requester-pays buckets observed while running CI pipeline for this PR

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -20,11 +20,13 @@ from google.cloud.storage.asyncio.async_grpc_client import AsyncGrpcClient
 from gcsfs import __version__ as version
 from gcsfs import zb_hns_utils
 from gcsfs.core import GCSFile, GCSFileSystem
+from gcsfs.retry import StorageControlRetryConfig, get_storage_control_retry_config
 from gcsfs.zonal_file import ZonalFile
 
 logger = logging.getLogger("gcsfs")
 
 USER_AGENT = "python-gcsfs"
+STORAGE_CONTROL_RPC_TIMEOUT = 30.0
 
 
 class BucketType(Enum):
@@ -51,6 +53,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
     """
 
     def __init__(self, *args, finalize_on_close=False, **kwargs):
+        self.retry_config = StorageControlRetryConfig.from_kwargs(**kwargs)
         super().__init__(*args, **kwargs)
         # By default, files in zonal buckets are left unfinalized to allow appends.
         self.finalize_on_close = finalize_on_close
@@ -78,6 +81,9 @@ class ExtendedGcsFileSystem(GCSFileSystem):
                 else self.project
             )
         return None
+
+    def _get_retry_config(self, **kwargs):
+        return get_storage_control_retry_config(self.retry_config, **kwargs)
 
     @property
     def grpc_client(self):
@@ -141,7 +147,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
             client = await self._get_control_plane_client()
             bucket_name_value = f"projects/_/buckets/{bucket}/storageLayout"
             logger.debug(f"get_storage_layout request for name: {bucket_name_value}")
-            response = await client.get_storage_layout(name=bucket_name_value)
+            response = await client.get_storage_layout(
+                name=bucket_name_value,
+                retry=self._get_retry_config(),
+                timeout=STORAGE_CONTROL_RPC_TIMEOUT,
+            )
 
             if response.location_type == "zone":
                 return BucketType.ZONAL_HIERARCHICAL
@@ -514,7 +524,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
 
                 logger.debug(f"rename_folder request: {request}")
                 client = await self._get_control_plane_client()
-                operation = await client.rename_folder(request=request)
+                operation = await client.rename_folder(
+                    request=request,
+                    retry=self._get_retry_config(),
+                    timeout=STORAGE_CONTROL_RPC_TIMEOUT,
+                )
                 await operation.result()
                 self._update_dircache_after_rename(path1, path2)
 
@@ -679,7 +693,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         try:
             logger.debug(f"create_folder request: {request}")
             client = await self._get_control_plane_client()
-            await client.create_folder(request=request)
+            await client.create_folder(
+                request=request,
+                retry=self._get_retry_config(),
+                timeout=STORAGE_CONTROL_RPC_TIMEOUT,
+            )
             # Instead of invalidating the parent cache, update it to add the new entry.
             parent_path = self._parent(path)
             if parent_path in self.dircache:
@@ -725,7 +743,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
 
                 # Verify existence using get_folder API
                 client = await self._get_control_plane_client()
-                response = await client.get_folder(request=request)
+                response = await client.get_folder(
+                    request=request,
+                    retry=self._get_retry_config(),
+                    timeout=STORAGE_CONTROL_RPC_TIMEOUT,
+                )
 
                 # If successful, return directory metadata
                 return {
@@ -798,7 +820,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
 
             logger.debug(f"delete_folder request: {request}")
             client = await self._get_control_plane_client()
-            await client.delete_folder(request=request)
+            await client.delete_folder(
+                request=request,
+                retry=self._get_retry_config(),
+                timeout=STORAGE_CONTROL_RPC_TIMEOUT,
+            )
 
             # Remove the directory from the cache and from its parent's listing.
             self.dircache.pop(path, None)
@@ -1121,7 +1147,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         logger.debug(f"list_folders request: {request}")
 
         client = await self._get_control_plane_client()
-        async for folder in await client.list_folders(request=request):
+        async for folder in await client.list_folders(
+            request=request,
+            retry=self._get_retry_config(),
+            timeout=STORAGE_CONTROL_RPC_TIMEOUT,
+        ):
             folders.append(self._create_folder_entry(bucket, folder))
 
         return folders
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -20,7 +20,7 @@ from google.cloud.storage.asyncio.async_grpc_client import AsyncGrpcClient
 from gcsfs import __version__ as version
 from gcsfs import zb_hns_utils
 from gcsfs.core import GCSFile, GCSFileSystem
-from gcsfs.retry import StorageControlRetryConfig, get_storage_control_retry_config
+from gcsfs.retry import get_storage_control_retry_config
 from gcsfs.zonal_file import ZonalFile
 
 logger = logging.getLogger("gcsfs")
@@ -53,7 +53,9 @@ class ExtendedGcsFileSystem(GCSFileSystem):
     """
 
     def __init__(self, *args, finalize_on_close=False, **kwargs):
-        self.retry_config = StorageControlRetryConfig.from_kwargs(**kwargs)
+        self.retry_config = {
+            k: v for k, v in kwargs.items() if k.startswith("retry_") and v is not None
+        }
         super().__init__(*args, **kwargs)
         # By default, files in zonal buckets are left unfinalized to allow appends.
         self.finalize_on_close = finalize_on_close
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -20,7 +20,7 @@ from google.cloud.storage.asyncio.async_grpc_client import AsyncGrpcClient
 from gcsfs import __version__ as version
 from gcsfs import zb_hns_utils
 from gcsfs.core import GCSFile, GCSFileSystem
-from gcsfs.retry import get_storage_control_retry_config
+from gcsfs.retry import DEFAULT_RETRY_CONFIG, get_storage_control_retry_config
 from gcsfs.zonal_file import ZonalFile
 
 logger = logging.getLogger("gcsfs")
@@ -53,8 +53,25 @@ class ExtendedGcsFileSystem(GCSFileSystem):
     """
 
     def __init__(self, *args, finalize_on_close=False, **kwargs):
+        """
+        Parameters
+        ----------
+        finalize_on_close : bool, default False
+            By default, files in zonal buckets are left unfinalized to allow appends.
+        **kwargs : dict
+            Additional arguments passed to GCSFileSystem.
+            Supports retry configuration overrides for Storage Control API:
+            - retry_timeout: Total time to spend retrying (seconds).
+            - retry_initial: Initial delay between retries (seconds).
+            - retry_maximum: Maximum delay between retries (seconds).
+            - retry_multiplier: Multiplier for delay between retries.
+            These map to `google.api_core.retry.AsyncRetry` arguments (without 'retry_' prefix).
+        """
+        valid_keys = DEFAULT_RETRY_CONFIG.keys()
         self.retry_config = {
-            k: v for k, v in kwargs.items() if k.startswith("retry_") and v is not None
+            k[6:]: v
+            for k, v in kwargs.items()
+            if k.startswith("retry_") and k[6:] in valid_keys and v is not None
         }
         super().__init__(*args, **kwargs)
         # By default, files in zonal buckets are left unfinalized to allow appends.
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -11,17 +11,11 @@ from google.api_core import exceptions as api_exceptions
 from google.api_core.retry import AsyncRetry
 
 logger = logging.getLogger("gcsfs")
-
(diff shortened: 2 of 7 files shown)


gcsfs/retry.py
jasha26, line 1: Instead of the custom asyncio.wait_for logic, using a library like [tenacity](https://tenacity.readthedocs.io/en/latest/) would look like this. Also did we evaluate google.api_core [AsyncRetry](https://googleapis.dev/python/google-api-core/latest/retry.html):

```
from tenacity import retry, wait_exponential, stop_after_attempt, retry_if_exception_type

@retry(
    wait=wait_exponential(multiplier=1, min=2, max=32),
    stop=stop_after_attempt(6),
    retry=retry_if_exception_type((api_exceptions.ServiceUnavailable, asyncio.TimeoutError)),
    reraise=True
)
async def call_with_retry(func, *args, **kwargs):
```

gcsfs/retry.py
```diff
@@ -176,3 +177,61 @@ async def retry_request(func, retries=6, *args, **kwargs):
                 continue
             logger.exception(f"{func.__name__} non-retriable exception: {e}")
             raise e
+
+
+async def execute_with_timebound_retry(
+    func, *args, retry_deadline=30.0, max_retries=6, **kwargs
```
jasha26: We can't really hard code these values, we need the ability so these can be overidden via multiple mechanisms like call site overrides, fsspec config overrides etc. 

So i'd recommend we do something like below:

```
from fsspec.config import conf

@dataclass
class RetryConfig:
    max_retries: int = 6
    min_delay: float = 2.0
    max_delay: float = 32.0
    retry_deadline: float = 30.0

def get_resolved_retry_config(call_kwargs) -> RetryConfig:
    """
    Resolves retry configuration with a clear hierarchy of overrides:
    1. Explicit call-site arguments (e.g., max_retries=10)
    2. fsspec.config settings (e.g., ~/.config/fsspec/conf.json)
    3. Hardcoded Defaults from the RetryConfig template
    """
    # 1. Start with the default template
    default = RetryConfig()

    # 2. Resolve parameters from Env Vars or fsspec.config, or use defaults
    resolved_max_retries = int(
        call_kwargs.get("max_retries")
        or conf.get("gcsfs.retry.max_retries", default.max_retries)
    )
    
    resolved_deadline = float(
        call_kwargs.get("retry_deadline")
        or conf.get("gcsfs.retry.deadline", default.retry_deadline)
    )

    return RetryConfig(
        max_retries=resolved_max_retries,
        retry_deadline=resolved_deadline,
        min_delay=default.min_delay,
        max_delay=default.max_delay
    )

async def with_retry(func, *args, **kwargs):
    config = get_resolved_retry_config(kwargs)
    
    # Define transient errors consistent with GCS client best practices.
    RETRYABLE_ERRORS = (
        api_exceptions.ServiceUnavailable,
        api_exceptions.DeadlineExceeded,
        api_exceptions.InternalServerError,
        api_exceptions.TooManyRequests,
        asyncio.TimeoutError,
    )

    # Replaces custom loop with a declarative tenacity decorator.
    @retry(
        stop=stop_after_attempt(config.max_retries),
        wait=wait_exponential(multiplier=1, min=config.min_delay, max=config.max_delay),
        retry=retry_if_exception_type(RETRYABLE_ERRORS),
        reraise=True
    )
    async def _wrapped_call():
    
    return await _wrapped_call()
```

martindurant: I am not a fan of adding all these environment variables. fsspec already has a way to specify instantiation kwargs using specially formatted environment variables or files ( https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration ).

Since retries are useful in multiple backends and have the same concepts (number of times, backoff factor, max wait, etc.), we could even make a general fsspec class for this.

gcsfs/retry.py
Mahalaxmibejugam: We need asyncio.wait_for logic on the client side to make sure that we handle the request stalls and won't wait indefinitely for the call to return. Replaced the custom logic with tenacity as it provides in-built support for retries. AsyncRetry also provides the same functionality but would still need the asyncio.wait_for logic on the client side.

martindurant: Do we really need another dependency?

gcsfs/retry.py
```diff
@@ -176,3 +189,93 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+        and "Invalid Credentials" in str(exception)
+    ):
+        is_transient = True
+    return is_transient
+
+
+def _get_retry_config(
+    retry_deadline=None, max_retries=None
+) -> StorageControlRetryConfig:
+    conf = fsspec.config.conf.get("gcs", {})
```
martindurant: the purpose of the conf is to pass default values to the class constructor at instantiation, so there is not normally any need to query the conf directly, but the values should be passed in.

gcsfs/retry.py
```diff
@@ -176,3 +189,93 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+
+    return StorageControlRetryConfig(
```
martindurant: As a dataclass, this already has these defaults.
I think you would be better of passing kwargs round, like in other places inthe codebase. The Config class isn't doing anything for you, except obscuring when the values are being set.

gcsfs/retry.py
```diff
@@ -176,3 +189,93 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+    )
+
+
+async def execute_with_timeout(func, timeout, *args, **kwargs):
+    """
+    Executes a function with a strict timeout using asyncio.wait_for.
+    It passes timeout to the function itself and uses timeout + 1.0 for wait_for.
+    """
+    return await asyncio.wait_for(
+        func(*args, timeout=timeout, **kwargs), timeout=timeout + 1.0
```
martindurant: Where does the arbitrary + 1.0 come from?
How doe we know that `func` takes a timeout parameter (and if it does, why do we need another one?) ?

gcsfs/retry.py
```diff
@@ -176,3 +189,93 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+
+    retryer = AsyncRetrying(
+        stop=stop_after_attempt(rc.max_retries),
+        wait=_custom_wait,
+        retry=retry_if_exception(_is_transient_exception),
+        before_sleep=_before_sleep,
+        reraise=True,
+    )
+
+    return await retryer(
```
martindurant: I don't see that we need an extra library for this one thing. One might argue that this retry logic belongs in fsspec, but it would be fine to write it out here.

martindurant (commented):
Is there anything about the retry logic here that's specific to gRPC? I doesn't appear so to me.

gcsfs/retry.py
Mahalaxmibejugam: I had a similar opinion, but @jasha26 recommended using Tenacity as it might help in future integrations like client-side throttling

AsyncRetry from google.api_core supports retries based on max_timeout instead of max_retries(which is followed in gcsfs for other JSON API retries). To keep the same retry behaviour for JSON APIs and storage control client calls(i.e., limiting the number of retries) I have implemented the custom logic. To use AsyncRetry with the constraint on number of attempts we would have to maintain a wrapper to track the number of attempts which would be almost same as the initial version of this PR without adding much benefit of using AsyncRetry

So we can either have entirely custom implementation or use Tenacity if we want to maintain the max_attempts behaviour across GCSFS

gcsfs/retry.py
Mahalaxmibejugam: The + 1.0 buffer: It gives the underlying gRPC library a 1-second grace period to cleanly abort and raise its own specific timeout error (e.g., DeadlineExceeded) before Python's asyncio aggressively kills the task.

How we know func takes timeout: This wrapper is specifically built for Google Cloud GAPIC (gRPC) client methods (like the ones in storage_control_v2 - create_folder, rename_folder). By standard design, all of these methods accept a timeout kwarg to set the RPC deadline.

Why we need both: The inner timeout is the actual RPC deadline sent to the server and gRPC core. The outer asyncio.wait_for is a hard fail-safe to ensure the Python event loop doesn't hang forever if the network layer freezes and fails to respect the inner deadline.

This is more specifically added based on the issues that were observed in GCSFuse related to request stalling.

gcsfs/retry.py
Mahalaxmibejugam: Are you referring to remove entire Tenacity dependency and go back to custom implementation?

jasha26: Hi @martindurant, 

I completely understand the hesitation around adding new dependencies—I’m usually of the same mind when it comes to keeping the project's footprint light. You're absolutely right that this logic might eventually be better suited for fsspec globally. 

The main reason I proposed tenacity is that it has become an industry standard for production-grade retries. It handles the nuances of exponential backoff with jitter and async/sync compatibility right out of the box, which can be surprisingly tricky to get 100% right in a custom implementation.

While we can certainly roll a simpler custom version as you suggested, I thought using a battle-tested library might save us from 'reinventing the wheel' and maintenance overhead later. However, if the priority is keeping the dependency list short, we're happy to strip it back. 

What do you think is the best balance for the project?

Mahalaxmibejugam: @martindurant Please let us know about your view on this

martindurant: OK leave it be. `tencity` certainly isn't big. 

Is there anything left to do here?

gcsfs/retry.py
Mahalaxmibejugam: Done

gcsfs/retry.py
Mahalaxmibejugam: Done

gcsfs/retry.py
Mahalaxmibejugam: Updated the code to use AsyncRetry, this PR is refactored to enable adding custom retries using AsyncRetry

gcsfs/retry.py
Mahalaxmibejugam: I have updated code to use AsyncRetry which is supported within the control client APIs

gcsfs/retry.py
```diff
@@ -2,14 +2,42 @@
...
+    multiplier: float = None
+
+    @classmethod
+    def from_kwargs(cls, **kwargs):
+        """Creates a config from kwargs, filtering for retry_ prefix."""
+        return cls(
+            timeout=kwargs.get("retry_timeout"),
+            initial=kwargs.get("retry_initial"),
+            maximum=kwargs.get("retry_maximum"),
+            multiplier=kwargs.get("retry_multiplier"),
```
martindurant: It looks like the attributes above could be `None`. Can we not use dataclass's autogenerated init?

gcsfs/retry.py
```diff
@@ -176,3 +204,58 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+            api_exceptions.InternalServerError,
+            api_exceptions.TooManyRequests,
+            api_exceptions.ResourceExhausted,
+            api_exceptions.Unknown,
+        ),
+    )
+    if (
+        not is_transient
+        and isinstance(exception, api_exceptions.Unauthenticated)
+        and "Invalid Credentials" in str(exception)
```
martindurant, line 204: or?

gcsfs/retry.py
```diff
@@ -176,3 +204,58 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+        val = getattr(overrides, attr, None)
+        if val is not None:
+            return val
+        # 2. Check base filesystem configuration
+        if base_config:
+            val = getattr(base_config, attr, None)
+            if val is not None:
+                return val
+        # 3. Fallback to package defaults
+        return globals()[f"DEFAULT_RETRY_{attr.upper()}"]
```
martindurant: Could be much simpler with something like
```
kw = DEFAULT_RETRY_CONFIG.copy() # a dict
kw.update(config or {}) # assume config can be dict-like
kw.update(kwargs)
return AsyncRetry(**kw)
```

gcsfs/retry.py
Mahalaxmibejugam: The idea was these can be None, and we resolve based on priority at one place while constructing AsyncRetry.

But as per other comment, I have used dict and entirely removed dataclass. Having both of them looks redundant

gcsfs/retry.py
Mahalaxmibejugam: Done

gcsfs/retry.py
Mahalaxmibejugam: Making it Or would retry any Unauthenticated error or any error that happens to have "Invalid Credentials" in its string representation regardless of the error type.

Keeping "and" is consistent with how it's handled for standard HTTP requests in the same file at line #111

gcsfs/extended_gcsfs.py
```diff
@@ -51,6 +53,9 @@ class ExtendedGcsFileSystem(GCSFileSystem):
     """
 
     def __init__(self, *args, finalize_on_close=False, **kwargs):
+        self.retry_config = {
+            k: v for k, v in kwargs.items() if k.startswith("retry_") and v is not None
+        }
```
martindurant, line 75: You can rename here, so that you don't have the "retry_" prefix in something that's already called "retry"
```suggestion
        self.retry_config = {
            k[6:]: v for k, v in kwargs.items() if k.startswith("retry_") and v is not None
        }
```

martindurant: The sent of extra kwargs that can be passed should be documented, with a reference to AsyncRetry (which actually defines how they are used)

gcsfs/retry.py
```diff
@@ -7,9 +7,23 @@
...
+DEFAULT_RETRY_MAXIMUM = 60.0
+DEFAULT_RETRY_MULTIPLIER = 2.0
+
+DEFAULT_RETRY_CONFIG = {
+    "retry_timeout": DEFAULT_RETRY_TIMEOUT,
+    "retry_initial": DEFAULT_RETRY_INITIAL,
+    "retry_maximum": DEFAULT_RETRY_MAXIMUM,
+    "retry_multiplier": DEFAULT_RETRY_MULTIPLIER,
+}
+
```
martindurant:
```suggestion
DEFAULT_RETRY_CONFIG = {
    "timeout": 60.0,
    "initial": 1.0,
    "maximum": 60.0,
    "multiplier": 2.0,
}

```

gcsfs/retry.py
```diff
@@ -176,3 +190,52 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+    """
+    retry_kwargs = DEFAULT_RETRY_CONFIG.copy()
+    if base_config:
+        retry_kwargs.update(base_config)
+
+    overrides = {
+        k: v for k, v in kwargs.items() if k.startswith("retry_") and v is not None
+    }
+    retry_kwargs.update(overrides)
+
```
martindurant: Just call them initial, maximum, multiplier, timeout

```suggestion
    retry_kwargs.update(kwargs)

```

gcsfs/retry.py
```diff
@@ -176,3 +190,52 @@ async def retry_request(func, retries=6, *args, **kwargs):
...
+    return AsyncRetry(
+        predicate=_is_transient_exception,
+        initial=retry_kwargs["retry_initial"],
+        maximum=retry_kwargs["retry_maximum"],
+        multiplier=retry_kwargs["retry_multiplier"],
+        timeout=retry_kwargs["retry_timeout"],
+    )
```
martindurant:
```suggestion
    return AsyncRetry(predicate=_is_transient_exception, **retry_kwargs)
```

martindurant (commented):
I prefer to simplify the arguments. Also, I know there is extra test in the rst documentation, but the arguments should also be mentioned where we expect the user might provide them.

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: Done, also made sure we are only storing valid keys in retry_config

gcsfs/retry.py
Mahalaxmibejugam: Done

gcsfs/retry.py
Mahalaxmibejugam: Done

gcsfs/retry.py
Mahalaxmibejugam: To make sure we don't pass invalid kwargs to AsyncRetry, I added filtering logic for base config and overrides. But renamed by removing "retry_" prefix

martindurant:
Please also check the rst docs, in case any of the latter changes needs an update there. One CI run appears to have timed out, I don't know why.

Mahalaxmibejugam:
> Please also check the rst docs, in case any of the latter changes needs an update there.

Retries doc only mentions about how users can pass the retry args which is unchanged, only changes are for internal processing. So no updates are needed in the docs.

PR #800, opened 2026-04-02
Use moveTo API to move files for standard buckets

Yonghui-Lee:
I update the implementation of move method for standard bucket to use the atomic moveTo API instead of the non-atomic copy-and-delete.

## Key Changes
- Replace copy-and-delete with atomic moveTo.
- List and Glob Pattern Support: Expand source and destination paths using fsspec's robust other_paths helper. It supports:
- Asynchronous Batched Execution: Moves are executed aggressively in parallel using _run_coros_in_chunks (governed by batch_size kwargs parameter).
- Error handling for Implicit Directories: Because standard un-hierarchical GCS buckets use implicit directories, _expand_path() will identify dummy directory paths that don't physically exist as objects. The exception handling loop gracefully catches and suppresses FileNotFoundError if it corresponds to an implicit directory.
- Add a specific microbenchmark for file renaming operations, segregating it from directory renaming benchmarks.

## Performance
For previous copy-and-delete implementation
- copy tasks are executed in coroutines.
- delele tasks can make use of batch API and are also executed in coroutines.

For moveTo API, tasks are executed in coroutines.

The micro benchmark results show that moveTo is 10-20% faster than copy-and-delete using the same batch size.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -14,6 +14,7 @@ import uuid
 import warnings
 import weakref
 from datetime import datetime, timedelta
+from glob import has_magic
 from urllib.parse import parse_qs
 from urllib.parse import quote as quote_urllib
 from urllib.parse import urlsplit
@@ -23,7 +24,7 @@ import fsspec
 from fsspec import asyn
 from fsspec.callbacks import NoOpCallback
 from fsspec.implementations.http import get_client
-from fsspec.utils import setup_logging, stringify_path
+from fsspec.utils import other_paths, setup_logging, stringify_path
 
 from . import __version__ as version
 from .checkers import get_consistency_checker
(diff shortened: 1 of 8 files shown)


martindurant:
How long has this API been available? :)

gcsfs/core.py
```diff
@@ -1272,16 +1267,70 @@ async def _merge(self, path, paths, acl=None):
 
     merge = asyn.sync_wrapper(_merge)
 
-    # mv method is already available as sync method in the fsspec.py
-    # Async version of it is introduced here so that mv can be used in async methods.
     # TODO: Add async mv method in the async.py and remove from GCSFileSystem.
```
Mahalaxmibejugam, line 1284: We can remove this ToDo now, as this is GCS specific custom implementation, we can't copy this method to async.py. We will have to keep it in GCSFileSystem only.

gcsfs/tests/test_core.py
```diff
@@ -2086,6 +1854,8 @@ def test_sign(gcs, monkeypatch):
 
 @pytest.mark.xfail(reason="emulator does not support condition")
 def test_write_x_mpu(gcs):
+    # if not gcs.on_google:
```
Mahalaxmibejugam: Please remove the commented code.

Mahalaxmibejugam:
Can you attach the benchmark values of standard buckets for both file and folder renames in the description or in a comment?

gcsfs/core.py
martindurant: Actually, the implementation looks generic - but no need to take any action about it now. In fact, I don't expect too much of a performance gain (but maybe some) - we just need to make sure it passes the abstract tests for `mv`.

Mahalaxmibejugam: ohh yes agree that it is generic, I missed that moveTo API is abstracted by mv_file.

Yonghui-Lee:
> How long has this API been available? :)

It has been available for about one year. https://docs.cloud.google.com/storage/docs/release-notes#February_24_2025

gcsfs/tests/test_core.py
Yonghui-Lee: Thanks for catching it. Currently we have 2 methods to handle emulator limitation: checking gcs.on_google and using @pytest.mark.xfail. I leave the comments because I'm considering whether we should change the code here to keep them consistent. May I know your opinion?

Yonghui-Lee:
> Can you attach the benchmark values of standard buckets for both file and folder renames in the description or in a comment?

The benchmark raw data for batch size 20 is as follows.
Copy-and-delete:
image

MoveTo:
image

gcsfs/tests/test_core.py
Mahalaxmibejugam: Previously, there was no CI pipeline to run the tests in test\_core.py on real GCS, and we only had an emulator-based pipeline where an unconditional xfail was acceptable. 

Now that we are running these tests on actual GCS, marking a test as xfail without checking the condition means legitimate failures will not be correctly reported. We should switch to using a conditional xfail or skip. Using the pytest.skip(...) pattern would ensure consistency with the other tests in the file.

Yonghui-Lee: Updated

martindurant:
(e2e test error is not showing up for me)

Mahalaxmibejugam:
> (e2e test error is not showing up for me)

Tests didn't run yet, we need to add `/gcbrun` as a comment to trigger the pipeline run. You and Ankita have the permissions to trigger the tests.

gcsfs/core.py
Mahalaxmibejugam: Should we consider moving the implementation from GCSFileSystem to fsspec? We might also see performance gains if other fsspec implementations also include atomic mv_file like we see in case of GCSFS.

There is also a significant code duplication in mv method from other fsspec methods like copy which can be extracted into a helper.

martindurant:
(the CI runs should show up as pending if they have not begun yet, but now I get it)

gcsfs/core.py
Yonghui-Lee: I think it's a good idea. But the default implementation of mv_file of AsyncFileSystem isn't atomic. If we use mv_file for mv by default, there may be performance drop for some fsspec implementations.

Mahalaxmibejugam: The existing mv implementation in AbstractFileSystem is also not atomic, it relies on a copy+delete logic. Therefore IMO changing from copy+delete to mv_file (which is also not atomic in AsyncFileSystem) should not have any performance implications.

@martindurant Let us know your thoughts about this

Yonghui-Lee:
The e2e test should be fixed now.

gcsfs/core.py
martindurant: Let's not move it right now, but open an issue at fsspec about a future intent to do so. I don't believe there are any existing implementations of _mv in subclasses of Async, but we should spend the time to check for sure.

Yonghui-Lee:
~~The codecov patch check failed. But I run the CI locally and check the coverage report, the lines are covered.~~

Fixed after I rebase master instead of merging.

PR #801, opened 2026-04-02
Update README.md

trinadhkotturu:
Updating the landing page for improved support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,85 @@
-gcsfs
-=====
+
+# gcsfs
 
 [|Build Status|](https://github.com/fsspec/gcsfs/actions)
 [|Docs|](https://gcsfs.readthedocs.io/en/latest/?badge=latest)
 
-Pythonic file-system for Google Cloud Storage for fsspec.
+**GCSFS** is a Python library that provides a familiar, file-system-like interface to [Google Cloud Storage (GCS)](https://docs.cloud.google.com/storage/docs/introduction). Built on top of [fsspec](https://github.com/fsspec), it allows you to interact with cloud buckets as if they were local directories, making it a favorite for data scientists and engineers.
 
+-----
 
-Support
--------
+## Getting Started
+
+### Installation
+
+Install via pip or conda:
+
+```bash
+# Using pip
+pip install gcsfs
+
+# OR using conda
+conda install -c conda-forge gcsfs
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,17 +1,85 @@
...
+
+### Authentication Modes
+
+  * **Default:** Uses your local gcloud credentials or environment service accounts.
+  * **Cloud:** Explicitly use Google Metadata service (`token='cloud'`).
+  * **Anonymous:** Access public data without a login (`token='anon'`).
+  * **Service Account:** Pass the path to your JSON key file (`token='path/to/key.json'`).
+
+> [\!TIP]
+> **Note on Async:** GCSFS is built on `aiohttp`. If you are building high-concurrency applications, you can use the asynchronous API by passing `asynchronous=True` to the `GCSFileSystem` constructor.
```
martindurant, line 77: Does Extended need aiohttp at all?

README.md
```diff
@@ -1,17 +1,85 @@
...
+
+-----
+
+## Support
+
+Work on this repository is supported in part by:
 
-<a href="https://anaconda.com/"><img src="https://camo.githubusercontent.com/b8555ef2222598ed37ce38ac86955febbd25de7619931bb7dd3c58432181d3b6/68747470733a2f2f626565776172652e6f72672f636f6d6d756e6974792f6d656d626572732f616e61636f6e64612f616e61636f6e64612d6c617267652e706e67" alt="anaconda logo" width="40%"/></a>
+**"Anaconda, Inc. - Advancing AI through open source."**
```
martindurant, line 85: Interestingly, the tagline and logo keeps changing, but I'm not minded to try to keep up.

martindurant (approved):
Thank you.

README.md
trinadhkotturu: I try to keep it but its not rendering properly. So have to take it out so that Anaconda support doesn't get eclipsed in the non-redended logo.

README.md
ankitaluthra1: for now we route to core.py for metadata calls so it uses aiohttp but we can move those to grpc as well after testing perf benefits

martindurant: OK. But still, the ability to use asynchronous mode is not exactly because of aiohttp - I would just remove it or perhaps say "GCSFS runs async code internally"

PR #803, opened 2026-04-03
Skip cloud credential method when `NO_GCE_CHECK=true`

0x2b3bfa0:
For consistency with https://github.com/googleapis/google-auth-library-python/pull/1610, see https://github.com/fsspec/filesystem_spec/issues/1768 for context.

> ## Summary
> - When auto-detecting credentials (`method=None`), skip the `"cloud"` method if the `NO_GCE_CHECK` environment variable is set to `"true"`
> - Avoids unnecessary GCE metadata server requests in environments where GCE is known to be unavailable
> - Only affects the auto-detection path; explicit `token="cloud"` still works
> 
> ## Test plan
> - [x] Added `test_no_gce_check_skips_cloud` to verify `_connect_cloud` is not called when `NO_GCE_CHECK=true`
> - [ ] Existing credential tests still pass
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gcsfs/credentials.py
+++ b/gcsfs/credentials.py
@@ -322,7 +322,10 @@ class GoogleCredentials:
         ]:
             self._connect_token(method)
         elif method is None:
-            for meth in ["google_default", "cache", "cloud", "anon"]:
+            methods = ["google_default", "cache", "cloud", "anon"]
+            if os.environ.get("NO_GCE_CHECK") == "true":
+                methods.remove("cloud")
+            for meth in methods:
                 try:
                     self.connect(method=meth)
                     logger.debug("Connected with method %s", meth)
--- a/gcsfs/tests/test_credentials.py
+++ b/gcsfs/tests/test_credentials.py
@@ -18,6 +18,20 @@ def test_googlecredentials_none():
     credentials.apply(headers)
 
 
+def test_no_gce_check_skips_cloud():
+    """When NO_GCE_CHECK=true, the 'cloud' method should not be attempted."""
+    with patch.dict(os.environ, {"NO_GCE_CHECK": "true"}):
+        with patch.object(GoogleCredentials, "_connect_cloud") as mock_cloud:
+            # google_default and cloud will fail; anon will succeed
+            with patch.object(
+                GoogleCredentials,
+                "_connect_google_default",
+                side_effect=ValueError("no default"),
+            ):
+                GoogleCredentials(project="myproject", token=None, access="read_only")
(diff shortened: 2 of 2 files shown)


martindurant:
While this clearly gives the option to skip the slow creds step, if a user knows to pass this environment variable, then they probably should know which of the credentials types they need, no?

0x2b3bfa0:
> _if a user knows to pass this environment variable, then they probably should know which of the credentials types they need, no?_

Indeed! However, there might be cases where the choice of enabling this environment variable isn't up to the user, but up to the platform running the user's code, or to some intermediary library. 🫣

martindurant:
Fair enough. @ankitaluthra1 , any thoughts on yet-another-envvar? Perhaps this can be a simple keyword argument (in init), so it can be set via the fsspec config system?

martindurant:
If we added a kwarg to init like `skip_gke_check=False`, then we could document this in a sensible place, and invoke it with env vars alone like:
```
FSSPEC_GCS='{"skip_gke_check": true}'
or
FSSPEC_GCS_SKIP_GKE_CHECK=true
```
or these same settings could be put in an fsspec config file.

What do you think, is this more discoverable? It at least fits in better with the fsspec configuration.

0x2b3bfa0:
My point for using `NO_GCE_CHECK` was that the `google-auth` library already exposes this variable (https://github.com/googleapis/google-auth-library-python/pull/1610).

Given that the current logic relies on both the default `google-auth` behavior, and an extra hardcoded check, the final consumer of `gcsfs` would only be able to avoid this check completely by either:

1. Setting **both** `FSSPEC_GCS_SKIP_GKE_CHECK=true` and `NO_GCE_CHECK=true`
2. Setting `FSSPEC_GCS_SKIP_GKE_CHECK=true` and modifying `gcsfs` to also set `NO_GCE_CHECK=true` internally when calling `google-auth`
3. Setting `NO_GCE_CHECK=true` and modifying `gcsfs` to honor `NO_GCE_CHECK=true` internally (what this pull request currently does)

I don't mind 2 or 3, but would rather prefer to avoid 1.

martindurant:
OK, now I follow.

I note you had the following in the original description:
> Existing credential tests still pass

I take it you have verified this? Of course, it's not something we can test in CI.

0x2b3bfa0:
> _I take it you have verified this?_

Yes; better said, Claude Code has 🤖 😇

PR #805, opened 2026-04-13
Integrate prefetcher engine with zonal buckets

googlyrahman:
This PR integrates the prefetcher engine with Zonal Buckets to improve read throughput and introduces concurrency support for fetching data. To maximize performance and avoid Python's memory and concurrency bottlenecks, this PR also introduces a highly optimized DirectMemmoveBuffer and an MRDPool for managing connection states.

### Zero-Copy, GIL-Free Memory Management (DirectMemmoveBuffer)

Standard Python buffers (like `bytearray` or `io.BytesIO`) were GIL-bound in my experiments and require multiple memory copies. For example, if you want to assemble fifty 2MB chunks into a 100MB object using a `bytearray`, you must create a completely separate 100MB copy when finally retrieving the bytes object because the origin object where you filled the data was `bytesarray`. Furthermore, these standard I/O objects heavily acquire the Global Interpreter Lock (e.g., holding the GIL for 20s during 60s of high I/O).

To solve this, we introduced `DirectMemmoveBuffer`. It bypasses these limitations by utilizing `ctypes.memmove` to write data directly to pre-allocated memory off the main thread, effectively bypassing the GIL and eliminating the final copy overhead.

### Concurrency Support for Zonal Buckets

- Implemented `_concurrent_mrd_fetch`: A new method to handle downloading multiple data chunks concurrently ([ source](https://github.com/fsspec/gcsfs/pull/805/changes?content_ref=async+def+_concurrent_mrd_fetch+self+offset+length+concurrency+mrd_or_pool#diff-5beea29ea8c70df972e3b372a1fe2a9787fe6b10114756283f93d25e6c4d7438)).
- Refactored `_fetch_range_split`: Upgraded this method (used by the readahead_chunked cache) to fully support concurrent chunk fetching ([ source](https://github.com/fsspec/gcsfs/pull/805/changes?content_ref=async+def+test_fetch_range_split#diff-5beea29ea8c70df972e3b372a1fe2a9787fe6b10114756283f93d25e6c4d7438)).
- Refactored `_cat_file`: Updated to support concurrency out of the box ([ source](https://github.com/fsspec/gcsfs/pull/805/changes?content_ref=33+async+def+init_mrd#diff-5beea29ea8c70df972e3b372a1fe2a9787fe6b10114756283f93d25e6c4d7438)).

### Connection Pooling (MRDPool)

Introduced MRDPool to manage a pool of AsyncMultiRangeDownloader (MRD) instances. This pool enables safe and efficient concurrent fetching by reusing connections and scaling downloaders on demand.

### Prefetcher Engine Integration & Heuristic Tuning

- Zonal Bucket Integration: Fully integrated the prefetcher engine into the zonal file read path.

- Tuned Prefetching Trigger (MIN_STREAKS_FOR_PREFETCHING): Modified the prefetching heuristic to start on the 3rd sequential read ( >= 3) instead of the 2nd ( >= 2) ([ source](https://github.com/fsspec/gcsfs/pull/805/changes?content_ref=min_streaks_for_prefetching+3#diff-5beea29ea8c70df972e3b372a1fe2a9787fe6b10114756283f93d25e6c4d7438)). Why? The readahead_chunked cache typically issues two requests to the prefetch engine simultaneously (the required chunk and the readahead chunk). Previously, this immediately and falsely signaled to the prefetcher that the user was performing sequential reads, triggering aggressive prefetching too early. Adjusting the limit to 3 prevents these false positives.


gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,240 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+        Raises:
+            Exception: Bubbles up any exceptions encountered during MRD creation.
+        """
+        spawn_new = False
+        used_from_queue = False
+        mrd = None
+
+        if self._free_mrds.empty():
+            async with self._creation_lock:
+                if not self._free_mrds.empty():
```
zhixiangli: Please take a look at the race conditions, e.g.

1. If the queue contains 1 free MRD and 2 tasks request one, both check empty() and get False.

1. Both skip the _creation_lock block completely.

1. Both will try to `await self._free_mrds.get()`

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,240 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+
+    async def close(self):
+        """
+        Cleanly shut down all MRDs.
+
+        Iterates through all instantiated downloaders and calls their close methods
+        """
+        tasks = []
+        for mrd in self._all_mrds:
+            tasks.append(mrd.close())
```
zhixiangli: Is it possible that some MRDs are being created but not added to `_all_mrds`—and that those MRDs are never closed?

gcsfs/zb_hns_utils.py
googlyrahman: The `MRDPool` class isn't designed to be thread-safe. As you can see in the `__init__` function, we are using `asyncio` locks. It is task safe, and not thread-safe.

Because of how concurrency programming works in python, two coroutines cannot check `self.queue.empty()` at the exact same time. Unlike standard threading which preempts randomly, `asyncio` uses cooperative multitasking. Coroutines only yield control at explicit await points.

This means `self.queue.empty()` and `self._free_mrds.get()` will execute together without being interrupted. If there is an element in `_free_mrds`, it will be returned immediately before a second coroutine even gets a chance to run.

Whether the class should be thread-safe instead of just task-safe is a separate discussion. For the `gcsfs` use case, everything runs in a single event loop fsspec thread, hence task-safe was enough to implement, Additionally asyncio primitives such as lock has lower overhead compared to threading primitives.

gcsfs/zb_hns_utils.py
googlyrahman: Ideally all newly created mrds should be in `_all_mrds` as `create_mrd` puts it into `_all_mrds`, Let me know if you feel otherwise.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,240 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                this boundary will be rejected to prevent overflows.
+            executor (concurrent.futures.Executor): The thread pool executor to run the
+                memmove operations.
+            max_pending (int, optional): The maximum number of pending write operations
+                allowed in the queue. Defaults to 5.
+        """
+        self.start_address = start_address
+        self.end_address = end_address
+        self.current_offset = 0
+        self.semaphore = threading.Semaphore(max_pending)
```
zhixiangli, line 207: What problem is semaphore trying to solve? Since memmove is cheap and doesn't require extra memory, why is it slowing down the process?

gcsfs/zb_hns_utils.py
googlyrahman: Thinking about it more, this could happen if a user runs `get_mrd` and `close` concurrently. That is a highly unlikely pattern and does not happen in gcsfs because we always await `get_mrd` before calling close. If they were run concurrently, close might execute before `get_mrd` actually finishes creating the MRD.

While it is an anti-pattern, I think we can prevent this. It would be good to introduce a closed variable and check it before starting a new MRD.

gcsfs/extended_gcsfs.py
```diff
@@ -67,6 +104,10 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
         if self.credentials.token == "anon":
             self.credential = AnonymousCredentials()
         self._storage_layout_cache = {}
+        self.memmove_executor = ThreadPoolExecutor(
```
zhixiangli: Nit: Please use `_memmove_executor` since it is for internal use only and should not be exposed to users.

gcsfs/zb_hns_utils.py
zhixiangli: OK, this assumes there are no `await` statements in between, otherwise the logic may break.

googlyrahman: Yes, that is correct. That was my rationale when I implemented this code. 

However, thinking about maintainability, it might be better to keep the lock at the top level so the coroutines do not interfere. Otherwise, someone unfamiliar with this setup could easily introduce bugs in the future. I'll probably move the lock at the top of the method.

gcsfs/zb_hns_utils.py
googlyrahman: Will add a close variable in MRDPool to avoid anti pattern requests.

gcsfs/zb_hns_utils.py
googlyrahman: The semaphore prevents too many offloading to worker threads.

We don't want to have a situation where our workers are pulling the data from the network without ensuring it's existence in the buffer. Hence this semaphore.

See the usage in the code.

gcsfs/extended_gcsfs.py
googlyrahman: Aah, that's correct! Thanks

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,147 @@ async def _is_zonal_bucket(self, bucket):
...
-                await zb_hns_utils.close_mrd(mrd)
+            if pool_created_here:
+                await mrd.close()
+
+    async def _concurrent_mrd_fetch(self, offset, length, concurrency, mrd_or_pool):
+        """Helper to handle concurrent chunk downloads into a DirectMemmoveBuffer."""
+        result_bytes = PyBytes_FromStringAndSize(None, length)
+        buffer_ptr = PyBytes_AsString(result_bytes)
+
+        part_size = length // concurrency
```
zhixiangli, line 365: When the length is less than the concurrency, the part_size becomes 0, which triggers an unnecessary _download operation with a size of 0. The MRD treats reading 0 bytes as the EOF.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,251 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+
+            self.current_offset += size
+        data_bytes = bytes(data) if not isinstance(data, bytes) else data
+
+        self.semaphore.acquire()
+        with self._lock:
+            if self._pending_count == 0:
+                self._done_event.clear()
+            self._pending_count += 1
+        return self.executor.submit(self._do_memmove, dest, data_bytes, size)
```
zhixiangli, line 254: If the `submit` raises an error, _do_memmove will not execute, causing the close operation to hang indefinitely.

gcsfs/zb_hns_utils.py
zhixiangli: Is it to limit the parallel runs of memmove since the GIL is released?

gcsfs/extended_gcsfs.py
```diff
@@ -67,6 +104,10 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
         if self.credentials.token == "anon":
             self.credential = AnonymousCredentials()
         self._storage_layout_cache = {}
+        self._memmove_executor = ThreadPoolExecutor(
+            max_workers=kwargs.get("memmove_max_workers", 8)
+        )
+        weakref.finalize(self, self._memmove_executor.shutdown)
```
zhixiangli, line 110: Does the `_Cached` meta class hold a strong reference to the `GCSFSSystem`? If so, `weakref.finalize` will never be fired, correct?

gcsfs/extended_gcsfs.py
martindurant: The part size should never be too small anyway, because at that point you're just adding overhead. So we need a reasonable minimum size here.

gcsfs/extended_gcsfs.py
martindurant: It is a strong reference by default, weak by option.
This will still run, but only at interpreter shutdown.

googlyrahman: Resolving this, as Martin already responded.

gcsfs/zb_hns_utils.py
googlyrahman: I’m not sure why `submit` would raise error here. The documentation doesn't mention any errors for `submit`. https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit

Do you see any reason why `submit` would fail here?

gcsfs/extended_gcsfs.py
googlyrahman: > The MRD treats reading 0 bytes as the EOF

Isn't this a bug in SDK? if user passes buffer of size 0, it should not consider it as EOF - it should rather ignore it, and do not fill anything in buffer.

> which triggers an unnecessary _download operation with a size of 0
>> The part size should never be too small anyway, because at that point you're just adding overhead. So we need a reasonable minimum size here.

Yeah, we've this check in standard bucket, where we call the `_cat_file_sequential` for chunk size less than 5MB, i think i missed it here. Thanks for pointing it.

gcsfs/zb_hns_utils.py
googlyrahman: Yes, it is for limiting memmove operations in parallel per buffer. However, I view this primarily as a way to limit the maximum number of pending writes to that buffer.

Since `buffer.write(data)` is time-consuming in general, this setup allows you to continue your operation ( pulling data from the network ) while GIL-free workers handle the memory writes. We shouldn't always continue pulling from the network without ensuring previous chunks are being handled; therefore, we use the `max_pending` variable and a semaphore on that variable to manage that limit.

gcsfs/zb_hns_utils.py
martindurant: I'm with @googlyrahman , I don't think there can be a state where _do_memove is instantiated but somehow submit fails to return. If the executor is closed, you would error immediately. You _might_ have to wait, if the executor is busy.

martindurant:
> if user passes buffer of size 0, it should not consider it as EOF

Good idea to check what happens with read(0), though - does it count as sequential? Can enough of them make the prefetch size go to 0?

googlyrahman:
> Good idea to check what happens with read(0), though - does it count as sequential? Can enough of them make the prefetch size go to 0?

It doesn't count as sequential and neither prefetch size goes to zero, as we [return early](https://github.com/ankitaluthra1/gcsfs/blob/de3a928a26faea384f7bba423a9881c96db3b95d/gcsfs/prefetcher.py#L618
) with an empty bytes in case of `size <= 0` (aka `start>=end`)

gcsfs/zb_hns_utils.py
zhixiangli: > If the executor is closed, you would error immediately.

If submit then raises a `RuntimeError`, `_do_memmove` never runs. This means `self._pending_count` is never decremented and `self._done_event` is never set back to true.

Consequently, if the user subsequently calls `close()`, it will hang indefinitely on `self._done_event.wait()`.

zhixiangli:
> [docs/readthedocs.org:gcsfs](https://app.readthedocs.org/projects/gcsfs/builds/32349243/) — Read the Docs build failed!

Rebasing from main may address this error.

gcsfs/extended_gcsfs.py
zhixiangli: I don't have much context, but it's most likely not a bug, since this is a special case mentioned in the document - https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-storage/google/cloud/storage/asyncio/async_multi_range_downloader.py#L391-L396, I think we should clean it up. We can avoid redundant calls by preventing an actual size of 0; this change should be cheap to implement.

gcsfs/zb_hns_utils.py
googlyrahman: Hmm, so this is about executor shutdowns mid-operation. In the context of `gcsfs`, it doesn't happen because the executor shutdown occurs when `ExtendedGCSFileSystem` is garbage collected, and it would not be garbage collected until `ZonalFile` is garbage collected, as `ZonalFile` holds a reference to `ExtendedGCSFileSystem`.

But yes, outside of `gcsfs`, if someone uses this directly and passes a bad `executor`, the `write()` would fail immediately. If someone catches that exception and further calls `close()`, it would hang.

I think it makes sense to release the semaphore, decrement the pending count here, and set the event if the pending count reaches zero to handle the case. I'll add a commit for this.

Good catch!

gcsfs/zonal_file.py
```diff
@@ -59,19 +60,21 @@ def __init__(
...
+        self.pool_size = pool_size
         object_size = None
         if "r" in self.mode:
-            self.mrd = asyn.sync(
-                self.gcsfs.loop, self._init_mrd, bucket, key, generation
+            self.mrd_pool = zb_hns_utils.MRDPool(
+                self.gcsfs, bucket, key, generation, self.pool_size
             )
-            object_size = self.mrd.persisted_size
+            asyn.sync(self.gcsfs.loop, self.mrd_pool.initialize)
```
martindurant, line 75: Shouldn't this only be done on first write?

gcsfs/zonal_file.py
```diff
@@ -38,6 +38,7 @@ def __init__(
         fixed_key_metadata=None,
         generation=None,
         kms_key_name=None,
+        pool_size=zb_hns_utils.DEFAULT_CONCURRENCY,
```
martindurant, line 41: Elsewhere instead of "pool", "concurrency" is the preferred term, right?

gcsfs/zonal_file.py
```diff
@@ -162,23 +165,61 @@ def _fetch_range(self, start=None, end=None, chunk_lengths=None):
                 "The end and chunk_lengths arguments are mutually exclusive and cannot be used together."
             )
 
-        try:
+        if self._prefetch_engine:
+            try:
+                if chunk_lengths is None:
```
martindurant, line 191: Document `chunk_lengths` in the signature?

gcsfs/zonal_file.py
```diff
@@ -162,23 +165,61 @@ def _fetch_range(self, start=None, end=None, chunk_lengths=None):
                 "The end and chunk_lengths arguments are mutually exclusive and cannot be used together."
             )
 
-        try:
+        if self._prefetch_engine:
```
martindurant, line 186: Here is where the cache/prefetch are crossing over - we hope to clean this up in the future, so a comment saying something to that effect would be nice.

gcsfs/zonal_file.py
```diff
@@ -311,10 +352,12 @@ def close(self):
         """
         if self.closed:
             return
+
         # super is closed before aaow since flush may need aaow
         super().close()
-        # Helper method safely handles mrd=None.
-        asyn.sync(self.gcsfs.loop, zb_hns_utils.close_mrd, self.mrd)
+
+        if hasattr(self, "mrd_pool"):
```
martindurant: A fragile condition?

gcsfs/zonal_file.py
```diff
@@ -162,23 +165,61 @@ def _fetch_range(self, start=None, end=None, chunk_lengths=None):
...
+                    concurrency=self.concurrency,
                     start=start,
                     chunk_lengths=chunk_lengths,
                     size=self.size,
-                    mrd=self.mrd,
+                    mrd=self.mrd_pool,
                 )
-            return self.gcsfs.cat_file(self.path, start=start, end=end, mrd=self.mrd)
+
+            return await self.gcsfs._cat_file(
```
martindurant, line 224:
```suggestion
            # non-prefetch route
            return await self.gcsfs._cat_file(
```

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
-                    offset=offset, length=length, mrd=mrd
-                )
-                return [data]
+        try:
+            results = await asyncio.gather(*tasks, return_exceptions=True)
+            for res in results:
+                if isinstance(res, Exception):
+                    has_error = True
+                    raise res
+        except BaseException:
```
martindurant, line 456: Can this be anything other that the `raise` on the line above? I wonder why `BaseException`, and why we need to set has_error again.

The code here for the types of error is a bit convoluted (as per the comment below), but I have no better idea.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+
+
+class DirectMemmoveBuffer:
+    """
+    A buffer-like object that writes data directly to memory asynchronously.
+
+    This class provides a `write` interface that queues `ctypes.memmove` operations
+    to a thread pool executor, limiting the maximum number of concurrent pending
+    writes using a semaphore. It is useful for high-performance data transfers
+    where memory copies need to be offloaded from the main thread.
```
martindurant, line 188: I am feeling like we may eventually want to spin out a low-level (C) compiled library implementing this kind of functionality.  Moving it here is a good start.

cf: https://docs.rs/cramjam/latest/cramjam/io/struct.RustyBuffer.html

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                allowed in the queue. Defaults to 5.
+        """
+        self.start_address = start_address
+        self.end_address = end_address
+        self.current_offset = 0
+        self.semaphore = threading.Semaphore(max_pending)
+        self._error = None
+        self._pending_count = 0
+        self._lock = threading.Lock()
+        self._done_event = threading.Event()
```
martindurant, line 219: Do we need profiling to determine if the various lock-based objects can cause degradation if, for instance, many IOPS are happening in multiple application threads?

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+        self._done_event.set()
+        self.executor = executor
+
+    def write(self, data):
+        """
+        Schedules a write operation to memory.
+
+        Calculates the destination address based on the current offset, increments the offset,
+        and submits the memory move operation to the executor. Blocks if the number of
```
martindurant, line 235: Many concurrent.futures -style APIs allow a _nowait variety which errors if the queue is full.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                memory move operation.
+
+        Raises:
+            Exception: If any previous asynchronous write operation encountered an error.
+        """
+        if self._error:
+            raise self._error
+        size = len(data)
+        with self._lock:
+            dest = self.start_address + self.current_offset
```
martindurant, line 254: comment where start_address and current_address are defined, that they are volatile and should only be amended within locks.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+            if dest + size > self.end_address:
+                error_msg = (
+                    f"Attempted to write {size} bytes "
+                    f"at offset {self.current_offset}. "
+                    f"Max capacity is {self.end_address - self.start_address} bytes."
+                )
+                raise BufferError(error_msg)
```
martindurant, line 261: Perhaps set a flag and create the message/error after the block to keep the locked time as short as possible?

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+
+            self.current_offset += size
+        data_bytes = bytes(data) if not isinstance(data, bytes) else data
```
martindurant, line 264: To be revisited: must it always be `bytes`? This will cause a copy here if something non-bytes is passed, even if we are in a position to guarantee that the buffer in question cannot change.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+
+        self.semaphore.acquire()
+        with self._lock:
+            if self._pending_count == 0:
+                self._done_event.clear()
+            self._pending_count += 1
```
martindurant, line 270: I would be worth explaining in complete detail the purpose of the semaphore versus done_event versus pending_count. You have a state machine here and the locks are needed to keep the variables consistent, but by laying out the rules, perhaps if could be found to be cleaner.

googlyrahman:
@martindurant please review this

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                self._pending_count -= 1
+                if self._pending_count == 0:
+                    self._done_event.set()
+            raise e
+
+    def _do_memmove(self, dest, data_bytes, size):
+        try:
+            ctypes.memmove(dest, data_bytes, size)
+        except Exception as e:
+            self._error = e
```
martindurant, line 283: You _could_ perhaps set the error directly in the other thread with https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetObject rather than waiting for the loop to pick it up. 

Is it expected that this Buffer is only ever called from async code?

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+            raise self._error
+
+        expected_size = self.end_address - self.start_address
+        if self.current_offset < expected_size:
+            error_msg = (
+                f"Expected {expected_size} bytes, "
+                f"but only received {self.current_offset} bytes. "
+                f"Buffer contains uninitialized data."
+            )
+            raise BufferError(error_msg)
```
martindurant, line 309: Explicitly check executor for leftover tasks?

We are assuming the executor outlives the buffer, to be used again. When does it close?

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,261 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+        try:
+            return self.executor.submit(self._do_memmove, dest, data_bytes, size)
+        except BaseException as e:
+            self._error = e
+            self.semaphore.release()
+            with self._lock:
+                self._pending_count -= 1
+                if self._pending_count == 0:
+                    self._done_event.set()
+            raise e
```
martindurant, line 277: perhaps consolidate the repeated patterns

gcsfs/extended_gcsfs.py
```diff
@@ -67,6 +104,10 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
         if self.credentials.token == "anon":
             self.credential = AnonymousCredentials()
         self._storage_layout_cache = {}
+        self._memmove_executor = ThreadPoolExecutor(
```
martindurant, line 129: Don'e we end up with one executor per filesystem instance? That could be a lot of workers in some circumstances.

gcsfs/extended_gcsfs.py
```diff
@@ -67,6 +104,10 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
...
+            max_workers=kwargs.get("memmove_max_workers", 8)
```
martindurant, line 130: Why hide the kwarg instead of documenting it and showing the default in the signature?

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
+        start,
+        chunk_lengths,
+        concurrency,
+        mrd=None,
+        size=None,
+        **kwargs,
     ):
         """
-        Reading multiple reads in one large stream.
+        Reading multiple disjoint ranges concurrently.
```
martindurant: = `_cat_ranges(), but for one path? I'm not sure we need the duplication.

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
-        Optimized for Zonal Buckets:
-        Leverages AsyncMultiRangeDownloader.download_ranges() to fetch all requested
-        'chunk_lengths' (chunks) concurrently in a single batch request, significantly
-        improving performance for ReadAheadV2.
+        Delegates concurrent fetching of individual chunks directly to `_cat_file`.
         """
+        file_size = size or await _get_mrd_size(mrd)
+        if file_size is None:
+            logger.warning(
+                f"AsyncMultiRangeDownloader (MRD) for {path} has no 'persisted_size'. "
```
martindurant, line 346: Why does this happen?

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
-                    f"AsyncMultiRangeDownloader (MRD) for {path} has no 'persisted_size'. "
-                    "Falling back to _info() to get the file size."
+        try:
+            tasks = []
+            current_offset = start_offset
+
+            cat_kwargs = kwargs.copy()
+
+            for length in chunk_lengths:
+                end_offset = current_offset + length
```
martindurant, line 374: So the chunks are actually adjacent, as opposed to disjoint?

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
+                tasks.append(
+                    asyncio.create_task(
+                        self._cat_file(
+                            path,
+                            start=current_offset,
+                            end=end_offset,
+                            mrd=mrd,
+                            concurrency=max(
+                                1, length * concurrency // sum(chunk_lengths)
+                            ),
```
martindurant, line 388: Explain: concurrency on the (already concurrent) tasks is typically 1, but may be higher if some chunks dominate the total.

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
 
-                read_ranges = []  # To pass to MRD
+            return results
 
-                for length in chunk_lengths:
-                    read_ranges.append((current_offset, length))
-                    current_offset += length
+        finally:
+            if pool_created_here:
+                await mrd.close()
```
martindurant, line 411: Do we not need to fetch the exceptions from gather this time?

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
-                return await zb_hns_utils.download_ranges(read_ranges, mrd)
-            else:
-                end = kwargs.get("end")
-                offset, length = await self._process_limits_to_offset_and_length(
-                    path, start, end, file_size
-                )
+    async def _concurrent_mrd_fetch(self, offset, length, concurrency, mrd_or_pool):
+        """Helper to handle concurrent chunk downloads into a DirectMemmoveBuffer."""
+        concurrency = (
+            concurrency if length >= self.MIN_CHUNK_SIZE_FOR_CONCURRENCY else 1
```
martindurant, line 416: If concurrency is 1, can we fastpath here?

gcsfs/extended_gcsfs.py
```diff
@@ -241,73 +282,150 @@ async def _is_zonal_bucket(self, bucket):
...
+
+        part_size = length // concurrency
+        tasks = []
+        buffers = []
+        loop = asyncio.get_running_loop()
+
+        # Track if the core download process failed
+        has_error = False
+
+        async def _download(o, s, b):
```
martindurant: pass mrd_or_pool to avoid using the closure?

martindurant (commented):
I have read the code but not the tests so far. 

Again, there is a lot of complexity here...

gcsfs/extended_gcsfs.py
googlyrahman: This happens when the passed MRD is `uninitialized`. Calling `_info` directly is much faster than going through a full `initialization` and `close` cycle. Note that this condition was already present in the existing code.

gcsfs/extended_gcsfs.py
googlyrahman: Yeah, i think adjacent is good wording here rather than disjoint - I'll replace the disjoint with adjacent.

gcsfs/extended_gcsfs.py
googlyrahman: Adding a comment.

gcsfs/extended_gcsfs.py
googlyrahman: what is the fast path here?

gcsfs/zonal_file.py
googlyrahman: Yes, we have two terms: pool and concurrency. The pool represents the number of underlying MRDs, while concurrency represents the number of splits we need to make for downloading.

gcsfs/zonal_file.py
googlyrahman: Ideally, yes, but take a look at the existing code. `ZonalFile` initializes the MRD, gets its persisted size, and passes it to super(). I just kept the workflow as is while integrating the pool.

I think it was done for some reason here https://github.com/fsspec/gcsfs/pull/744

gcsfs/zonal_file.py
googlyrahman: Adding a comment

gcsfs/zonal_file.py
googlyrahman: Addressing it.

gcsfs/zonal_file.py
googlyrahman: Added the comment

gcsfs/zonal_file.py
googlyrahman: Yeah, This should be `hasattr(self, "mrd_pool") and self.mrd_pool`

gcsfs/zb_hns_utils.py
googlyrahman: It would be bytes in case of `gcsfs`.

gcsfs/zb_hns_utils.py
googlyrahman: Good point on keeping the lock time minimal! However, I think we can safely leave the string building and error raising inside the block in this specific case.

The error message is only generated on the exceptional path when the buffer overflows. If this happens, it immediately raises a `BufferError` and halts execution anyway. Since it’s a terminal 'sad path,' lock contention at that exact moment isn't a performance concern.

Skipping the extra flag keeps the code simpler without impacting the performance of the 'happy path' we actually care about.

gcsfs/zb_hns_utils.py
googlyrahman: We currently do not have a use case of this in `gcsfs`. We can explore this in future if required.

gcsfs/zb_hns_utils.py
googlyrahman: Adding an explanation in __init__ method.

gcsfs/zb_hns_utils.py
googlyrahman: Removing the duplication of code.

gcsfs/zb_hns_utils.py
googlyrahman: I agree, We were also thinking along the similar lines few weeks back, and have a internal action item for this.

The `RustyBuffer` you shared is something which needs benchmarking, i'm seeing it for the first time, will see if it is useful and update here.

gcsfs/zb_hns_utils.py
googlyrahman: Adding comment

gcsfs/zb_hns_utils.py
googlyrahman: > Is it expected that this Buffer is only ever called from async code?

Yes, currently our usecase expects this method to be called from async code in synchronous manner. 

> You could perhaps set the error directly in the other thread with https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetObject rather than waiting for the loop to pick it up.

I haven't used these methods before, but it doesn't look like it will improve performance since it only happens on the error path. Switching the architecture to this method would require benchmarking again, which won't yield any performance gains. So, unless there is a problem with the current approach, I think it is fine as is.

gcsfs/zb_hns_utils.py
googlyrahman: > Explicitly check executor for leftover tasks?

`self._done_event.wait()` would do that, That event only get set once there's no pending task.

> We are assuming the executor outlives the buffer, to be used again. When does it close?

The executor gets closed on garbage collection of `ExtendedGCSFileSystem`

gcsfs/extended_gcsfs.py
martindurant: You don't need to calculate ranges, make tasks or use a memory copier if the chunk you want is the final output anyway, right? You don't even really need an a mrd pool (but would use one if it already exists).

gcsfs/zb_hns_utils.py
googlyrahman: The numbers are good in case of multithreading, Here are the numbers with DirectMemmoveBuffer (Implemented here), io.BytesIO (current SDK default and python native), RustyBuffer (shared above by you)

Notes:
- While you may see it being slow for smaller workloads <= 1MB, but there the bottleneck is not assembly, or the cpu - it's the network. We've tested it with 64KB io size, and do not see any degradation in throughput of gcsfs. The network call itself takes so many ms, there 0.015ms of difference addition would not show any gains. This would be 0.3% of difference if it takes 5ms of time to fetch 512KB.
- The big chunk sizes shows massive gains under multithreading environment, and it is reflected even in our benchmarks. this pushed our throughput to 2.2 GB/s - 2.3 GB/s from existing 600 - 700 MB/s on concurrency=4
- The RustyBuffer performed worse in any scenario.
- The GRPC pass SDK chunks in 2MB payload, so even if you pass the 100MB workload, you would actually get in 2MB * 50 - The benchmarks written below are designed with this behaviour.

```
=====================================================================================
Testing Scenario: 512 KB Request (Tiny Payload)
=====================================================================================

Threads: 1 | Total Processed per cycle: 0.500 MB
DirectMemmoveBuffer      :   0.030 ms/payload | 16,514.32 MB/s | 0.84x parallelism
io.BytesIO (Native)      :   0.014 ms/payload | 35,207.41 MB/s | 1.01x parallelism
cramjam.Buffer (Rust)    :   0.098 ms/payload |  5,124.59 MB/s | 1.00x parallelism

Threads: 4 | Total Processed per cycle: 2.000 MB
DirectMemmoveBuffer      :   0.079 ms/payload | 25,168.95 MB/s | 1.89x parallelism
io.BytesIO (Native)      :   0.054 ms/payload | 37,003.21 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   0.413 ms/payload |  4,842.82 MB/s | 1.11x parallelism

Threads: 16 | Total Processed per cycle: 8.000 MB
DirectMemmoveBuffer      :   0.357 ms/payload | 22,432.45 MB/s | 1.77x parallelism
io.BytesIO (Native)      :   0.230 ms/payload | 34,736.17 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   1.530 ms/payload |  5,229.09 MB/s | 1.02x parallelism

Threads: 48 | Total Processed per cycle: 24.000 MB
DirectMemmoveBuffer      :   1.147 ms/payload | 20,915.99 MB/s | 1.83x parallelism
io.BytesIO (Native)      :   0.738 ms/payload | 32,499.24 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   4.570 ms/payload |  5,252.21 MB/s | 1.03x parallelism

=====================================================================================
Testing Scenario: 1 MB Request (Less than 1 chunk)
=====================================================================================

Threads: 1 | Total Processed per cycle: 5.000 MB
DirectMemmoveBuffer      :   0.269 ms/payload | 18,558.53 MB/s | 2.08x parallelism
io.BytesIO (Native)      :   0.341 ms/payload | 14,654.59 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   1.473 ms/payload |  3,393.45 MB/s | 1.00x parallelism

Threads: 4 | Total Processed per cycle: 20.000 MB
DirectMemmoveBuffer      :   0.572 ms/payload | 34,982.43 MB/s | 3.55x parallelism
io.BytesIO (Native)      :   1.648 ms/payload | 12,133.22 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   5.756 ms/payload |  3,474.41 MB/s | 1.00x parallelism

Threads: 16 | Total Processed per cycle: 80.000 MB
DirectMemmoveBuffer      :   3.035 ms/payload | 26,358.66 MB/s | 2.98x parallelism
io.BytesIO (Native)      :   6.126 ms/payload | 13,060.00 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :  29.760 ms/payload |  2,688.19 MB/s | 1.05x parallelism

Threads: 48 | Total Processed per cycle: 240.000 MB
DirectMemmoveBuffer      :   9.465 ms/payload | 25,357.65 MB/s | 3.31x parallelism
io.BytesIO (Native)      :  14.472 ms/payload | 16,583.22 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :  89.473 ms/payload |  2,682.39 MB/s | 1.04x parallelism

=====================================================================================
Testing Scenario: 5 MB Request (2 full chunk, 1 half chunk)
=====================================================================================

Threads: 1 | Total Processed per cycle: 5.000 MB
DirectMemmoveBuffer      :   0.251 ms/payload | 19,939.43 MB/s | 2.03x parallelism
io.BytesIO (Native)      :   0.308 ms/payload | 16,237.56 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   1.402 ms/payload |  3,567.24 MB/s | 1.01x parallelism

Threads: 4 | Total Processed per cycle: 20.000 MB
DirectMemmoveBuffer      :   0.620 ms/payload | 32,244.91 MB/s | 3.43x parallelism
io.BytesIO (Native)      :   1.259 ms/payload | 15,886.71 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   5.817 ms/payload |  3,437.98 MB/s | 1.01x parallelism

Threads: 16 | Total Processed per cycle: 80.000 MB
DirectMemmoveBuffer      :   3.154 ms/payload | 25,368.11 MB/s | 2.90x parallelism
io.BytesIO (Native)      :   6.246 ms/payload | 12,808.86 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :  31.844 ms/payload |  2,512.28 MB/s | 1.05x parallelism

Threads: 48 | Total Processed per cycle: 240.000 MB
DirectMemmoveBuffer      :   9.691 ms/payload | 24,764.66 MB/s | 3.18x parallelism
io.BytesIO (Native)      :  17.377 ms/payload | 13,811.61 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    : 104.878 ms/payload |  2,288.38 MB/s | 1.07x parallelism

=====================================================================================
Testing Scenario: 10 MB Request (5 full chunks) - See multithreading here
=====================================================================================

Threads: 1 | Total Processed per cycle: 10.000 MB
DirectMemmoveBuffer      :   0.367 ms/payload | 27,253.25 MB/s | 2.74x parallelism
io.BytesIO (Native)      :   0.663 ms/payload | 15,092.21 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :   3.107 ms/payload |  3,218.11 MB/s | 1.00x parallelism

Threads: 4 | Total Processed per cycle: 40.000 MB
DirectMemmoveBuffer      :   0.907 ms/payload | 44,101.19 MB/s | 4.19x parallelism
io.BytesIO (Native)      :   3.203 ms/payload | 12,487.88 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :  13.744 ms/payload |  2,910.31 MB/s | 1.01x parallelism

Threads: 16 | Total Processed per cycle: 160.000 MB
DirectMemmoveBuffer      :   4.693 ms/payload | 34,092.17 MB/s | 3.92x parallelism
io.BytesIO (Native)      :  11.076 ms/payload | 14,445.87 MB/s | 1.01x parallelism
cramjam.Buffer (Rust)    :  74.503 ms/payload |  2,147.57 MB/s | 1.04x parallelism

Threads: 48 | Total Processed per cycle: 480.000 MB
DirectMemmoveBuffer      :  15.496 ms/payload | 30,976.08 MB/s | 4.22x parallelism
io.BytesIO (Native)      :  28.639 ms/payload | 16,760.64 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    : 232.991 ms/payload |  2,060.16 MB/s | 1.06x parallelism

=====================================================================================
Testing Scenario: 100 MB Request (50 full chunks) - See multithreading here
=====================================================================================

Threads: 1 | Total Processed per cycle: 100.000 MB
DirectMemmoveBuffer      :   4.294 ms/payload | 23,290.08 MB/s | 5.78x parallelism
io.BytesIO (Native)      :  28.687 ms/payload |  3,485.88 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    :  81.647 ms/payload |  1,224.79 MB/s | 1.00x parallelism

Threads: 4 | Total Processed per cycle: 400.000 MB
DirectMemmoveBuffer      :  14.336 ms/payload | 27,902.40 MB/s | 8.95x parallelism
io.BytesIO (Native)      : 127.689 ms/payload |  3,132.61 MB/s | 1.00x parallelism
cramjam.Buffer (Rust)    : 351.918 ms/payload |  1,136.63 MB/s | 1.01x parallelism

Threads: 16 | Total Processed per cycle: 1600.000 MB
DirectMemmoveBuffer      :  71.464 ms/payload | 22,389.03 MB/s | 8.89x parallelism
io.BytesIO (Native)      : 523.191 ms/payload |  3,058.15 MB/s | 1.01x parallelism
cramjam.Buffer (Rust)    : 1463.646 ms/payload |  1,093.16 MB/s | 1.03x parallelism

Threads: 48 | Total Processed per cycle: 4800.000 MB
DirectMemmoveBuffer      : 227.062 ms/payload | 21,139.63 MB/s | 8.29x parallelism
io.BytesIO (Native)      : 1580.193 ms/payload |  3,037.60 MB/s | 1.02x parallelism
cramjam.Buffer (Rust)    : 4464.484 ms/payload |  1,075.15 MB/s | 1.06x parallelism
```


Script for your reference:

```
import ctypes
import io
import threading
import time
from concurrent.futures import ThreadPoolExecutor

try:
    from cramjam import Buffer
except ImportError:
    Buffer = None
    print("Warning: 'cramjam' is not installed. Cramjam tests will be skipped.")
    print("Run 'pip install cramjam' to enable them.\n")

# -----------------------------------------------------------------------------
# CPython API Setup
# -----------------------------------------------------------------------------
PyBytes_FromStringAndSize = ctypes.pythonapi.PyBytes_FromStringAndSize
PyBytes_FromStringAndSize.argtypes = (ctypes.c_void_p, ctypes.c_ssize_t)
PyBytes_FromStringAndSize.restype = ctypes.py_object

PyBytes_AsString = ctypes.pythonapi.PyBytes_AsString
PyBytes_AsString.argtypes = (ctypes.py_object,)
PyBytes_AsString.restype = ctypes.c_void_p

# -----------------------------------------------------------------------------
# Buffer Implementation (Original Async Version)
# -----------------------------------------------------------------------------
class DirectMemmoveBuffer:
    def __init__(self, start_address, end_address, executor, max_pending=5):
        self.start_address = start_address
        self.end_address = end_address
        self.executor = executor
        self.current_offset = 0
        self._pending_count = 0
        self._error = None

        self.semaphore = threading.Semaphore(max_pending)
        self._lock = threading.Lock()
        self._done_event = threading.Event()
        self._done_event.set()

    def _decrement_pending(self):
        self.semaphore.release()
        with self._lock:
            self._pending_count -= 1
            if self._pending_count == 0:
                self._done_event.set()

    def write(self, data):
        if self._error:
            raise self._error

        size = len(data)
        with self._lock:
            dest = self.start_address + self.current_offset
            if dest + size > self.end_address:
                raise BufferError("Buffer overflow")
            self.current_offset += size
        
        data_bytes = bytes(data) if not isinstance(data, bytes) else data

        self.semaphore.acquire()
        with self._lock:
            if self._pending_count == 0:
                self._done_event.clear()
            self._pending_count += 1

        try:
            if size <= 512 * 1024:
                self._do_memmove(dest, data_bytes, size)
                return None
            else:
                return self.executor.submit(self._do_memmove, dest, data_bytes, size)
        except BaseException as e:
            self._error = e
            self._decrement_pending()
            raise e

    def _do_memmove(self, dest, data_bytes, size):
        try:
            ctypes.memmove(dest, data_bytes, size)
        except Exception as e:
            self._error = e
            raise e
        finally:
            self._decrement_pending()

    def close(self):
        self._done_event.wait()
        if self._error:
            raise self._error

# -----------------------------------------------------------------------------
# Chunked Backend Simulation Benchmarks
# -----------------------------------------------------------------------------
def run_chunked_backend_benchmark():
    # Simulate a backend returning exactly 2MB chunks
    BACKEND_CHUNK_SIZE = 2 * 1024 * 1024 
    dummy_2mb_chunk = b'x' * BACKEND_CHUNK_SIZE

    # GRPC RETURNS DATA IN 2MB chunk size.
    request_sizes = [
        (512 * 1024, "512 KB Request (Tiny Payload)"), 
        (5 * 1024 * 1024, "1 MB Request (Less than 1 chunk)"), 
        (5 * 1024 * 1024, "5 MB Request (2 full chunk, 1 half chunk)"), 
        (10 * 1024 * 1024, "10 MB Request (5 full chunks) - See multithreading here"),
        (100 * 1024 * 1024, "100 MB Request (50 full chunks) - See multithreading here")
    ]
    
    thread_counts = [1, 4, 16, 48]
    
    for total_requested, size_label in request_sizes:
        print(f"\n{'='*85}")
        print(f"Testing Scenario: {size_label}")
        print(f"{'='*85}")

        # Adjust cycles so the benchmark finishes in a reasonable time
        if total_requested <= 1 * 1024 * 1024:
            num_cycles = 200
        elif total_requested <= 10 * 1024 * 1024:
            num_cycles = 50
        else:
            num_cycles = 10 

        full_chunks = total_requested // BACKEND_CHUNK_SIZE
        remainder = total_requested % BACKEND_CHUNK_SIZE
        remainder_chunk = b'x' * remainder if remainder else b''

        for num_threads in thread_counts:
            total_mb_per_cycle = (total_requested * num_threads) / (1024 * 1024)
            print(f"\nThreads: {num_threads} | Total Processed per cycle: {total_mb_per_cycle:.3f} MB")
            
            # -----------------------------------------------------------------
            # 1. Test DirectMemmoveBuffer 
            # -----------------------------------------------------------------
            def direct_worker():
                executor = ThreadPoolExecutor(max_workers=8)
                
                for _ in range(num_cycles):
                    result_bytes = PyBytes_FromStringAndSize(None, total_requested)
                    start_addr = PyBytes_AsString(result_bytes)
                    
                    buf = DirectMemmoveBuffer(
                        start_address=start_addr,
                        end_address=start_addr + total_requested,
                        executor=executor,
                        max_pending=10
                    )
                    
                    for _ in range(full_chunks):
                        buf.write(dummy_2mb_chunk)
                    if remainder:
                        buf.write(remainder_chunk)
                        
                    buf.close()
                    _ = result_bytes 
                
                executor.shutdown()

            # -----------------------------------------------------------------
            # 2. Test standard io.BytesIO 
            # -----------------------------------------------------------------
            def bytesio_worker():
                for _ in range(num_cycles):
                    buf = io.BytesIO()
                    
                    for _ in range(full_chunks):
                        buf.write(dummy_2mb_chunk)
                    if remainder:
                        buf.write(remainder_chunk)
                    
                    extracted_data = buf.getvalue()
                    _ = extracted_data 
                    buf.close()

            # -----------------------------------------------------------------
            # 3. Test cramjam.Buffer (Rust-backed)
            # -----------------------------------------------------------------
            def cramjam_worker():
                for _ in range(num_cycles):
                    # Initialize the Rust buffer
                    buf = Buffer()
                    
                    # Write the chunks to the Rust stream
                    for _ in range(full_chunks):
                        buf.write(dummy_2mb_chunk)
                    if remainder:
                        buf.write(remainder_chunk)
                    
                    # Reset the cursor to the start and extract all bytes
                    buf.seek(0)
                    extracted_data = buf.read()
                    _ = extracted_data

            # -----------------------------------------------------------------
            # Execution and Measurement Runner
            # -----------------------------------------------------------------
            def run_and_time(name, worker_func):
                start_wall = time.perf_counter()
                start_cpu = time.process_time()
                
                threads = []
                for _ in range(num_threads):
                    t = threading.Thread(target=worker_func)
                    threads.append(t)
                    t.start()
                    
                for t in threads:
                    t.join()
                    
                wall_duration = time.perf_counter() - start_wall
                cpu_duration = time.process_time() - start_cpu
                
                total_mb_processed = (total_mb_per_cycle * num_cycles)
                throughput = total_mb_processed / wall_duration if wall_duration > 0 else 0
                
                parallel_factor = cpu_duration / wall_duration if wall_duration > 0 else 1.0
                avg_latency_ms = (wall_duration / num_cycles) * 1000
                
                print(f"{name.ljust(25)}: {avg_latency_ms:>7.3f} ms/payload | {throughput:>9,.2f} MB/s | {parallel_factor:>4.2f}x parallelism")

            run_and_time("DirectMemmoveBuffer", direct_worker)
            run_and_time("io.BytesIO (Native)", bytesio_worker)
            
            if Buffer:
                run_and_time("cramjam.Buffer (Rust)", cramjam_worker)

if __name__ == "__main__":
    run_chunked_backend_benchmark()
```

gcsfs/zb_hns_utils.py
googlyrahman: RustyBuffer comparision is shared here https://github.com/fsspec/gcsfs/pull/805/changes#r3130946279

Please refer to that thread for more information.

gcsfs/extended_gcsfs.py
googlyrahman: This DirectMemmoveBuffer may not be around for long. I want to avoid introducing a parameter in one release only to deprecate it in the next.

We want to try a C++ wrapper with fsspec, which means we might no longer need DirectMemmoveBuffer. This depends entirely on the experiment.

If the experiment shows a gain, data assembly and other tasks will move to C++. If the experiment does not show a gain, we will make it public as a keyword argument.

gcsfs/extended_gcsfs.py
googlyrahman: Yes, if user creates multiple GCSFileSystem instances in single process. 

Do you want it to a class level executor? This way it would be created once per process.

gcsfs/extended_gcsfs.py
martindurant: (I recommend pure C for this, my opinion. Note that rust via pyo3 is currently incompatible with subinterpreters)

gcsfs/extended_gcsfs.py
googlyrahman: `asyncio.CancelledError` inherits from `BaseException` rather than `Exception`. This means a standard except `Exception:` block will not catch it. If a task is cancelled, any background tasks it spawned will not automatically be cancelled and may be left running. 

To properly clean up and cancel those background tasks, we must explicitly catch `asyncio.CancelledError`. The same principle applies to other BaseException subclasses, such as KeyboardInterrupt.

Why to set `has_error`: We need to close the buffers to ensure that there are no pending writes on buffer. If a network exception has already occurred, The buffer would also throw error saying that we didn't completely filled the buffer. That is exactly why we use `has_error = True`.

- If the download raises an error, we set the flag, call close, and ignore any errors thrown by the closing process. The closing process here will raise an error as buffer filling process was incomplete. See the close method of the buffer.
- If the download is successful, we do not set the flag, call close, and raise any error thrown by the closing process.

gcsfs/extended_gcsfs.py
googlyrahman: Yes, we should - I will add it

gcsfs/extended_gcsfs.py
googlyrahman: Done

gcsfs/extended_gcsfs.py
googlyrahman: Added this in our internal ticket.

gcsfs/extended_gcsfs.py
googlyrahman: Yeah, i agree that this is too much duplicated code here.

We're planning to remove all of this, and move to a downloader class for `Zonal` and `Standard` bucket, and then utilize the downloader in both `_fetch_range_split` and `_cat_file`

This is planned in next followup PR, and we've an internal tracking item for this.

gcsfs/extended_gcsfs.py
googlyrahman: So we need memmove buffers, because prefetcher when fetching multiple blocks, it sends concurrency=1 for each block, because then concurrency is being managed at parent level, so child do not need to be concurrent.

The loop/range calculation are very fast operation, maybe of order `ns`.

The remaining thing is `gather` vs `await t`, I just checked it produces 16.175 µs of difference which isn't much for having a separate branch.

googlyrahman:
@martindurant all your comments are addressed, please take a look again - Thanks for taking time to review this.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+
+    async def initialize(self):
+        """Initializes the MRDPool by creating the first downloader instance."""
+        async with self._lock:
+
+            if self._closed:
+                raise RuntimeError("Cannot initialize a closed MRDPool.")
+
+            if not self._initialized:
+                mrd = await self._create_mrd()
```
martindurant, line 367: lock should apply only to this inner block: if closed or already initialised, we don't need it.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                self.persisted_size = mrd.persisted_size
+                self._free_mrds.put_nowait(mrd)
```
martindurant, line 369: Can raise? We are assuming the q should be empty to start, but it's not enforced here.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                self._active_count += 1
```
martindurant, line 370: Any way to tie active count (or _initialized) to the q size?

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+            if self._free_mrds.empty():
+                if self._active_count < self.pool_size:
+                    self._active_count += 1
+                    spawn_new = True
+                elif self.mrd_supports_multi_request and self._all_mrds:
+                    # Pool is full, queue is empty, and we are allowed to share a busy MRD.
+                    mrd = self._all_mrds[self._rr_index]
+                    self._rr_index = (self._rr_index + 1) % len(self._all_mrds)
+
+            if spawn_new:
```
martindurant: I don't like the word "spawn". We are creating connections, not new execution contexts.

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                used_from_queue = True
+
+        try:
+            yield mrd
+        finally:
+            # Only return the MRD to the free queue if we were the ones who took it out
+            # or if we just spawned it. This prevents duplicate entries in the queue
+            # when multiple concurrent tasks share the same MRD via round-robin.
+            if (spawn_new or used_from_queue) and not self._closed:
+                self._free_mrds.put_nowait(mrd)
```
martindurant, line 426: q could have become full

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                raise RuntimeError("MRDPool is closed.")
+
+            if self._free_mrds.empty():
+                if self._active_count < self.pool_size:
+                    self._active_count += 1
+                    spawn_new = True
+                elif self.mrd_supports_multi_request and self._all_mrds:
+                    # Pool is full, queue is empty, and we are allowed to share a busy MRD.
+                    mrd = self._all_mrds[self._rr_index]
+                    self._rr_index = (self._rr_index + 1) % len(self._all_mrds)
```
martindurant, line 405: _rr_index is undocumented. Round-robin on which connection to use?

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+        Raises:
+            Exception: Bubbles up any exceptions encountered during MRD creation.
+        """
+        spawn_new = False
+        used_from_queue = False
+        mrd = None
+
+        async with self._lock:
+            if self._closed:
+                raise RuntimeError("MRDPool is closed.")
```
martindurant, line 395:
```suggestion
        if self._closed:
            raise RuntimeError("MRDPool is closed.")
        async with self._lock:
```

gcsfs/zb_hns_utils.py
```diff
@@ -163,3 +176,271 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+                return
+
+            tasks = []
+            for mrd in self._all_mrds:
+                tasks.append(mrd.close())
+            try:
+                results = await asyncio.gather(*tasks, return_exceptions=True)
+                for result in results:
+                    if isinstance(result, Exception):
+                        raise result
```
martindurant, line 445: Why not return_exceptions=False if we raise them anyway

gcsfs/zonal_file.py
martindurant: But the pool represents concurrency too

martindurant (commented):
I had a closer look at MRDpool and have a few more comments.
Am I right in thinking that this pool concept should/will be upstreamed to the grpc layer?

gcsfs/zb_hns_utils.py
googlyrahman: I'll document it as round robin.

gcsfs/zb_hns_utils.py
googlyrahman: I'll change it to `create_new` variable

gcsfs/zb_hns_utils.py
googlyrahman: The number of elements in `_free_mrds` will never exceed the `pool_size`, so this scenario is mathematically impossible. We only ever place an MRD into the queue if we just took it out, or if it is a newly created connection. Since the total number of MRDs we create is strictly capped at `pool_size`, the queue can never become full.

gcsfs/zb_hns_utils.py
googlyrahman: If we set return_exceptions=False, the cleanup will fail fast on the first error and leave the remaining tasks orphaned. We need to avoid this. If a user catches that initial exception and keeps their application running, those abandoned background tasks will stick around, and their unhandled exceptions will not be properly garbage collected.

gcsfs/zonal_file.py
googlyrahman: They actually serve two distinct purposes.

Think of an MRD as a single connection stream to Google Cloud Storage. The pool_size parameter decides how many of these streams you want to open, while concurrency dictates how many requests your application is trying to make at the same time.

For example, if you set concurrency=4 and pool_size=1, your application will send all four concurrent requests through a single MRD. The MRD is capable of handling multiple requests simultaneously, but funneling everything through one stream can create a bottleneck and slow things down.

On the other hand, if you set concurrency=4 and pool_size=4, those four requests will be distributed evenly across four separate MRD connections in a round-robin fashion.

A helpful way to look at it is to treat pool_size as your underlying resource limit (similar to the worker count in a ThreadPoolExecutor) and concurrency as the volume of tasks your application is trying to execute at once.

gcsfs/zb_hns_utils.py
googlyrahman: This might introduce race condition. If an await yields execution while a close operation is in progress, this task could resume in between without knowing the state changed. It would then blindly create a new MRD and add it to the queue, even though the MRD pool is already closed.

gcsfs/zb_hns_utils.py
googlyrahman: The active count tracks created MRDs, while the queue shows the free MRDs waiting. Because of this, there is no direct tie between them. Let me know if I'm off base here.

gcsfs/zb_hns_utils.py
googlyrahman: Hmmm, ideally shouldn't happen - but would put a check for `active_count==0`

gcsfs/zb_hns_utils.py
googlyrahman: This suggestion might actually introduces a race condition. If multiple tasks call `initialize()` concurrently, the first task will yield control during the `await self._create_mrd()` step. If the check is outside the lock, a second task could jump in during that await, see that `_initialized` is still False, and erroneously spin up a duplicate connection.

googlyrahman:
> I had a closer look at MRDpool and have a few more comments.
Am I right in thinking that this pool concept should/will be upstreamed to the grpc layer?

The pool concept would be upstreamed to python SDK in future, and not the grpc layer.

googlyrahman:
@martindurant, Addressed your comments - please take a look again

gcsfs/zb_hns_utils.py
martindurant: This little loop really seems like something that asyncio ought to do for us. So be it.

gcsfs/zonal_file.py
martindurant: Sounds like something that would be helped by a well-crafted diagram :)

martindurant (approved):

gcsfs/zb_hns_utils.py
Yonghui-Lee: From Python 3.11, we can use asyncio.TaskGroup to ensure that no dangling tasks are left.

googlyrahman: Yeah i know that, we'd [discussed](https://github.com/fsspec/gcsfs/pull/795#discussion_r3031303400) already in the past. Python 3.10 end of life is scheduled later this year, until then we can't drop support for python 3.10

gcsfs/zonal_file.py
googlyrahman: Took a note of it, and created an internal ticket https://b.corp.google.com/issues/508054742 - will ensure the diagram is added in followup PR.

googlyrahman:
The e2e tests are failing for this issue - https://github.com/fsspec/gcsfs/issues/830

@martindurant please check the existing comments, if you feel everything is addressed. We can merge this.

gcsfs/extended_gcsfs.py
martindurant: Let's keep an issue open about this, we may want to revisit.

martindurant (approved):

gcsfs/extended_gcsfs.py
googlyrahman: Yeah, created a ticket - https://b.corp.google.com/issues/508289429

googlyrahman:
e2e tests passes now as https://github.com/fsspec/gcsfs/issues/830 is fixed , we can merge this.

PR #807, opened 2026-04-14
Fix 'Task exception was never retrieved' in parallel_tasks_first_completed

yuxin00j:
This PR addresses an issue where `parallel_tasks_first_completed` could leave exceptions unretrieved for tasks that were cancelled or failed after the first task completed, leading to "Task exception was never retrieved" warnings.

Instead of awaiting all tasks in the `finally` block (which would defeat the "first completed" optimization by blocking on cancelled tasks), this PR adds a `done` callback to all tasks to consume any exceptions they might raise. This ensures that exceptions are properly acknowledged by the event loop without adding latency to the execution.

### DIFF
--- a/gcsfs/concurrency.py
+++ b/gcsfs/concurrency.py
@@ -10,6 +10,17 @@ async def parallel_tasks_first_completed(coros):
     when exiting the context.
     """
     tasks = [asyncio.create_task(c) for c in coros]
+
+    def _silence_exception(t):
+        try:
+            if not t.cancelled():
+                t.exception()
+        except Exception:
+            pass
+
+    for t in tasks:
+        t.add_done_callback(_silence_exception)
+
     try:
         # Suspend until the first task finishes for maximum responsiveness
         done, pending = await asyncio.wait(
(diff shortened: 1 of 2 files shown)


googlyrahman:
I'll take a look on this later today

Just wanted to highlight:

> Instead of awaiting all tasks in the finally block (which would defeat the "first completed" optimization by blocking on cancelled tasks)

When you call `await asyncio.gather` after `t.cancel()`, it doesn't wait for API request to be completed. It's just process the cancellation. It's standard practice in the industry to await on task once you cancel them, because just doing `t.cancel()` do not guarantee the cancellation, it just schedule the cancellation. You need to explicitly put an await statement to ensure that the task is successfully cancelled.  See https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel

googlyrahman:
LGTM, I probably won't have merge access on this repo, you can ask @martindurant or @ankitaluthra1 for further review and merge process.

googlyrahman (approved):

ankitaluthra1 (approved):

PR #808, opened 2026-04-14
Update the benchmark config, and fix the block size propogation

googlyrahman:
- Update and simplify the read configuration. This ensures we accurately capture the performance gains of prefetching. Small files finish reading before the prefetcher has a chance to provide any real benefit.

- Fix the block size propagation issue. A previous [PR](https://github.com/fsspec/gcsfs/pull/779) separated the block size and chunk size parameters but missed updating the test_read file. Because of that, the block size argument was being ignored entirely. This ensures the parameter is passed down correctly.

### DIFF
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -6,6 +6,7 @@ substitutions:
   _BENCHMARK_CONFIG: ""
   # Format: "hns regional zonal"
   _BUCKET_TYPES: ""
+  _ENABLE_EXPERIMENTAL_READS: "false"
 
 steps:
   # 1. Generate a persistent SSH key for this build run.
@@ -104,6 +105,10 @@ steps:
     id: "run-benchmarks"
     allowFailure: true
     entrypoint: "bash"
+    env:
+      - "BUCKET_TYPES=${_BUCKET_TYPES}"
+      - "BENCHMARK_CONFIG=${_BENCHMARK_CONFIG}"
+      - "ENABLE_EXPERIMENTAL_READS=${_ENABLE_EXPERIMENTAL_READS}"
     args:
       - "-c"
       - |
@@ -204,7 +209,7 @@ steps:
           echo "[$$VM_NAME]: Launching benchmark run for $$group:$$config"
 
           CUSTOM_ENV=""
-          if [[ "$$group" == *"read"* ]] && [[ " ${_BUCKET_TYPES} " =~ " regional " ]]; then
+          if [[ "$$group" == *"read"* ]] && [[ " $${BUCKET_TYPES} " =~ " regional " ]] && [[ "$${ENABLE_EXPERIMENTAL_READS}" == "true" ]]; then
             CUSTOM_ENV="export USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING='true'; export DEFAULT_GCSFS_CONCURRENCY=4"
           fi
 
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -203,6 +203,11 @@ steps:
           IFS=':' read -r group config <<< "$$entry"
           echo "[$$VM_NAME]: Launching benchmark run for $$group:$$config"
 
+          CUSTOM_ENV=""
+          if [[ "$$group" == *"read"* ]] && [[ " ${_BUCKET_TYPES} " =~ " regional " ]]; then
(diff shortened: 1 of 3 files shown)


gcsfs/tests/perf/microbenchmarks/read/configs.yaml
```diff
@@ -3,8 +3,9 @@ common:
     - "regional"
     - "zonal"
   file_sizes_mb:
-    - 1024 # 1 GB
+    - 20480 # 20 GB
   chunk_sizes_mb:
+    - 0.0625
```
zhixiangli, line 8: For my own learning, why don't we target a larger chunk & block size, such as 100MB?

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
```diff
@@ -203,6 +203,11 @@ steps:
           IFS=':' read -r group config <<< "$$entry"
           echo "[$$VM_NAME]: Launching benchmark run for $$group:$$config"
 
+          CUSTOM_ENV=""
+          if [[ "$$group" == *"read"* ]] && [[ " ${_BUCKET_TYPES} " =~ " regional " ]]; then
+            CUSTOM_ENV="export USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING='true'; export DEFAULT_GCSFS_CONCURRENCY=4"
```
zhixiangli, line 213: We should document this to guide users on how to use it to achieve the best performance.

googlyrahman: It's already documented here https://github.com/ankitaluthra1/gcsfs/blob/regional/docs/source/prefetcher.rst

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
googlyrahman: Ideally we should, and I always use 100MB files for my own benchmarking (see [this](https://github.com/ankitaluthra1/gcsfs/blob/regional/docs/source/prefetcher.rst#single-stream-performance-1-process)). However, the existing file only included 1MB, 16MB, and 0.06MB sizes, so I didn't want to introduce a new change. But if others also see the value in a 100MB benchmark, I think it's a good time to add it.

googlyrahman:
@ankitaluthra1 can you review this?

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
jasha26: @zhixiangli i guess after you introduce the benchmark type we will not need * "read" * check here

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
```diff
@@ -3,28 +3,22 @@ common:
...
 scenarios:
   - name: "read_seq_fixed_duration"
     pattern: "seq"
-    files: 30
-
-  - name: "read_seq_fixed_duration_multi_process"
-    pattern: "seq"
-    processes: [16, 48]
+    processes: [1, 16, 48]
+    files: 1
```
jasha26: we don't need to mention this as it's default value

jasha26 (approved):

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
```diff
@@ -3,28 +3,21 @@ common:
...
     pattern: "rand"
-    block_sizes_mb: [0.0625, 16]
-    files: 30
-
-  - name: "read_rand_fixed_duration_multi_process"
-    pattern: "rand"
-    processes: [16, 48]
-    block_sizes_mb: [0.0625, 16]
+    processes: [1, 16, 48]
+    files: 1
```
jasha26, line 24: please remove this as it's default value

jasha26 (approved):

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
googlyrahman: If this is dropped, we would end up creating a 20GB file for each of the $X$ processes per scenario. For 48 processes, this totals 960GB per scenario. While the storage volume might seem manageable, it pushed our benchmark timing to more than 4 hours, resulting in the timeout seen [here](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/5fb82b4f-2c14-45f3-8f56-98aca98cea28?e=13803378&mods=monitoring_api_prod&project=gcs-aiml-clients-testing-101).

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
googlyrahman: Another thread already active for this discussion

PR #811, opened 2026-04-15
Update ptable to prettytable in microbenchmarks

zhixiangli:
## Summary
Update `ptable` to `prettytable` and use markdown table style in microbenchmarks.
This change makes it easier to read and render the benchmark results when sharing the table.
The dependencies in `requirements.txt` were also sorted alphabetically.

## Example
```
| Bucket Type | Group | Pattern | Files | Folders | Threads | Processes | Depth | Target Type | File Size (MiB) | Chunk Size (MiB) | Block Size (MiB) | Mean Latency (s) | Mean Throughput (MiB/s) | Max CPU (%) | Max Memory (MiB) |
| :---------: | :---: | :-----: | :---: | :-----: | :-----: | :-------: | :---: | :---------: | :-------------: | :--------------: | :--------------: | :--------------: | :---------------------: | :---------: | :--------------: |
|   regional  |  info |   info  |   1   |    1    |    1    |     1     |   0   |    bucket   |       N/A       |       N/A        |       N/A        |      0.2237      |           N/A           |     0.00    |      124.73      |
```

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/run.py
+++ b/gcsfs/tests/perf/microbenchmarks/run.py
@@ -8,7 +8,7 @@ import sys
 from datetime import datetime
 
 import numpy as np
-from prettytable import PrettyTable
+from prettytable import PrettyTable, TableStyle
 
 from gcsfs.tests.perf.microbenchmarks.conftest import MB
 
@@ -304,6 +304,7 @@ def _print_csv_to_shell(report_path):
             "Max Memory (MiB)",
         ]
         table = PrettyTable()
+        table.set_style(TableStyle.MARKDOWN)
         table.field_names = display_headers
 
         for row in rows:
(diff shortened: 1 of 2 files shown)


gcsfs/tests/perf/microbenchmarks/run.py
```diff
@@ -304,6 +304,7 @@ def _print_csv_to_shell(report_path):
             "Max Memory (MiB)",
         ]
         table = PrettyTable()
+        table.set_style(TableStyle.MARKDOWN)
```
jasha26, line 307: Can you add to description how the final table looks on console after this change?

jasha26 (approved):

gcsfs/tests/perf/microbenchmarks/run.py
zhixiangli: Added to PR description

PR #813, opened 2026-04-15
benchmarks: implement cleanup-on-failure in Cloud Build

zhixiangli:
This PR ensures that cleanup steps (deleting VMs and buckets) are always executed even if earlier benchmark steps fail.

### Key Changes:
- Added a mechanism to track failed steps using a file in `/workspace/FAILED`.
- Marked setup steps (`create-buckets`, `create-vm`) with `allowFailure: true` to prevent early build termination.
- Introduced `_SKIP_IF_FAILED` to conditionally skip steps if a failure was previously recorded.
- Updated cleanup steps to wait for relevant predecessors and ignore non-critical errors.
- Final `check-failure` step now reports all steps that failed during the build execution.

### DIFF
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -10,6 +10,7 @@ substitutions:
   _BENCHMARK_TYPE: ""
   _IO_VM_TYPE: "c4-standard-192"
   _METADATA_VM_TYPE: "c4-standard-8"
+  _SKIP_IF_FAILED: '[[ -f /workspace/FAILED ]] && echo "Skipping: previous step failed" && exit 0'
 
 steps:
   # 1. Generate a persistent SSH key for this build run.
@@ -51,6 +52,8 @@ steps:
       - "-c"
       - |
         set -e
+        trap 'echo "create-buckets" >> /workspace/FAILED' ERR
+        ${_SKIP_IF_FAILED}
         source /workspace/build_vars.env
 
         # Create Test Buckets, in parallel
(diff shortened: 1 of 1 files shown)


cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
```diff
@@ -15,10 +18,13 @@ steps:
     args:
       - "-c"
       - |
+        set -e
+        trap 'echo "generate-ssh-key" >> /workspace/FAILED' ERR
         mkdir -p /workspace/.ssh
         ssh-keygen -t rsa -f /workspace/.ssh/google_compute_engine -N '' -C gcb
         cat /workspace/.ssh/google_compute_engine.pub > /workspace/gcb_ssh_key.pub
     waitFor: ["-"]
+    allowFailure: true
```
jasha26: if this step fails let's fail everything

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
```diff
@@ -38,6 +46,7 @@ steps:
         echo "export RESULTS_BUCKET=${_INFRA_PREFIX}-run-results" >> /workspace/build_vars.env
         echo "export VM_NAME='${_INFRA_PREFIX}-vm-$${SHORT_BUILD_ID}'" >> /workspace/build_vars.env
     waitFor: ["-"]
+    allowFailure: true
```
jasha26: same as above

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
jasha26: Since the cleanup-ssh-key step performs a project-level OS Login operation that doesn't require the VM to be active, we can trigger the VM deletion immediately after the benchmarks complete. This allows the VM to start its teardown process in parallel with the SSH key cleanup.

jasha26 (approved):

PR #815, opened 2026-04-16
docs: add prefetcher to index to fix ci

zhixiangli:
This PR adds the 'prefetcher' section to the documentation index to ensure it is properly linked in the table of contents and visible to users.

It's to fix the broken ci - docs/readthedocs.org:gcsfs

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -241,7 +241,7 @@ Contents
    rapid_storage_support
    retries
    fuse
-   prefetcher
+   Read Optimization: Prefetching <prefetcher>
    changelog
    code-of-conduct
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -241,6 +241,7 @@ Contents
    rapid_storage_support
    retries
    fuse
+   prefetcher
    changelog
```
ankitaluthra1, line 245: should we add something like `read_prefetch_optimization` to give more context to the reader ?

zhixiangli: OK. Updated to 'Read Optimization: Prefetching'

PR #816, opened 2026-04-17
Changes log level to warning instead of error storage layout API

ankitaluthra1:
- Changes log level to warning instead of error in case bucket type is not detected as code successfully falls back to standard buckets instead of exiting. 
- Error level adds confusion for the user that a fatal exception has occurred in the code execution.
-  Makes the log more verbose

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -142,7 +142,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         except api_exceptions.NotFound:
             logger.warning(
                 f"Error: Bucket {bucket} not found or you lack permissions for "
-                f"storage layout api. Falling back to GCSFileSystem."
+                f"storage layout api used to detect bucket type. Falling back to GCSFileSystem."
             )
             return BucketType.UNKNOWN
         except Exception as e:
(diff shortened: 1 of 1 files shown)


gcsfs/extended_gcsfs.py
```diff
@@ -140,11 +140,11 @@ async def _get_bucket_type(self, bucket):
...
             return BucketType.NON_HIERARCHICAL
         except api_exceptions.NotFound:
-            logger.warning(f"Error: Bucket {bucket} not found or you lack permissions.")
+            logger.warning(f"Error: Bucket {bucket} not found or you lack permissions for storage layout api. Falling back to standard buckets")
             return BucketType.UNKNOWN
         except Exception as e:
-            logger.error(
-                f"Could not determine bucket type for bucket name {bucket}: {e}"
+            logger.warning(
+                f"Falling back to GCS Standard bucket. Could not determine bucket type for bucket name {bucket}: {e}"
```
Mahalaxmibejugam: Should we say "Falling back to GCSFileSystem"? Also we can add error message in the beginning and then add about fallback?

Something like: "Could not determine bucket type for bucket name {bucket}: {e}, falling back to GCSFileSystem"

Mahalaxmibejugam (approved):

gcsfs/extended_gcsfs.py
ankitaluthra1: updated

martindurant:
Thanks for fixing, @ankitaluthra1  - now I understand what the actual problem was :)

Issue #814, opened 2026-04-16
ExtendedGcsFileSystem not working with requester pays buckets


wdobbels:
Since switching to gcsfs 2026.3.0, we've noticed we get this error when trying to access a bucket:

```
Could not determine bucket type for bucket name my-bucket-name:
400 Bucket is a requester pays bucket but no user project provided.
```

The code to reproduce is just:

```
import gcsfs

fs = gcsfs.GCSFileSystem(requester_pays=True)

with fs.open("gs://my-bucket/my_file.json", "r") as f:
    print(f.read(10))
```

Previously, with `GcsFileSystem`, this did not produce any errors, these only happen with `ExtendedGcsFileSystem`.

martindurant:
For the moment, you can explicitly use `gcsfs.core.GCSFileSystem` or pass environment variable GCSFS_EXPERIMENTAL_ZB_HNS_SUPPORT=False. We will fix this.

wdobbels:
Thank you for the quick reply. I will use the environment variable for now, until the fix is live.

ankitaluthra1:
When we initialize the file system with requester_pays=True, the base GCSFileSystem correctly includes the billing project in its HTTP requests by appending the userProject query parameter. However, ExtendedGcsFileSystem introduces new gRPC calls. These gRPC requests via StorageControlAsyncClient do not automatically receive the user_project context.

As a result, server rejects the get_storage_layout request with a 400 error which is the first call done in ExtendedFileSystem to identify the bucket type. 

Suggested Fix
To fix this, we need to extract the user project and pass it as the x-goog-user-project header within the gRPC metadata for all Storage Control API calls. We ll add the fix soon for this edge case, thanks @wdobbels for bringing it up !!

ankitaluthra1:
@wdobbels Can you confirm if the code flow exited with exception or you are able to access bucket but seeing this recurring error log. We have a fallback to GCSFileSystem after logging this error. Right after logging the error, the code should ideally recover by returning the unknown type which uses GCSFileSystem

https://github.com/fsspec/gcsfs/blob/ad8914cb8ea413bb31733e1d832bc44c32d3d3ac/gcsfs/extended_gcsfs.py#L150

Mahalaxmibejugam:
[...]

PR #817, opened 2026-04-18
Support for Requester Pays buckets in ExtendedGcsFileSystem

Mahalaxmibejugam:
Summary

This PR adds support for Requester Pays buckets in ExtendedGcsFileSystem, specifically for Hierarchical Namespace (HNS) operations that utilize the Storage Control gRPC API. 

The primary fix addresses an issue where billing/quota headers were missing on gRPC requests to the Control Plane because the quota_project_id was not being propagated to the manually created gRPC channel.

  Changes

  Core Logic
   - gcsfs/extended_gcsfs.py: 
       - Added a user_project helper property to resolve the billing project from the requester_pays and project configurations.
       - Updated _get_control_plane_client to explicitly pass quota_project_id to transport_cls.create_channel. This ensures that the x-goog-user-project metadata is correctly attached to the gRPC channel used by the StorageControlAsyncClient.

  Testing & Infrastructure
   - Integration Tests: Added TestExtendedGcsFileSystemHnsRequesterPays in gcsfs/tests/integration/test_extended_hns.py to verify that HNS operations (like mkdir and bucket type detection) succeed on requester-pays buckets when properly configured and fail with an informative message when they are not.
   - Fixtures: Updated gcsfs/tests/conftest.py to include fixtures for creating and managing requester-pays HNS buckets during integration tests.
   - CI/CD: Updated cloudbuild scripts to support the automated creation and teardown of requester-pays resources for testing.

Note on AsyncGrpcClient: 

During development, a similar issue was identified in the underlying [AsyncGrpcClient](https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-storage/google/cloud/storage/asyncio/async_grpc_client.py#L99). Currently, AsyncGrpcClient also ignores quota_project_id and api_endpoint from ClientOptions during manual channel creation. While the fix for the Storage Control client is included in this PR, a separate PR will be needed for the python-storage library to fix the data plane gRPC client.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -115,7 +115,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
             channel = transport_cls.create_channel(
                 credentials=self.credential,
                 options=[("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
-                quota_project_id=self.user_project,
+                # quota_project_id=self.user_project,
             )
             transport = transport_cls(channel=channel)
             self._storage_control_client = storage_control_v2.StorageControlAsyncClient(
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -115,7 +115,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
             channel = transport_cls.create_channel(
                 credentials=self.credential,
                 options=[("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
-                # quota_project_id=self.user_project,
+                quota_project_id=self.user_project,
             )
             transport = transport_cls(channel=channel)
             self._storage_control_client = storage_control_v2.StorageControlAsyncClient(
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -70,6 +70,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
 
     @property
     def _user_project(self):
+        """Value used for billing - enabling "requestor pays" access"""
         if self.requester_pays:
             return (
                 self.requester_pays
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -69,6 +69,16 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         self._storage_layout_cache = {}
 
     @property
+    def user_project(self):
+        if self.requester_pays:
+            return (
+                self.requester_pays
+                if isinstance(self.requester_pays, str)
+                else self.project
+            )
+        return None
+
+    @property
     def grpc_client(self):
         if self.asynchronous and self._grpc_client is None:
             raise RuntimeError(
@@ -105,6 +115,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
             channel = transport_cls.create_channel(
                 credentials=self.credential,
                 options=[("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
+                quota_project_id=self.user_project,
             )
             transport = transport_cls(channel=channel)
             self._storage_control_client = storage_control_v2.StorageControlAsyncClient(
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -69,7 +69,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         self._storage_layout_cache = {}
 
     @property
-    def user_project(self):
+    def _user_project(self):
         if self.requester_pays:
             return (
                 self.requester_pays
@@ -115,7 +115,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
             channel = transport_cls.create_channel(
                 credentials=self.credential,
                 options=[("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
-                quota_project_id=self.user_project,
+                quota_project_id=self._user_project,
             )
             transport = transport_cls(channel=channel)
             self._storage_control_client = storage_control_v2.StorageControlAsyncClient(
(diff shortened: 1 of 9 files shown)


gcsfs/tests/settings.py
```diff
@@ -5,7 +5,12 @@
 TEST_HNS_BUCKET = os.getenv("GCSFS_HNS_TEST_BUCKET", "gcsfs_hns_test")
 TEST_ZONAL_BUCKET = os.getenv("GCSFS_ZONAL_TEST_BUCKET", "gcsfs_zonal_test")
 TEST_PROJECT = os.getenv("GCSFS_TEST_PROJECT", "project")
-TEST_REQUESTER_PAYS_BUCKET = f"{TEST_BUCKET}_req_pay"
+TEST_REQUESTER_PAYS_BUCKET = os.getenv(
+    "GCSFS_TEST_REQ_PAYS_BUCKET", "gcsfs_test_req_pay"
+)
+TEST_HNS_REQUESTER_PAYS_BUCKET = os.getenv(
+    "GCSFS_HNS_TEST_REQ_PAYS_BUCKET", "gcsfs_hns_test_req_pay"
```
zhixiangli: Can we just set the default as None? gcsfs_hns_test_req_pay and gcsfs_test_req_pay are not used anywhere else.

gcsfs/extended_gcsfs.py
```diff
@@ -68,6 +68,16 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
             self.credential = AnonymousCredentials()
         self._storage_layout_cache = {}
 
+    @property
+    def user_project(self):
```
zhixiangli: `_user_project`, the user_project isn't supposed to be exposed yet, I believe?

gcsfs/tests/test_core.py
```diff
@@ -1538,18 +1538,34 @@ def test_user_project_fallback_google_default(monkeypatch):
...
+    file_path = f"{requester_pays_bucket}/test_file_project.txt"
+    data = b"test data requester pays with project"
+
+    try:
+        gcs.pipe(file_path, data)
+        assert gcs.cat(file_path) == data
+    finally:
+        try:
+            gcs.rm(file_path)
+        except Exception:
```
zhixiangli: Could you explain why the exception is being passed here? If it's expected, we should assert the exception; if it's unexpected, we should fix the code to ensure it works correctly.

zhixiangli (changes requested):

gcsfs/tests/settings.py
Mahalaxmibejugam: The defaults are set so that if the user does not provide the environment variable, the default value will be used when creating the bucket, and the bucket will be deleted at the end of the test session.

Same applies for all the variables in this file.

gcsfs/tests/test_core.py
Mahalaxmibejugam: This cleanup block is focused on emptying the bucket and is not related to the scenario being tested, so assertions are not needed. It serves only as a safeguard to ensure that a cleanup failure is not reported as a test failure.

I have observed that this cleanup block is being repeated, so I have moved it to fixture.

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: Done

gcsfs/tests/settings.py
zhixiangli: Actually, the tests are skipped when these variables use the default strings, so no bucket should be created. Setting the default to None would be cleaner and remove the need for magic strings. WDYT?

zhixiangli (approved):

gcsfs/tests/settings.py
Mahalaxmibejugam: Skips are initially added to make sure these tests don't run on the Zonal bucket, but these are not needed in the fixture, I have moved the necessary skips to cloudbuild.

jasha26 (approved):

Mahalaxmibejugam:
@martindurant Can you please review and merge this PR, this fixes https://github.com/fsspec/gcsfs/issues/814. Also when can we plan a release with this fix?

gcsfs/extended_gcsfs.py
```diff
@@ -68,6 +68,16 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
             self.credential = AnonymousCredentials()
         self._storage_layout_cache = {}
 
+    @property
+    def _user_project(self):
+        if self.requester_pays:
```
martindurant, line 74:
```suggestion
        """Value used for billing - enabling "requestor pays" access"""
        if self.requester_pays:
```

gcsfs/tests/conftest.py
```diff
@@ -229,7 +273,12 @@ def final_cleanup(gcs_factory, buckets_to_delete):
     yield
     # This code runs after the entire test session finishes
 
-    gcs = gcs_factory()
+    # Check if we have any requester pays buckets in the deletion list
+    use_requester_pays = any(
+        bucket in [TEST_REQUESTER_PAYS_BUCKET, TEST_HNS_REQUESTER_PAYS_BUCKET]
+        for bucket in buckets_to_delete
+    )
```
martindurant: A set intersection? 
Is there any reason not to just say True here? I ask because it took me a moment to figure out what this did.

martindurant (approved):
The tests are very permutative, perhaps pytest.parametrize or something would have been cleaner - but I'm OK going with this. 
One comment.

gcsfs/tests/conftest.py
Mahalaxmibejugam: This was needed to delete the buckets which are enabled with requester_pays. I was setting it conditionally only when there is any bucket of that category to be deleted but I think there would no issue in always setting it to True.

Mahalaxmibejugam:
>  The tests are very permutative, perhaps pytest.parametrize or something would have been cleaner

Parameterized tests to make it more readable

PR #818, opened 2026-04-20
Refactor prefetcher code

googlyrahman:
Summary of the changes:

#### Refactor Prefetcher file: 

- Replaced the passing of individual lambda functions to the `PrefetchProducer` and `PrefetchConsumer` with direct object references. This streamlines the initialization process and makes the internal dependency graph much cleaner.

#### Start prefetching on 3rd read, instead of second

- Previously, prefetching started on the second read. However, when used in conjunction with caches like `readahead_chunked` (which generates two underlying requests for a single user read), this triggered false positive signals, causing the engine to prefetch unnecessary extra data. Delaying the start until the third read corrects this behavior.

#### Disable Prefetching for variable reads with average greater than 64MB

When read chunk sizes are highly variable, the producer fetches data based on a rolling average, forcing the consumer to slice and stitch multiple partial chunks to match the exact user request.
- For average chunk sizes <= 64MB, this memory assembly is fast, the bottleneck remains network I/O, and prefetching is highly beneficial.
- For average chunk sizes > 64MB, memory allocation and byte concatenation shift the bottleneck to the CPU, which slows down the entire operation.

<details>
<summary>Numbers before Change</summary>

### Before
| SCENARIO SUMMARY | CONCURRENCY (GB/s) | PREFETCHER+CONCURRENCY (GB/s) | DIFF |
|---|---|---|---|
| Sequential fixed 64KB | 0.03 | 0.37 | +1065.2% |
| Sequential fixed 1MB | 0.17 | 1.13 | +558.7% |
| Sequential fixed 16MB | 1.08 | 2.04 | +89.7% |
| Sequential fixed 150MB | 2.09 | 2.32 | +11.2% |
| Sequential fixed 256MB | 2.32 | 2.30 | -1.0% |
| Sequential fixed 1GB | 2.43 | 2.41 | -0.7% |
| Sequential fixed 4MB (Below MIN_CHUNK) | 0.36 | 1.44 | +293.5% |
| Sequential fixed 6MB (Above MIN_CHUNK) | 0.66 | 1.59 | +140.0% |
| Sparse Read: Read 1MB, Skip 10MB forward | 0.20 | 0.19 | -3.6% |
| Sparse Read: Read 16MB, Skip 100MB forward | 1.22 | 1.20 | -1.5% |
| Sequential variable (1KB - 1MB) [Micro Jitter] | 0.14 | 1.18 | +717.4% |
| Sequential variable (1MB - 32MB) [Medium Jitter] | 1.07 | 1.94 | +80.7% |
| Sequential variable (64KB - 100MB) [Good Jitter] | 1.74 | 1.63 | -6.6% |
| Sequential variable (64KB - 1GB) [Large Jitter] | 2.42 | 1.16 | -51.9% |
| Sequential variable (54MB - 62MB) [Jitter Below 64MB] | 1.67 | 1.41 | -15.9% |
| Sequential variable (66MB - 74MB) [Jitter Above 64MB] | 1.73 | 1.34 | -22.7% |
| Sequential variable (90MB - 98MB) [Deep into CPU limits] | 1.98 | 1.56 | -21.2% |
| Step Up/Down (10MB to 100MB to 10MB) | 1.25 | 1.74 | +39.5% |
| Step Up/Down (10MB to 150MB to 120MB) | 1.57 | 1.56 | -0.4% |
| Step Up/Down Extreme Drop (1MB to 500MB to 1MB) | 0.77 | 1.33 | +73.1% |
| Pure Seek fixed (1MB) | 0.17 | 0.19 | +11.6% |
| Pure Seek fixed (16MB) | 1.12 | 1.07 | -5.0% |
| Pure Seek variable (64KB - 100MB) | 1.71 | 1.79 | +4.7% |
| 30% Seq / 70% Seek (fixed 16MB) | 1.09 | 0.76 | -30.6% |
| 50% Seq / 50% Seek (fixed 16MB) | 1.12 | 0.70 | -37.2% |
| 70% Seq / 30% Seek (fixed 16MB) | 1.14 | 0.77 | -33.0% |
| 90% Seq / 10% Seek (fixed 16MB) | 1.06 | 1.09 | +2.8% |
| 50% Seq/Seek variable (64KB - 100MB) | 1.58 | 0.96 | -39.2% |

</details>

<details>
<summary>Numbers After Change</summary>

### After
| SCENARIO SUMMARY | CONCURRENCY (GB/s) | PREFETCHER+CONCURRENCY (GB/s) | DIFF |
|---|---|---|---|
| Sequential fixed 64KB | 0.03 | 0.37 | +1002.9% |
| Sequential fixed 1MB | 0.20 | 1.28 | +524.2% |
| Sequential fixed 16MB | 1.06 | 2.15 | +102.9% |
| Sequential fixed 150MB | 2.09 | 2.11 | +1.3% |
| Sequential fixed 256MB | 2.22 | 2.31 | +3.9% |
| Sequential fixed 1GB | 2.41 | 2.41 | +0.4% |
| Sequential fixed 4MB (Below MIN_CHUNK) | 0.40 | 1.20 | +197.9% |
| Sequential fixed 6MB (Above MIN_CHUNK) | 0.56 | 1.46 | +160.4% |
| Sparse Read: Read 1MB, Skip 10MB forward | 0.21 | 0.15 | -26.2% |
| Sparse Read: Read 16MB, Skip 100MB forward | 1.14 | 1.17 | +3.3% |
| Sequential variable (1KB - 1MB) [Micro Jitter] | 0.14 | 1.17 | +727.9% |
| Sequential variable (1MB - 32MB) [Medium Jitter] | 1.13 | 1.91 | +69.8% |
| Sequential variable (64KB - 100MB) [Good Jitter] | 1.71 | 1.57 | -8.1% |
| Sequential variable (64KB - 1GB) [Large Jitter] | 2.32 | 2.38 | +2.6% |
| Sequential variable (54MB - 62MB) [Jitter Below 64MB] | 1.70 | 1.49 | -11.8% |
| Sequential variable (66MB - 74MB) [Jitter Above 64MB] | 1.69 | 1.85 | +9.1% |
| Sequential variable (90MB - 98MB) [Deep into CPU limits] | 1.91 | 1.83 | -3.9% |
| Step Up/Down (10MB to 100MB to 10MB) | 1.25 | 1.61 | +29.2% |
| Step Up/Down (10MB to 150MB to 120MB) | 1.70 | 1.37 | -19.4% |
| Step Up/Down Extreme Drop (1MB to 500MB to 1MB) | 0.90 | 1.69 | +87.0% |
| Pure Seek fixed (1MB) | 0.19 | 0.21 | +8.5% |
| Pure Seek fixed (16MB) | 1.05 | 1.04 | -0.4% |
| Pure Seek variable (64KB - 100MB) | 1.55 | 1.71 | +10.4% |
| 30% Seq / 70% Seek (fixed 16MB) | 1.09 | 1.02 | -6.4% |
| 50% Seq / 50% Seek (fixed 16MB) | 1.08 | 0.94 | -13.4% |
| 70% Seq / 30% Seek (fixed 16MB) | 1.09 | 0.87 | -19.7% |
| 90% Seq / 10% Seek (fixed 16MB) | 1.12 | 1.09 | -2.4% |
| 50% Seq/Seek variable (64KB - 100MB) | 1.68 | 1.44 | -14.2% |

</details>


martindurant:
> when used in conjunction with caches like readahead_chunked

This seems like unnecessary coupling between the behaviours of the "cache" and "fetcher",  just the kind of thing you were wishing to avoid. This way around, if the cache is None (which is recommended), you may end up waiting, no?

googlyrahman:
> This seems like unnecessary coupling between the behaviours of the "cache" and "fetcher", just the kind of thing you were wishing to avoid.

Yes Ideally, one shouldn't layer the existing caches on top of the `prefetcher`. However, as we discussed, customers might still enable this experimental flag to test it with default cache (currently `readahead_chunked` for zonal) during a trial. This change should support that use case. 

> if the cache is None (which is recommended), you may end up waiting, no?

This is correct, if `cache_type = none` - we do not start prefetching on the first sequential read, we start prefetching on second sequential read.

---

While I agree that starting prefetching on the second read is generally more logical, we have to consider existing users who might enable this experimental flag while still using current caches. We haven't transitioned the default cache setting to none yet because removing buffering entirely could negatively impact performance for large portion of our user base who do not use this experimental flag.

zhixiangli:
What happens if read-ahead runs out of cache? I think it will trigger the exact same false-positive signals again?

googlyrahman:
> What happens if read-ahead runs out of cache? I think it will trigger the exact same false-positive signals again?

Yes, so basically in `cache_type="none"` the prefetching would start from actual third read, and in `cache_type="readahead_chunked"` the prefetching would start from actual second second.

**I'm not concerned about handling false-positive here**, I'm only concerned about the prefetching shouldn't happen at the first read, because then it would destroy the throughput of users doing completely random reads.

If the goal is to fix this false positive, then this variable `MIN_STREAKS_FOR_PREFETCHING` needs to be adjusted with respect to cache_types, which would introduce some un-necessary code in `core.py` which i think might not be desirable at the moment, The core.py is anyway very big, and we wish to simplify the code in future + `readahead_chunked` will gonna deprecate in future (while there are other caches with similar behaviour of `readahead_chunked` such as `background` block  but they are rarely used as per my past converation with Martin).

googlyrahman:
@martindurant can you review this?

gcsfs/prefetcher.py
```diff
@@ -101,17 +119,34 @@ class PrefetchProducer:
...
+    # requests random massive sizes (e.g., jumping between 100MB and INF), the
+    # producer still fetches chunks based on the rolling average. The consumer
+    # then has to pick up multiple chunks and stitch them together to match the
+    # exact requested size.
+    #
+    # For small average read sizes, this byte assembly is fast and the bottleneck
+    # remains the network I/O. However, for massive reads (>= 100MB), the extra
+    # step of copying and assembling huge byte strings in memory severely slows
+    # down the operation.
+    VARIABLE_IO_THRESHOLD = 100 * 1024 * 1024
```
zhixiangli: nit: VARIABLE_IO_THRESHOLD_BYTES

gcsfs/prefetcher.py
```diff
@@ -389,7 +443,11 @@ async def _advance(self, size: int, save_data: bool) -> list[bytes]:
             available = len(self._current_block) - self._current_block_idx
 
             if not available:
-                if self.is_producer_stopped() and self.queue.empty():
+                is_producer_stopped = (
+                    not hasattr(self.orchestrator, "producer")
```
zhixiangli: nit: how about setting the producer to None then we do `self.orchestrator.producer is None
` check here? It's more readable.

zhixiangli (approved):

gcsfs/prefetcher.py
googlyrahman: done

googlyrahman:
@martindurant, can you take a look if you're interested? The benchmarks before and after this change is attached in the description.

martindurant:
The change should affect the small reads primarily?

martindurant:
(sorry no, that comment is for #840)

googlyrahman:
>  The change should affect the small reads primarily?
> (sorry no, that comment is for https://github.com/fsspec/gcsfs/pull/840)

Yes, it should impact small reads primarily.

docs/source/prefetcher.rst
```diff
@@ -189,3 +189,18 @@ Multi Stream Performance (48 Process)
...
+
+.. image:: _static/prefetch_dynamic_transition.png
+
+Notice how perfectly the prefetcher mirrors the application's changing behaviour across three distinct phases:
+
+1. **Smooth Sequential Reading (0–10s):** The application starts by reading steady 16MB chunks. After three consistent reads, the prefetcher confirms the pattern and confidently ramps up its background fetching. It maintains a comfortable buffer ahead of the application so the user never waits on the network.
+2. **Pausing for Random Seeks (10–20s):** The application suddenly stops reading in a straight line and starts jumping randomly around the file. Prefetching is actively harmful here. The engine instantly detects the broken streak and drops the background buffer to zero, ensuring no network bandwidth or memory is wasted downloading unneeded data.
+3. **Adjusting to Massive Reads (20–30s):** The application resumes reading sequentially, but this time stepping up to massive 100MB chunks. The algorithm quickly catches on. It detects the new streak, calculates the new rolling average, and rebuilds the background buffer—this time scaling it up safely to handle the much larger data chunks.
```
martindurant: Removing superlatives - I think it's clearer with neutral language.

```suggestion
Notice how the prefetcher mirrors the application's behaviour across three distinct phases:

1. **Sequential reading (0–10s):** The application starts by reading 16MB chunks. After three reads, the prefetcher ramps up its background fetching. It maintains a buffer ahead of the application so the user never waits on the network.
2. **Random seeks (10–20s):** The application starts jumping randomly around the file. Prefetching would be harmful here. The engine detects the broken streak and drops the background buffer to zero, ensuring no network bandwidth or memory is wasted downloading data.
3. **Larger sequential reads (20–30s):** The application resumes reading sequentially, but this time stepping up to 100MB chunks. The algorithm detects the new streak, calculates the new rolling average, and rebuilds the background buffer—this time scaling it up to handle the larger data chunks.
```

gcsfs/tests/test_prefetcher.py
```diff
@@ -136,9 +136,9 @@ def test_producer_concurrency_streak_and_min_chunk():
     original_min_chunk = bp.producer.MIN_CHUNK_SIZE
     bp.producer.MIN_CHUNK_SIZE = 10
 
-    bp._fetch(0, 50)
-    bp._fetch(50, 100)
-    bp._fetch(100, 150)
+    # Do 6 reads to push the streak well past the MIN_STREAKS threshold
+    for i in range(6):
+        bp._fetch(i * 50, (i + 1) * 50)
```
martindurant, line 150: It feels like the number here are somewhat arbitrary - or at least, they depend on the current set of defaults assumed by the prefetcher. Maybe they should be explicitly derived from those values?
I don't mind if not, but do add a comment, because they will need to be updated should the defaults change.

gcsfs/tests/test_prefetcher.py
```diff
@@ -172,17 +172,14 @@ def test_producer_loop_space_constraints():
...
 def test_producer_error_propagation():
-    fetcher = MockFetcher(b"A" * 1000, fail_at_call=3)
-    bp = BackgroundPrefetcher(fetcher=fetcher, size=1000, concurrency=4)
-    bp.read_tracker.add(100)
+    fetcher = MockFetcher(b"A" * 2000, fail_at_call=3)
+    bp = BackgroundPrefetcher(fetcher=fetcher, size=2000, concurrency=4)
 
-    assert bp._fetch(0, 100) == b"A" * 100
+    for i in range(2):
+        bp._fetch(i * 100, (i + 1) * 100)
```
martindurant, line 186: Before, the read_tracker was directly mutated. Why was that insufficient?

gcsfs/tests/test_prefetcher.py
```diff
@@ -533,3 +533,95 @@ async def trigger_stop_and_wake():
...
+    bp = BackgroundPrefetcher(
+        fetcher=fetcher, size=1000, concurrency=4, max_prefetch_size=40
+    )
+
+    # Do enough reads to build a sequential streak and trigger large averages
+    # Reading 60 bytes at a time. Average = 60. Threshold = 50.
+    for i in range(4):
+        bp._fetch(i * 60, (i + 1) * 60)
+
+    fsspec.asyn.sync(bp.loop, asyncio.sleep, 0.1)
```
martindurant, line 545: The sleep may need to be bigger in CI machines - can there be some form of wait here?

gcsfs/prefetcher.py
```diff
@@ -78,6 +78,24 @@ def average(self) -> int:
...
 
+    @property
+    def is_variable(self) -> bool:
+        """Determines if the history contains distinct chunk sizes."""
+        count = len(self._history)
+        if count < 2:
+            return False
+
+        first_val = self._history[0]
+        return any(val != first_val for val in self._history)
```
martindurant:
```suggestion
        return len(set(self._history)) > 1
```
?

Whether this is faster probably depends on where the first non-equal value is and how long the list can be.

gcsfs/prefetcher.py
```diff
@@ -101,17 +119,34 @@ class PrefetchProducer:
     # to maximum of 2 * io_size and 128MB
     MIN_PREFETCH_SIZE = 128 * 1024 * 1024
 
+    # The prefetching starts on the third read.
+    MIN_STREAKS_FOR_PREFETCHING = 3
+
+    # Threshold for disabling proactive prefetching on large, variable reads.
+    #
+    # If the average read size exceeds this value and patterns are variable,
+    # prefetching shifts from an I/O bottleneck to a CPU bottleneck. When a user
```
martindurant:
```suggestion
    # prefetching shifts from an I/O bottleneck to a memcopy (CPU) bottleneck. When a user
```

gcsfs/prefetcher.py
```diff
@@ -101,17 +119,34 @@ class PrefetchProducer:
...
+    # requests random massive sizes (e.g., jumping between 64MB and INF), the
+    # producer still fetches chunks based on the rolling average. The consumer
+    # then has to pick up multiple chunks and stitch them together to match the
+    # exact requested size.
+    #
+    # For small average read sizes, this byte assembly is fast and the bottleneck
+    # remains the network I/O. However, for massive reads (>= 64MB), the extra
+    # step of copying and assembling huge byte strings in memory severely slows
+    # down the operation.
+    VARIABLE_IO_THRESHOLD = 64 * 1024 * 1024
```
martindurant, line 141: Does the best value of this depend on the network bandwidth? I bet on slow connections, we always prefer any amount of prefetching and copy time is irrelevant.

gcsfs/prefetcher.py
```diff
@@ -140,10 +174,9 @@ def __init__(
         self.queue = queue
         self.wakeup_event = wakeup_event
 
-        self.get_user_offset = get_user_offset
-        self.get_io_size = get_io_size
-        self.get_sequential_streak = get_sequential_streak
-        self.on_error = on_error
+        self.consumer = consumer
+        self.tracker = tracker
+        self.orchestrator = orchestrator
```
martindurant: Reference cycle - prefer a weakref?

gcsfs/prefetcher.py
```diff
@@ -246,23 +279,60 @@ async def _loop(self):
...
+                streak = self.consumer.sequential_streak
+                is_variable = self.tracker.is_variable
+                last_read_size = self.tracker.last_value
+
+                exceeds_user_max = (
+                    self._user_max_prefetch_size is not None
+                    and avg_io_size > self._user_max_prefetch_size
+                )
+
+                # Disable prefetching ahead if highly variable AND average > 64MB, or if it exceeds user max
```
martindurant: > if highly variable

If anything other than exactly constant, no?

gcsfs/prefetcher.py
martindurant: (or set it to None as a class attribute)

gcsfs/prefetcher.py
```diff
@@ -401,15 +487,38 @@ async def _advance(self, size: int, save_data: bool) -> list[bytes]:
 
                 if isinstance(task, Exception):
                     logger.error("Consumer retrieved an exception: %s", task)
-                    self.on_error(task)
+                    self.orchestrator._set_error(task)
```
martindurant: _set_error doesn't need the leading _ really - the only calling context is from this class, and it is meant to be called.

gcsfs/tests/test_prefetcher.py
googlyrahman: Made them derive from the main code.

gcsfs/tests/test_prefetcher.py
googlyrahman: I didn't get this, can you detail this comment more?

gcsfs/tests/test_prefetcher.py
googlyrahman: I think, given this is just a test, this should be fine. I've tested it on smaller machine (4 cores) as well as bigger machine (192 cores). This test primarily checks whether prefetching remains disable if average is greater than user specified maximum size. a 0.1s gives event loop enough space to schedule things up if it is required.

gcsfs/prefetcher.py
googlyrahman: I think `any(...)` is better, it would exit early as soon as it finds the first non equal value, additionally it doesn't need to create any extra space for set.

gcsfs/prefetcher.py
googlyrahman: Yes, this is true, for slow connections it's always best to prefetch any amount. I'm just playing safe here by putting a lower value, and setting this value based on the numbers derived from fastest network. Do you think we should adjust this variable based on bucket type?

gcsfs/prefetcher.py
googlyrahman: Made it public.

gcsfs/prefetcher.py
googlyrahman: Yeah, reworded this.

gcsfs/prefetcher.py
googlyrahman: Done.

gcsfs/prefetcher.py
googlyrahman: Done

docs/source/prefetcher.rst
googlyrahman: Done

googlyrahman:
@martindurant, Happy Monday!

Sorry for the delay, I got tied up with a few other things and couldn't follow up on this PR sooner. I've addressed your comments and included one additional small fix. (I kept them in separate commits to make reviewing easier.)

Regarding the small fix: Previously, if the background prefetcher entered an error state, it would permanently reject incoming read requests. This is problematic for users who want to catch the exception and retry on their end, because any subsequent retries would just return the cached exception. The latest commit solves this by triggering a hard seek to the requested offset. The hard seek takes care of clearing the queue, resetting the internal prefetch state, and starting fresh with a new network request.

Let me know what you think!

gcsfs/tests/test_prefetcher.py
```diff
@@ -171,33 +172,50 @@ def test_producer_loop_space_constraints():
...
+    bp = BackgroundPrefetcher(fetcher=MockFetcher(b"X" * 100), size=100, concurrency=4)
+
+    # Simulate an error state from a previous failed read
+    bp._error = ValueError("Pre-existing error")
+
+    # The new error-recovery logic allows a subsequent read to clear the error and succeed
+    assert bp._fetch(0, 10) == b"X" * 10
+    assert bp._error is None
+
+    bp.close()
```
martindurant: Any problem if close() is not reached due to an error in the test? Should this be a fixture?

gcsfs/prefetcher.py
martindurant: My testing suggests `set` is always faster if the history isn't big (which it never is!) *or* the non-equal value is not at the start. The iterator and set of hashes probably take us similar, negligible space.

gcsfs/prefetcher.py
martindurant: _In general_, we should be able to establish how close to the data we are running and make sensible decisions based on it. That would make nice follow-up work.

gcsfs/prefetcher.py
```diff
@@ -246,23 +286,60 @@ async def _loop(self):
...
+                is_variable = self.tracker.is_variable
+                last_read_size = self.tracker.last_value
+
+                exceeds_user_max = (
+                    self._user_max_prefetch_size is not None
+                    and avg_io_size > self._user_max_prefetch_size
+                )
+
+                # Disable prefetching ahead if variable AND average > 64MB, or if it exceeds user max
+                if (
```
martindurant: This method got very long with all these cases, maybe should split the "choices" and the "action" parts.

gcsfs/prefetcher.py
```diff
@@ -307,6 +392,10 @@ async def _loop(self):
                     await self.queue.put(download_task)
                     self.current_offset += actual_size
 
+                if self.current_offset >= self.size:
+                    logger.debug("Producer reached EOF. Exiting background loop.")
+                    self.is_stopped = True
+                    break
         except asyncio.CancelledError:
```
martindurant: This big indent feels like a context?

gcsfs/prefetcher.py
```diff
@@ -307,6 +392,10 @@ async def _loop(self):
...
             logger.debug("PrefetchProducer loop was cancelled.")
             pass
```
martindurant:
```suggestion
```

gcsfs/prefetcher.py
```diff
@@ -401,15 +498,38 @@ async def _advance(self, size: int, save_data: bool) -> list[bytes]:
...
 
                     self.sequential_streak += 1
-                    if self.sequential_streak >= 2:
-                        self.wakeup_event.set()
+                    if (
+                        self.sequential_streak
+                        >= PrefetchProducer.MIN_STREAKS_FOR_PREFETCHING
+                    ):
+                        is_variable = self.tracker.is_variable
+                        avg_io_size = self.tracker.average
```
martindurant: May as well inline these in the calculations below, since they are already multiline. We don't use the variables again.

gcsfs/tests/test_prefetcher.py
googlyrahman: You're right, they should be a fixture. Made them a fixture.

gcsfs/prefetcher.py
googlyrahman: Hmm, that’s interesting! I just ran the test myself, and you're absolutely right - set() is consistently faster for small size deque. I honestly didn't expect it to outperform the short-circuiting logic of any(), but the overhead of the generator clearly outweighs the set() when the history size is small.

Good catch, Martin! I'll update it to `return len(set(self._history)) > 1`.

gcsfs/prefetcher.py
googlyrahman: Done, Extracted the sizing math into _calculate_prefetch_params() (the 'choices') and the actual task enqueuing into _process_prefetch_cycle() (the 'action').

gcsfs/prefetcher.py
googlyrahman: By pulling that logic out, the try/while block in _loop is now completely flattened, which fixes the 'big indent' issue without needing a custom context manager, let me know otherwise.

gcsfs/prefetcher.py
googlyrahman: Removed un-necessary pass.

gcsfs/prefetcher.py
googlyrahman: Made them inline.

gcsfs/prefetcher.py
googlyrahman: Sounds good, Created a issue in github for dynamic setting for `variable_io_threshold`. The issue can be tracked here https://github.com/fsspec/gcsfs/issues/852

googlyrahman:
Addressed your comments @martindurant, please take a look again

martindurant (approved):
Thanks for working with me ;)

PR #819, opened 2026-04-22
Enable branch wise tracking

googlyrahman:
This PR adds branch-specific tracking in cloud builders script for the PLX dashboard. The cloud builder propagates a `{BRANCH_NAME}` environment variable based on user input on UI, which we use now to track benchmarks separately for each branch.

### DIFF
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -26,11 +26,15 @@ steps:
     entrypoint: "bash"
     env:
       - "BUILD_ID=${BUILD_ID}"
+      - "BRANCH_NAME=${BRANCH_NAME}"
     args:
       - "-c"
       - |
         SHORT_BUILD_ID=$${BUILD_ID:0:8}
         # Define shared variables
+        SAFE_BRANCH=$(echo "$${BRANCH_NAME:-unknown}" | tr -c 'a-zA-Z0-9_.-' '-')
+        echo "export BRANCH_NAME=$${SAFE_BRANCH}" >> /workspace/build_vars.env
+
         echo "export RUN_ID=$${BUILD_ID}" >> /workspace/build_vars.env
         echo "export REGIONAL_BUCKET=${_INFRA_PREFIX}-regional-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
         echo "export ZONAL_BUCKET=${_INFRA_PREFIX}-zonal-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
@@ -253,8 +257,8 @@ steps:
             DATE_DIR=\$(date +%d%m%Y)
             RESULTS_DIR='gcsfs/tests/perf/microbenchmarks/__run__'
             if [ -d \"\$$RESULTS_DIR\" ]; then
-              echo \"Uploading from \$$RESULTS_DIR to gs://$${RESULTS_BUCKET}/\$${DATE_DIR}/$${RUN_ID}/\"
-              cd \"\$$RESULTS_DIR\" && gcloud storage cp --recursive . gs://$${RESULTS_BUCKET}/\$${DATE_DIR}/$${RUN_ID}/ && rm -rf *
+              echo \"Uploading from \$$RESULTS_DIR to gs://$${RESULTS_BUCKET}/branch=$${BRANCH_NAME}/\$${DATE_DIR}/$${RUN_ID}/\"
+              cd \"\$$RESULTS_DIR\" && gcloud storage cp --recursive . gs://$${RESULTS_BUCKET}/branch=$${BRANCH_NAME}/\$${DATE_DIR}/$${RUN_ID}/ && rm -rf *
             else
               echo \"No results directory found at \$$RESULTS_DIR\"
             fi
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -6,6 +6,7 @@ substitutions:
   _BENCHMARK_CONFIG: ""
   # Format: "hns regional zonal"
   _BUCKET_TYPES: ""
+  _RESULTS_BUCKET: ""
 
 steps:
   # 1. Generate a persistent SSH key for this build run.
@@ -33,13 +34,20 @@ steps:
         SHORT_BUILD_ID=$${BUILD_ID:0:8}
         # Define shared variables
         SAFE_BRANCH=$(echo "$${BRANCH_NAME:-unknown}" | tr -c 'a-zA-Z0-9_.-' '-')
+
+        # Error out if the branch is unknown
+        if [ "$${SAFE_BRANCH}" = "unknown" ]; then
+          echo "ERROR: BRANCH_NAME is not set or evaluates to unknown. Failing the build."
+          exit 1
+        fi
+
         echo "export BRANCH_NAME=$${SAFE_BRANCH}" >> /workspace/build_vars.env
 
         echo "export RUN_ID=$${BUILD_ID}" >> /workspace/build_vars.env
         echo "export REGIONAL_BUCKET=${_INFRA_PREFIX}-regional-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
         echo "export ZONAL_BUCKET=${_INFRA_PREFIX}-zonal-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
         echo "export HNS_BUCKET=${_INFRA_PREFIX}-hns-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
-        echo "export RESULTS_BUCKET=${_INFRA_PREFIX}-run-results" >> /workspace/build_vars.env
+        echo "export RESULTS_BUCKET=${_RESULTS_BUCKET}" >> /workspace/build_vars.env
         echo "export VM_NAME='${_INFRA_PREFIX}-vm-$${SHORT_BUILD_ID}'" >> /workspace/build_vars.env
     waitFor: ["-"]
 
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -6,7 +6,6 @@ substitutions:
   _BENCHMARK_CONFIG: ""
   # Format: "hns regional zonal"
   _BUCKET_TYPES: ""
-  _RESULTS_BUCKET: ""
   # Target benchmark workload (e.g., "IO" or "METADATA")
   _BENCHMARK_TYPE: ""
   _IO_VM_TYPE: "c4-standard-192"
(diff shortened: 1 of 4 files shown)


cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
```diff
@@ -31,11 +26,15 @@ steps:
     entrypoint: "bash"
     env:
       - "BUILD_ID=${BUILD_ID}"
+      - "BRANCH_NAME=${BRANCH_NAME}"
     args:
       - "-c"
       - |
         SHORT_BUILD_ID=$${BUILD_ID:0:8}
         # Define shared variables
+        SAFE_BRANCH=$(echo "$${BRANCH_NAME:-unknown}" | tr -c 'a-zA-Z0-9_.-' '-')
```
zhixiangli, line 40: nit: should we directly fail this step if the branch name is unknown?

zhixiangli (approved):

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
googlyrahman: I think, it's fine to run the benchmarks on `unknown` track, rather than failing.

ankitaluthra1:
@Mahalaxmibejugam Can you please review this

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
```diff
@@ -1,5 +1,6 @@
 substitutions:
   _DATASET_NAME: ""
+  _RESULTS_BUCKET: ""
```
Mahalaxmibejugam: Are we also going to modify the ingestion trigger to add this new substitution variable?

Is the plan to create multiple triggers each pointing to a different bucket and a branch?

cloudbuild/benchmarks/ingest.sql
```diff
@@ -37,6 +38,7 @@ SELECT
   REGEXP_EXTRACT(_FILE_NAME, r'/([0-9a-fA-F-]{36})/') as build_id,
   PARSE_TIMESTAMP('%d%m%Y-%H%M%S', REGEXP_EXTRACT(_FILE_NAME, r'/(\d{8}-\d{6})/')) as run_timestamp,
   _FILE_NAME as source_uri,
+  REGEXP_EXTRACT(_FILE_NAME, r'/branch=([^/]+)/') as branch_name,
```
Mahalaxmibejugam, line 41: What is the plan with DATASET_NAME? Is it going to be same for all the branches?

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
googlyrahman: You can either have a different bucket, or a different branch. Currently all the ingestions are hardcoded to a bucket which means running on any branch would land and mix in same bucket. Since there are no branch tracking, we don't have a way to differentiate things.

I'm thinking to maintain a different dataset, as branch would need rebasing every time fsspec have a commit, and yes we need to add a separate trigger for this with different dataset, and result bucket.

cloudbuild/benchmarks/ingest.sql
googlyrahman: Yes, it is going to be same for all branches. we've one column which tracks branch name, the output in the table will be of this form

`branch name, block size, io size, throughput, max memory, etc...`

googlyrahman:
@Mahalaxmibejugam could you please complete your review, and @ankitaluthra1 please merge it afterwards.

cloudbuild/benchmarks/ingest.sql
Mahalaxmibejugam: Just to confirm my understanding: we are planning for two history tables. One will be specific to the main branch, where the branch name column will explicitly contain "main," and the other will hold results from different branches, where the branch name can be any random branch name?

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
Mahalaxmibejugam: Since we are adding a new column in the history table (which points to the branch name) it will allow us to distinguish branch-wise results even if the results are uploaded to the same bucket, why do we still need a different bucket for storing the results?

cloudbuild/benchmarks/ingest.sql
Mahalaxmibejugam: This understanding is based on the other comment

> I'm thinking to maintain a different dataset, as branch would need rebasing every time fsspec have a commit, and yes we need to add a separate trigger for this with different dataset, and result bucket.

I think we should maintain a single history table that includes results from all branches, including main. Could you please clarify why we decided on a different dataset? Is this to prevent interference with the existing microbenchmarks dashboard?

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
Mahalaxmibejugam: In benchmarks-cloudbuild.yaml, the pipeline hardcodes the bucket name to ${_INFRA_PREFIX}-run-results where the results are uploaded. In benchmarks-ingestion-cloudbuild.yaml, the pipeline requires the user to know about this hardcoding and pass the generated bucket name in full as _RESULTS_BUCKET.

We must ensure these two substitution variables always remain in sync. Otherwise ingestion will not work as expected. To improve this, we can stop relying on the hardcoded "-run-results" suffix for the buckets and define _RESULTS_BUCKET as a mandatory substitution in both pipelines or use _INFRA_PREFIX here instead of new substitution variable.

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
zhixiangli: It's probably a waste of time and resources; the user wants to run on their own branch, but that leads to an unknown that is difficult to debug. We should just fail the process and let the user know there is an env issue.

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
googlyrahman: I think, my previous answer bought some confusion here, let me clarify what i want to do:

- I want to  run the performance benchmarks with an environment variable set, and without an environment variable set.
- This particular PR enables that support. It enables me to run benchmarks on different branch in the same dataset, and enables me to use different dataset if i wish.

Deciding between a bucket or a dataset is outside the scope of this PR. Both options work fine, but let me clarify my thought process.

The main drawback to using branches is the constant need to rebase against main. Here is my specific use case:
- I need to run triggers both with and without an environment variable. This requires two separate triggers. If I point both to the same branch, the results will merge.
- To fix that, I would need two branches: one with the environment variable and one without. I want one branch to stay on main to see how it performs with the latest commits. **For a true apple-to-apple comparison, the second branch would also need regular rebasing against main (This is human effort, and i want to avoid doing that over time)**

We could potentially solve the first issue by introducing a labeling system which basically do not mix the results on the same branch, but I am not sure if that would get enough use to justify the effort and code maintainability.

For now, branch wise is something we need to allow developers to track their performance against their branch, but labels inside a branch isn't that much required in my opinion.

cloudbuild/benchmarks/ingest.sql
googlyrahman: Responded on another comment.

Let's not mix discussions as discussion is active on another thread. Resolving it.

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
googlyrahman: Yes, I think we should remove that hardcoded reference too.

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
googlyrahman: I don't have a strong preference here, if you find value in failing it - I've added a commit to do that.

Also user would not run this directly, the cloud builder would run this, and cloud builder propagate branch name environment variable, so this is essentially a dead condition.

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
googlyrahman: Removed the hardcoded reference.

Mahalaxmibejugam: Please make sure to add _RESULTS_BUCKET to substitution variables of the benchmarks trigger to avoid any interruption to daily runs.

Mahalaxmibejugam: I understand the need for branch-wise tracking, but I am wondering: since we already include the branch name in both the bucket folder path and the history table, is it still mandatory to use a separate bucket when implementing branch-wise tracking?

I am aligned with using the bucket name as a substitution variable, but I am trying to understand if this is mandatory or if it is only included to provide the flexibility to use any bucket. Essentially, my question is: If we add the branch name to the table and bucket(as did in this PR), will your use case still work if we use the same bucket?

Mahalaxmibejugam (approved):

cloudbuild/benchmarks/benchmarks-ingestion-cloudbuild.yaml
googlyrahman: No, there's no need for having a different bucket if we're using branch-wise tracking, and i'm not creating a new one.

googlyrahman: The need for bucket only exists if we're using a new dataset, and want to isolate things.

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
```diff
@@ -31,16 +32,27 @@ steps:
...
+        fi
+
+        echo "export BRANCH_NAME=$${SAFE_BRANCH}" >> /workspace/build_vars.env
+
         echo "export RUN_ID=$${BUILD_ID}" >> /workspace/build_vars.env
         echo "export REGIONAL_BUCKET=${_INFRA_PREFIX}-regional-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
         echo "export ZONAL_BUCKET=${_INFRA_PREFIX}-zonal-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
         echo "export HNS_BUCKET=${_INFRA_PREFIX}-hns-$${SHORT_BUILD_ID}" >> /workspace/build_vars.env
-        echo "export RESULTS_BUCKET=${_INFRA_PREFIX}-run-results" >> /workspace/build_vars.env
+        echo "export RESULTS_BUCKET=${_RESULTS_BUCKET}" >> /workspace/build_vars.env
```
jasha26: I'd prefer to enforce _INFRA_PREFIX so i think we should default to ${_INFRA_PREFIX}-run-results if results bucket is not passed otherwise use results bucket with some other prefix.

jasha26 (approved):

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
googlyrahman: Done.

PR #822, opened 2026-04-23
Add mixed pattern in gcsfs microbenchmarks

googlyrahman:
The read benchmarks in our microbenchmarks are either (100% sequential 0% seek) or (0% sequential 100% seek) with fixed io size. These are often not correct representation of real world scenarios.

Real world scenarios are mixed of sequential, and seeks with a variable of `io` size. The following PR add this feature in microbenchmarks to correctly measure the typical real world performance. 

Sample output after this PR
```
| Bucket Type | Group | Pattern | Seq Prob | Files | Folders | Threads | Processes | Depth | Target Type | File Size (MiB) | Chunk Size (MiB) | Min Chunk (MiB) | Max Chunk (MiB) | Block Size (MiB) | Mean Latency (s) | Mean Throughput (MiB/s) | Max CPU (%) | Max Memory (MiB) |
+-------------+-------+---------+----------+-------+---------+---------+-----------+-------+-------------+-----------------+------------------+-----------------+-----------------+------------------+------------------+-------------------------+-------------+------------------+
|   regional  |  read |   seq   |   N/A    |   1   |   N/A   |    1    |     1     |  N/A  |     N/A     |     1024.00     |      16.00       |       N/A       |       N/A       |       5.00       |      30.00       |          143.47         |     0.47    |      224.88      |
|   regional  |  read |  mixed  |   0.3    |   1   |   N/A   |    1    |     1     |  N/A  |     N/A     |     1024.00     |       N/A        |       0.06      |     1024.00     |       5.00       |      30.00       |          328.90         |     1.04    |     2868.44      |
|   regional  |  read |  mixed  |   0.5    |   1   |   N/A   |    1    |     1     |  N/A  |     N/A     |     1024.00     |       N/A        |       0.06      |     1024.00     |       5.00       |      30.00       |          249.38         |     1.14    |     2934.86      |
+-------------+-------+---------+----------+-------+---------+---------+-----------+-------+-------------+-----------------+------------------+-----------------+-----------------+------------------+------------------+-------------------------+-------------+------------------+
```


gcsfs/tests/perf/microbenchmarks/read/test_read.py
```diff
@@ -48,6 +48,38 @@ def _read_op_rand(gcs, file_paths, chunk_size, offsets, runtime):
...
+    total_bytes_read = 0
+    start_time = time.perf_counter()
+    files_it = itertools.cycle(file_paths)
+
+    while time.perf_counter() - start_time < runtime:
+        path = next(files_it)
+        with gcs.open(path, "rb") as f:
+            while time.perf_counter() - start_time < runtime:
+                if random.random() >= seq_probability:
+                    max_offset = max(0, file_size_bytes - max_chunk_size)
```
zhixiangli: When `file_size_bytes <= max_chunk_size`, `max_offset` is always 0, resulting in `f.seek(0)` every time. We should instead randomize the `chunk_size` first and then randomize the `offset`.

zhixiangli (changes requested):

gcsfs/tests/perf/microbenchmarks/read/test_read.py
googlyrahman: Makes sense!

zhixiangli (approved):

jasha26 (approved):

ankitaluthra1:
Can you please fix the test failure so its ready for merge

googlyrahman:
Aah this needs to rebased with main to fix the failure, let me do it.

PR #824, opened 2026-04-28
Add workflow to automate PyPI package publishing on release

Yonghui-Lee:
I add a new GitHub workflow to automate the process of publishing the gcsfs package to PyPI when a new release is published.

In addition, we need to config the workflow as a Trusted Publisher in the PyPI project.

Closes #791


.github/workflows/pypipublish.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.x"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install hatch
```
zhixiangli: Are tests excluded?

.github/workflows/pypipublish.yml
```diff
@@ -0,0 +1,29 @@
+name: Upload Python Package
+
+on:
+  release:
```
zhixiangli: Will there be a script to publish release?

.github/workflows/pypipublish.yml
Yonghui-Lee: `hatch build` will exclude tests according to `pyproject.toml`

.github/workflows/pypipublish.yml
Yonghui-Lee: I think release will be manually executed. The workflow is that a maintainer releases a version and the pipeline uploads it to PyPI.

zhixiangli (approved):

martindurant:
What do you need from me to enable this workflow?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '+([0-9]).+([0-9]).*'
```
martindurant: Dp we want to be more specific here? I think this matches more boradly than "2026.04.0[post|dev]" pattern we have.

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
+name: Release
+
+on:
+  push:
```
martindurant, line 6: This runs on a tag, not a GH release - but we create a release in the run below. This should be documented, pushing the tag is what triggers a push to pypi.

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.x"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install hatch
+      - name: Build
+        run: hatch build
```
martindurant: `python -m build` is more canonical in my opinion. Also, it should fetch hatch automatically I think. 

We also need hatch-vcs?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          generate_release_notes: true
+          files: dist/*
+
+  publish:
+    name: Upload release to PyPI
+    needs: release
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
```
martindurant: pypa/gh-action-pypi-publish docs only mention the id-token permission. What does this one do?

.github/workflows/release.yml
Yonghui-Lee: GitHub Actions workflows use a custom glob matcher which only supports basic wildcard symbols. It's difficult to express the exact constraint.

```
on:
  push:
    tags:
      - '2[0-9][0-9][0-9].[0-9].[0-9]'
      - '2[0-9][0-9][0-9].[0-9][0-9].[0-9]'
      - '2[0-9][0-9][0-9].[0-9].[0-9]post[0-9]'
      - '2[0-9][0-9][0-9].[0-9][0-9].[0-9]post[0-9]'
      - '2[0-9][0-9][0-9].[0-9].[0-9]dev[0-9]'
      - '2[0-9][0-9][0-9].[0-9][0-9].[0-9]dev[0-9]'
```

Given that release tags are only pushed by maintainers, I think we may accept the looseness

.github/workflows/release.yml
Yonghui-Lee: Done.

.github/workflows/release.yml
Yonghui-Lee: Change to use build. Hatch and hatch-vcs will be fetched automatically.

.github/workflows/release.yml
Yonghui-Lee: Removed.

Yonghui-Lee:
> What do you need from me to enable this workflow?

We need to config the workflow as a Trusted Publisher on PyPI.

.github/workflows/release.yml
lmmx: an idiomatic approach is to check if it ends in dev like this: https://github.com/astropy/astropy/blob/4ea3d15b03f8869d71a730bfae7df621dac915f3/.github/workflows/publish.yml#L79

gating the job with an `if`

lmmx:
> > What do you need from me to enable this workflow?
> 
> We need to config the workflow as a Trusted Publisher on PyPI.

Here is a checklist, not sure if this all got done but without it the release will fail to publish

- [ ] Package owner on PyPI has to set up Trusted Publishing on PyPI (else the workflow will fail)
  - Note that this can only be done by the package admin (a maintainer is not necessarily an admin)
  - PyPI will request:
    - repo owner: `fsspec`
    - repo ID `gcsfs`
    - CI workflow name `release.yml`
    - environment name (typically `pypi`)
- [x] (Will be made automatically) ~~Create a “pypi” environment in the GitHub repo to match the one registered with PyPI~~
  - or can configure manually: https://github.com/fsspec/gcsfs/settings/environments/new
- [ ] Delete PyPI API key secrets from repo CI settings (not sure if this was done)
- [ ] Delete/invalidate PyPI API keys if no longer needed

No environment was set in the workflow here, I think it’s best to set one as it lets you add security controls. Example: https://github.com/astropy/astropy/blob/4ea3d15b03f8869d71a730bfae7df621dac915f3/.github/workflows/publish.yml#L83

PR #831, opened 2026-05-04
Add support for partial prefixes in find method for HNS buckets

Mahalaxmibejugam:
Fsspec recently made changes in [PR](https://github.com/fsspec/filesystem_spec/pull/1996/changes) to glob method where it passes partial prefixes to backends like gcsfs to the find method. Listing files in both standard buckets and HNS buckets handles the partial prefixes but fetching the directories in HNS buckets doesn't handle the partial prefixes. This PR includes that support

Implementation details:

   1. Fixed HNS Folder Listing (gcsfs/extended_gcsfs.py):
       * Refactored _get_all_folders to correctly handle partial prefixes. It now calculates the deepest
         valid parent directory for the API call (e.g., folder_ searches in the current directory, while
         sub/folder_ searches in sub/).
       * Applied local filtering to the API results to support partial name matches

   2. Improved Consistency in Standard GCS (gcsfs/core.py):
       * Updated GCSFileSystem._find to ensure that synthesized directories also match the provided
         prefix. 
       * Previously, standard GCS would return the search root (the path argument) even if a prefix was
         provided that the root did not match. Now, both standard and HNS buckets consistently only 
         include the directories that match prefix, matching the behavior for files.

   3. Enhanced Testing:
       * HNS Unit Tests: Added test_hns_find_partial_prefix in gcsfs/tests/test_extended_hns_gcsfs.py to
         verify API request construction and result filtering.
       * Standard GCS Parity: Added test_find_withdirs_prefix in gcsfs/tests/test_core.py to enforce and
         document the consistent behavior across bucket types.
       * Integration Tests: Added test_find_partial_prefix in
         gcsfs/tests/integration/test_extended_hns.py for end-to-end validation.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -1152,26 +1152,43 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         Recursively fetches all folder objects under a given path using the
         Storage Control API.
         """
-        folders = []
-        base_path = self.split_path(path)[1].rstrip("/")
-        full_prefix = f"{base_path}/{prefix}".strip("/") if base_path else prefix
+        _, base_path, _ = self.split_path(path)
+        base_path = "" if not base_path else base_path.rstrip("/") + "/"
+        full_prefix = f"{base_path}{prefix}"
 
-        folder_id = full_prefix
-        if folder_id and not folder_id.endswith("/"):
-            folder_id += "/"
+        # To find folders matching a partial prefix, we need to list their parent.
+        if full_prefix and not full_prefix.endswith("/"):
+            start_dir = full_prefix.rpartition("/")[0] + "/"
+        else:
+            start_dir = full_prefix
+
+        folders = []
+        client = await self._get_control_plane_client()
         parent = f"projects/_/buckets/{bucket}"
+
         request = storage_control_v2.ListFoldersRequest(
-            parent=parent, prefix=folder_id, request_id=str(uuid.uuid4())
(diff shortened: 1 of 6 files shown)


gcsfs/extended_gcsfs.py
```diff
@@ -1152,26 +1179,43 @@ async def _get_all_folders(self, path, bucket, prefix=""):
...
-        folders = []
-        base_path = self.split_path(path)[1].rstrip("/")
-        full_prefix = f"{base_path}/{prefix}".strip("/") if base_path else prefix
+        _, base_path, _ = self.split_path(path)
+        base_path = "" if not base_path else base_path.rstrip("/") + "/"
+        full_prefix = f"{base_path}{prefix}"
+
+        # To find folders matching a partial prefix, we need to list their parent.
+        if full_prefix and not full_prefix.endswith("/"):
+            start_dir = full_prefix.rpartition("/")[0] + "/"
```
zhixiangli: Could you add a unit test for when full_prefix.rpartition("/")[0] is empty, the start_dir becomes "/" while the expected result is ""?

```
parent, sep, _ = full_prefix.rpartition("/")
start_dir = parent + sep
```

zhixiangli (changes requested):

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: Done

zhixiangli (approved):

Issue #833, opened 2026-05-06
Wrong exception raised by GCSFileSystem._get_file_request


gilesbrown:
Issue:
When using gcsfs we are seeing a `TypeError` raised rather than an `HTTPError` for a 429 response (too many requests).  This is problematic because it is not retried properly.

Summary:
The passing of `None` into `validate_response` [here](https://github.com/fsspec/gcsfs/blob/5cb0d32633f32dfd2fa15667e55127244a5b6a08/gcsfs/core.py#L1907) causes the code to raise a `TypeError` when the status code is any 400 code other than a 404. This happens because the function tries to read JSON from the `content` passed in as `None`.

Steps to Reproduce:

This `pytest` test demonstrates the problem (dependencies pytest, gcsfs, aioresponses, pytest-asyncio).
```
import pytest
import gcsfs
from aioresponses import aioresponses

@pytest.mark.parametrize("status", [404, 429])
@pytest.mark.asyncio
async def test_gcsfs_get_file_request_by_status(status):
    fs = gcsfs.GCSFileSystem()

    rpath = f"my-bucket/{status}.txt"
    lpath = f"{status}/txt"

    with aioresponses() as mocked:
        mocked.get(rpath, status=status)

        with pytest.raises(Exception) as excinfo:
            await fs._get_file_request(rpath, lpath)

    # The exception should not be a TypeError but is for a 429 status
    assert type(excinfo.value) is not TypeError
```

Expected Behavior:
An `HTTPError` is raised that will support retry operations correctly.

Actual Behavior:
A `TypeError` is raised and retries are prevented.

gilesbrown:
Awesome! Thanks you for the quick response!

PR #835, opened 2026-05-06
Fix HttpError message formatting and handle None content in validate_response

zhixiangli:
Fix for https://github.com/fsspec/gcsfs/issues/833

## Summary
- Handle `content=None` in `validate_response` to avoid `TypeError` when attempting to parse JSON.
- Handle non-JSON error responses in `validate_response` by using the raw content as the error message instead of failing to parse.
- Add test coverage for `validate_response` with `content=None` and invalid JSON content.

## Test Plan
- [x] Added `test_validate_response_content_none` and `test_validate_response_invalid_json` in `gcsfs/tests/test_retry.py`.
- [x] Verified that these tests pass and correctly raise `HttpError` instead of `TypeError`.

### DIFF
--- a/gcsfs/retry.py
+++ b/gcsfs/retry.py
@@ -114,17 +114,19 @@ def validate_response(status, content, path, args=None):
             raise FileNotFoundError(path)
 
         error = None
-        if hasattr(content, "decode"):
-            content = content.decode()
-        try:
-            error = json.loads(content)["error"]
-            # Sometimes the error message is a string.
-            if isinstance(error, str):
-                msg = error
-            else:
-                msg = error["message"]
-        except json.decoder.JSONDecodeError:
-            msg = content
+        msg = ""
+        if content:
+            if hasattr(content, "decode"):
+                content = content.decode()
+            try:
+                error = json.loads(content)["error"]
+                # Sometimes the error message is a string.
+                if isinstance(error, str):
+                    msg = error
+                else:
+                    msg = error["message"]
+            except json.decoder.JSONDecodeError:
+                msg = content
 
         if status == 403:
             raise OSError(f"Forbidden: {path}\n{msg}")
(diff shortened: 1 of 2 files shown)


gcsfs/retry.py
```diff
@@ -114,17 +114,19 @@ def validate_response(status, content, path, args=None):
...
+            if hasattr(content, "decode"):
+                content = content.decode()
+            try:
+                error = json.loads(content)["error"]
+                # Sometimes the error message is a string.
+                if isinstance(error, str):
+                    msg = error
+                else:
+                    msg = error["message"]
+            except json.decoder.JSONDecodeError:
```
Yonghui-Lee, line 128: Will it be more robust if we also catch KeyError, TypeError?

For example, If the returned JSON is valid but does not contain the "error" key, a KeyError is raised.

zhixiangli: That is the original implementation and falls outside the scope of this PR. This change is intended to avoid passing None/"" to the JSON parser, which reduces overhead and resolves the issue.

Yonghui-Lee: It’s kind of related. I think getting message from content is not critical to fail, and we can set the message to content or empty.

Edge cases keep coming up. Last year, we had to handle the case where error is a string. Now, we need to handle that content is None. We can catch more errors to fix it for all. For example, passing none can be handled because it raises TypeError.

zhixiangli: The root cause of the TypeError is actually `json.loads(content)` failing when content is None, rather than the `["error"]` key access itself.

Since passing `content=None` is a deliberate design choice in our codebase (used to validate headers before downloading the payload), it represents an expected, valid state rather than an exceptional one. Handling this expected state explicitly with if content: is much cleaner and more efficient than letting the JSON parser intentionally fail and then catching the resulting `TypeError`.

Yonghui-Lee (approved):

PR #840, opened 2026-05-11
Implement zero-copy optimization for single-read operations

googlyrahman:
### Implement zero-copy optimization for single-write operations

Currently, when `gcsfs` calls `_cat_file`, it allocates memory and passes it to the underlying SDK to write into. Since gRPC typically returns payloads in 2MB chunks, any user operation requesting less than 2MB will only trigger a single write to the buffer.

Because there is only one chunk, we do not need to assemble multiple pieces of data. This means allocating new memory is unnecessary. Instead, we can just store and return a reference to the bytes originally returned by gRPC. If we know no assembly is required, it makes sense to skip the extra allocation step and return the exact same bytes passed to us.

This PR also refactors the buffer class to keep its logic in one place. Previously, the memory allocation logic lived inside `_cat_file` while the memory move logic was handled in `zb_hns_util.py`. Now, both allocation and memory movement are managed within the same isolated

While you may see after performing better, I feel this is just a network variance, and not actual improvement. This data just shows that performance doesn't degrade with this change. This is because, without this change as well we touch 2.2 GB/s - it's just about time :)

<details>
<summary>Numbers before Change</summary>

### Before
| threads | io size (MB) | pattern | cache type | default throughput | concurrency | prefetcher+concurrency | default mem | concurrency mem | prefetcher mem |
|---|---|---|---|---|---|---|---|---|---|
| 1 | 16.00 | sequential | none | 721.35 | 1127.78 | 1891.62 | 266.13 | 320.71 | 483.95 |
| 1 | 1.00 | sequential | none | 179.34 | 183.32 | 1153.28 | 388.35 | 385.72 | 541.94 |
| 1 | 0.06 | sequential | none | 26.84 | 24.86 | 337.24 | 382.19 | 380.15 | 534.52 |
| 1 | 16.00 | random | none | 1004.04 | 1121.51 | 1157.96 | 478.91 | 444.98 | 460.21 |
| 1 | 1.00 | random | none | 199.52 | 195.88 | 178.36 | 450.32 | 407.06 | 406.30 |
| 1 | 0.06 | random | none | 23.95 | 30.97 | 26.24 | 398.33 | 385.62 | 385.32 |
| 16 | 16.00 | sequential | none | 2242.48 | 2019.02 | 2036.10 | 1401.24 | 1533.28 | 4013.57 |
| 16 | 1.00 | sequential | none | 1259.63 | 1258.52 | 1481.27 | 3288.06 | 950.04 | 3754.41 |
| 16 | 0.06 | sequential | none | 161.75 | 161.05 | 449.21 | 3611.83 | 1201.81 | 1817.69 |
| 16 | 16.00 | random | none | 2185.34 | 1962.58 | 1915.06 | 2198.41 | 2356.74 | 2354.56 |
| 16 | 1.00 | random | none | 1207.06 | 1185.41 | 1119.91 | 1555.76 | 1551.03 | 1540.96 |
| 16 | 0.06 | random | none | 153.89 | 155.04 | 138.66 | 1500.15 | 1362.15 | 1345.37 |
| 32 | 16.00 | sequential | none | 2151.50 | 1864.77 | 1792.89 | 3048.99 | 3352.23 | 8224.82 |
| 32 | 1.00 | sequential | none | 1338.03 | 1364.94 | 1306.88 | 7984.98 | 2102.76 | 7143.57 |
| 32 | 0.06 | sequential | none | 166.18 | 165.81 | 422.57 | 2088.15 | 2028.59 | 2870.22 |
| 32 | 16.00 | random | none | 2162.62 | 1837.32 | 1861.27 | 3976.30 | 4295.86 | 4321.93 |
| 32 | 1.00 | random | none | 1298.25 | 1301.49 | 1253.09 | 4274.32 | 2697.81 | 2697.52 |
| 32 | 0.06 | random | none | 163.22 | 165.90 | 152.75 | 2675.91 | 2410.22 | 2381.91 |

</details>

<details>
<summary>Numbers After Change</summary>

### After

| threads | io size (MB) | pattern | cache type | default throughput | concurrency | prefetcher+concurrency | default mem | concurrency mem | prefetcher mem |
|---|---|---|---|---|---|---|---|---|---|
| 1 | 16.00 | sequential | none | 791.44 | 1117.53 | 2124.81 | 323.39 | 331.88 | 478.51 |
| 1 | 1.00 | sequential | none | 189.05 | 201.20 | 1421.31 | 392.52 | 394.18 | 467.52 |
| 1 | 0.06 | sequential | none | 27.95 | 27.62 | 347.78 | 354.49 | 353.52 | 502.54 |
| 1 | 16.00 | random | none | 862.19 | 1197.17 | 1096.15 | 448.83 | 429.15 | 430.60 |
| 1 | 1.00 | random | none | 183.73 | 181.87 | 234.73 | 382.70 | 383.69 | 382.17 |
| 1 | 0.06 | random | none | 26.91 | 29.58 | 28.74 | 362.57 | 362.71 | 359.82 |
| 16 | 16.00 | sequential | none | 2249.92 | 2033.24 | 2092.52 | 1333.67 | 1508.81 | 3301.77 |
| 16 | 1.00 | sequential | none | 1387.34 | 1409.44 | 1511.73 | 794.82 | 814.70 | 3155.70 |
| 16 | 0.06 | sequential | none | 214.92 | 214.22 | 472.76 | 1031.62 | 1030.50 | 1571.19 |
| 16 | 16.00 | random | none | 2227.98 | 2010.46 | 2018.08 | 2029.57 | 2164.73 | 2194.00 |
| 16 | 1.00 | random | none | 1378.09 | 1370.59 | 1242.10 | 2150.26 | 1354.07 | 1362.34 |
| 16 | 0.06 | random | none | 205.90 | 206.04 | 181.90 | 1205.16 | 1203.26 | 1194.04 |
| 32 | 16.00 | sequential | none | 2233.26 | 1936.31 | 1740.38 | 2876.52 | 3186.90 | 6463.75 |
| 32 | 1.00 | sequential | none | 1513.23 | 1540.23 | 1376.60 | 3776.41 | 1926.75 | 6723.48 |
| 32 | 0.06 | sequential | none | 218.94 | 218.97 | 453.59 | 2104.26 | 2063.58 | 2960.17 |
| 32 | 16.00 | random | none | 2256.55 | 1943.16 | 1915.21 | 3979.39 | 4294.62 | 4414.93 |
| 32 | 1.00 | random | none | 1501.28 | 1496.94 | 1412.68 | 3255.36 | 2743.42 | 2690.53 |
| 32 | 0.06 | random | none | 212.00 | 210.28 | 191.67 | 2631.54 | 2398.37 | 2380.11 |

</details>


gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,114 @@ def _decrement_pending(self):
...
+                    fut.set_result(None)
+                return fut
+            else:
+                # Slow path, schedule it on executor.
+                return self.executor.submit(self._do_memmove, dest, data_bytes, size)
         except BaseException as e:
-            self._error = e
+            with self._lock:
+                self._error = e
             self._decrement_pending()
```
zhixiangli: In the finally block of `_do_memmove`, wouldn't calling `_decrement_pending` cause a duplicate decrement?

gcsfs/extended_gcsfs.py
```diff
@@ -461,18 +453,17 @@ async def _download(o, s, b, mrd_or_pool):
...
-                    # If we are already handling a network/download exception,
-                    # ignore the BufferError (which is just a symptom of the drop).
-                    # If there's no download error, this means the buffer logic
-                    # itself failed, so we must surface the error.
-                    if not has_error:
-                        raise
-
-        return result_bytes
+            try:
+                master_buffer.close()
```
zhixiangli, line 535: Large reads can block the asyncio loop while waiting for memmove, as it calls master_buffer.close() synchronously within the finally block of _concurrent_mrd_fetch. Since this coroutine executes directly on the main asyncio event loop thread, this causes a block.

zhixiangli (changes requested):
> Since gRPC typically returns payloads in 2MB chunks, any user operation requesting less than 2MB will only trigger a single write to the buffer.

Could you elaborate on this? My understanding of gRPC in Python is that we join all chunks, deserialize the data via Protobuf, and then trigger the data_buffer write. The data_buffer write interacts with the Python Protobuf upb memory. How does the 2MB limit help?

gcsfs/extended_gcsfs.py
googlyrahman: This was actually an intentional change, and I want to block the async event loop in this specific case. I know it seems like an anti-pattern, but it is required for performance in these heavily multithreaded scenarios.

When testing with 32 coroutines, I found that running `master.close()` synchronously was about 500 MB/s faster than using `await loop.run_in_executor(None, buff.close)` with SDK. Because the system is completely bombarded by `ctypes.memmove` operations, the overhead of offloading the close operation becomes too high.

If you look at the previous implementation, the initial write operation using `io.BytesIO` was synchronous and consumed the GIL. We were already blocking the event loop to write data into the buffer. With the new approach, we are simply blocking the event loop to wait for a thread to finish writing that data. The end result on the event loop is effectively the same, but with much better throughput.

I'm open for suggestions, so feel free to let me know if we can do it in a better way :) - The only condition here is that we need a synchronous `def write` method to a buffer (similar to `io.BytesIO` interface) called from async method.

Regarding the concern about large reads, I do not think it will cause problems. Downloading large blocks of data over the network will inherently take significantly longer than writing that data into the buffer.

I have included a script below for your reference. If you run it with `master.close()`, you will see it hits over 2 GB/s. If you switch to `await loop.run_in_executor(None, master.close)`, it drops to around 1.4 GB/s on average.

<details>
<summary>Script</summary>

```
import asyncio
import argparse
import logging
import os
import random
import time
import warnings
import ctypes
import threading
import concurrent.futures
from google.cloud.storage.asyncio.async_grpc_client import AsyncGrpcClient
from google.cloud.storage.asyncio.async_multi_range_downloader import AsyncMultiRangeDownloader 

os.environ["GRPC_VERBOSITY"] = "ERROR"
warnings.filterwarnings(
    "ignore",
    category=FutureWarning,
    module="google.api_core._python_version_support"
)

BUCKET_NAME = None
OBJECT_NAME = None
TOTAL_OBJECT_SIZE = 20 * 1024 * 1024 * 1024
NUM_RUNS = 5


PyBytes_FromStringAndSize = ctypes.pythonapi.PyBytes_FromStringAndSize
PyBytes_FromStringAndSize.argtypes = (ctypes.c_void_p, ctypes.c_ssize_t)
PyBytes_FromStringAndSize.restype = ctypes.py_object

PyBytes_AsString = ctypes.pythonapi.PyBytes_AsString
PyBytes_AsString.argtypes = (ctypes.py_object,)
PyBytes_AsString.restype = ctypes.c_void_p

class DirectMemmoveBuffer:
    """
    A buffer-like object that writes data directly to memory asynchronously.

    This class provides an interface that queues `ctypes.memmove` operations
    to a thread pool executor. It provides synchronous backpressure: if `max_pending`
    operations are currently writing, the `write()` call will safely block the
    calling thread (e.g., an asyncio loop) until capacity frees up.

    Memory allocation is natively deferred. If the payload precisely aligns
    with expected bounds sequentially, it gracefully overrides manual memmoves
    using true Zero-Copy payload replacement safely under the hood.

    Note: This class is now strictly Thread-Safe
    """

    THRESHOLD_BYTES_FOR_SCHEDULING = 128 * 1024

    def __init__(self, expected_size, executor, max_pending=5):
        """
        Initializes the DirectMemmoveBuffer.

        Args:
            expected_size (int): The total amount of bytes expected to populate memory.
            executor (concurrent.futures.Executor): The thread pool executor to run the
                memmove operations. The lifecycle of this executor is managed by the caller.
            max_pending (int, optional): The maximum number of pending write operations
                allowed in the queue. Defaults to 5.
        """
        self.expected_size = expected_size
        self.executor = executor

        self._pending_count = 0
        self._error = None
        self._total_bytes_written = 0
        self._stop_accepting_writes = False
        self._is_closed = False

        self._allocated_intervals = []
        self._result_bytes = None
        self._start_address = None
        self.semaphore = threading.Semaphore(max_pending)
        self._lock = threading.Lock()
        self._done_event = threading.Event()
        self._done_event.set()

    class PartialView:
        """A bounded memory writer providing robust overfill/underfill constraint validations."""

        def __init__(self, parent, start_offset, expected_size):
            self.parent = parent
            self.start_offset = start_offset
            self.expected_size = expected_size
            self.current_offset = 0
            self._view_lock = threading.Lock()

        def write(self, data):
            """
            Schedules a write operation to memory mapping.
            """
            if not isinstance(data, bytes):
                raise ValueError(f"Expected bytes, but got {type(data)}")
            size = len(data)
            with self._view_lock:
                if self.current_offset + size > self.expected_size:
                    error_msg = (
                        f"Attempted to write {size} bytes "
                        f"at offset {self.current_offset}. "
                        f"Max capacity is {self.expected_size} bytes."
                    )
                    raise BufferError(error_msg)

                abs_offset = self.start_offset + self.current_offset
                self.current_offset += size

            return self.parent._submit_write(abs_offset, data, size)

        def close(self):
            """
            Validates boundaries enforcing complete local payload consistency.
            """
            with self._view_lock:
                if self.current_offset < self.expected_size:
                    error_msg = (
                        f"Expected {self.expected_size} bytes, "
                        f"but only received {self.current_offset} bytes. "
                        f"Buffer contains uninitialized data."
                    )
                    raise BufferError(error_msg)

    def get_view(self, offset, size):
        """Constructs secure mapped offset references correctly handling constraint layouts."""
        if offset < 0 or offset + size > self.expected_size:
            raise ValueError("Invalid view requested: exceeds physical boundaries!")

        start = offset
        end = offset + size

        with self._lock:
            if self._stop_accepting_writes or self._is_closed:
                raise ValueError("Cannot get view on a closed/closing buffer.")

            for a_start, a_end in self._allocated_intervals:
                if max(start, a_start) < min(end, a_end):
                    raise ValueError(
                        f"Overlapping view requested: [{start}, {end}) "
                        f"overlaps with already allocated view [{a_start}, {a_end})"
                    )

            self._allocated_intervals.append((start, end))

        return self.PartialView(self, offset, size)

    def _decrement_pending(self):
        """Helper to cleanly release concurrency primitives after a task finishes."""
        self.semaphore.release()
        with self._lock:
            self._pending_count -= 1
            if self._pending_count == 0:
                self._done_event.set()

    def _submit_write(self, dest_offset, data_bytes, size):
        self.semaphore.acquire()

        try:
            with self._lock:
                if self._stop_accepting_writes or self._is_closed:
                    raise ValueError("I/O operation on closed buffer.")

                if self._error:
                    raise self._error

                if self._result_bytes is None:
                    if dest_offset == 0 and size == self.expected_size:
                        self._result_bytes = data_bytes
                        self.semaphore.release()  # Release because we skip the executor
                        fut = concurrent.futures.Future()
                        fut.set_result(None)
                        self._total_bytes_written += size
                        return fut
                    else:
                        self._result_bytes = PyBytes_FromStringAndSize(
                            None, self.expected_size
                        )
                        self._start_address = PyBytes_AsString(self._result_bytes)

                # Defensive programming: gracefully catch internal overwrite attempts
                if self._start_address is None:
                    raise BufferError(
                        "Attempted to execute standard write over a Zero-Copied payload."
                    )

                dest = self._start_address + dest_offset
                if self._pending_count == 0:
                    self._done_event.clear()
                self._pending_count += 1

        except BaseException:
            self.semaphore.release()
            raise

        try:
            if size <= self.THRESHOLD_BYTES_FOR_SCHEDULING:
                # Fast path, no need to send it to executor
                self._do_memmove(dest, data_bytes, size)                
                fut = concurrent.futures.Future()
                with self._lock:
                    local_err = self._error
                if local_err:
                    fut.set_exception(local_err)
                else:
                    fut.set_result(None)
                return fut
            else:
                # Slow path, schedule it on executor.
                return self.executor.submit(self._do_memmove, dest, data_bytes, size)
        except BaseException as e:
            with self._lock:
                self._error = e
            self._decrement_pending()
            raise e

    def _do_memmove(self, dest, data_bytes, size):
        try:
            with self._lock:
                if self._error:
                    return

            ctypes.memmove(dest, data_bytes, size)
            with self._lock:
                self._total_bytes_written += size

        except BaseException as e:
            with self._lock:
                if self._error is None:
                    self._error = e
            raise
        finally:
            self._decrement_pending()

    def get_value(self):
        with self._lock:
            if self._error:
                raise self._error
            if not self._is_closed:
                raise RuntimeError("Buffer is still not closed yet!")
            if self._result_bytes is None and self.expected_size == 0:
                return b""
            if self._total_bytes_written < self.expected_size:
                raise BufferError(
                    f"Buffer incomplete: Expected {self.expected_size} bytes but "
                    f"only populated {self._total_bytes_written}. Returning this "
                    f"payload would leak uninitialized memory."
                )
            return self._result_bytes

    def close(self):
        """
        Locks the buffer preventing further incoming writes, waits for all pending
        write operations to complete, and checks for errors.
        """
        with self._lock:
            self._stop_accepting_writes = True

        self._done_event.wait()
        with self._lock:
            self._is_closed = True
            if self._error:
                raise self._error


async def async_worker_task(read_size, mode, warmup_secs, run_secs):
    """
    An individual task that loops and sequentially fetches data for a specific duration.
    Each task gets its own mrd, its own thread executor, and starts from 0 sequentially.
    """
    client = AsyncGrpcClient()
    mrd = await AsyncMultiRangeDownloader.create_mrd(
        client, 
        bucket_name=BUCKET_NAME, 
        object_name=OBJECT_NAME
    )
    
    file_offset = 0
    recorded_bytes = 0
    
    loop_start = time.monotonic()
    warmup_end = loop_start + warmup_secs
    test_end = warmup_end + run_secs
    
    loop = asyncio.get_running_loop()    
    with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
        while True:
            if time.monotonic() >= test_end:
                break
                
            xread = read_size
            
            if mode == "sequential":
                if file_offset + xread > TOTAL_OBJECT_SIZE:
                    xread = TOTAL_OBJECT_SIZE - file_offset
            else:
                max_offset = max(0, TOTAL_OBJECT_SIZE - read_size)
                file_offset = random.randint(0, int(max_offset))
                xread = read_size
                
            buffer = DirectMemmoveBuffer(
                expected_size=xread,
                executor=executor
            )
            
            view = buffer.get_view(0, xread)                
            await mrd.download_ranges([(file_offset, xread, view)])            
            view.close()
            await loop.run_in_executor(None, buffer.close)
            
            if time.monotonic() >= warmup_end:
                recorded_bytes += xread
            
            if mode == "sequential":
                file_offset += xread
                if file_offset >= TOTAL_OBJECT_SIZE:
                    file_offset = 0
            
    await mrd.close()
    if hasattr(client, 'close'):
        if asyncio.iscoroutinefunction(client.close):
            await client.close()
        else:
            client.close()
    return recorded_bytes


async def run_benchmark_step(current_read_size, mode, coro_count, warmup_secs, run_secs):
    """
    Spawns `coro_count` concurrent async tasks for one test phase.
    """
    tasks = []
    for i in range(coro_count):
        task = asyncio.create_task(
            async_worker_task(current_read_size, mode, warmup_secs, run_secs)
        )
        tasks.append(task)
        
    results = await asyncio.gather(*tasks)
    total_bytes = sum(results)
    
    total_gb = total_bytes / (1024**3)
    gb_per_sec = total_gb / run_secs if run_secs > 0 else 0
    return gb_per_sec


def async_main(args, chunk_size_bytes):
    print(f"Testing: {args.mb} MB chunks")
    
    print(f"   Warmup run (establishing TCP connections)...", end="", flush=True)
    warmup_throughput = asyncio.run(run_benchmark_step(chunk_size_bytes, args.mode, args.coro_count, 0, args.warmup))
    print(f" Done. (Ignored: {warmup_throughput:.2f} GB/s)")

    run_values = []
    for run_idx in range(1, NUM_RUNS + 1):
        print(f"   Run {run_idx}/{NUM_RUNS} (Benchmarking)...", end="", flush=True)
        throughput = asyncio.run(run_benchmark_step(chunk_size_bytes, args.mode, args.coro_count, 0, args.time))
        run_values.append(throughput)
        print(f" Done. ({throughput:.2f} GB/s)")

    return run_values


def main():
    parser = argparse.ArgumentParser(description="AsyncMultiRangeDownloader Benchmark")
    parser.add_argument(
        "--mode", 
        choices=["sequential", "random"], 
        default="sequential", 
        help="Choose the read pattern: sequential (default) or random"
    )
    parser.add_argument(
        "--mb", 
        type=int,
        required=True,
        help="Size of each chunk read in MB (e.g., 1 for 1MB IO size)"
    )
    parser.add_argument(
        "--type", 
        choices=["zonal", "regional"], 
        default="zonal", 
        help="Choose the bucket type"
    )
    parser.add_argument(
        "--coro-count", 
        type=int, 
        default=4, 
        help="Number of concurrent asyncio tasks issuing read ranges"
    )
    parser.add_argument(
        "--time", 
        type=int, 
        default=30, 
        help="Duration of the recorded benchmark in seconds (default: 30)"
    )
    parser.add_argument(
        "--warmup", 
        type=int, 
        default=5, 
        help="Duration of the warmup period in seconds before counting bytes (default: 5)"
    )

    args = parser.parse_args()

    global BUCKET_NAME, OBJECT_NAME
    if args.type == 'zonal':
        BUCKET_NAME = "do-not-delete-margubur-zonal-usc1"
        OBJECT_NAME = "20gb-file"
    else:
        BUCKET_NAME = "rahman-bucket"
        OBJECT_NAME = "100gb-file"

    try:
        os.sched_setaffinity(0, {i for i in range(20, 180)})
    except AttributeError:
        pass 

    chunk_size_bytes = args.mb * 1024 * 1024

    print(f"Starting AsyncMultiRangeDownloader (gRPC Bidi-Stream) Benchmark")
    print(f"Read Mode: {args.mode.upper()}")
    print(f"Config: {args.time}s test time + {args.warmup}s network warmup | {NUM_RUNS} Runs | {args.coro_count} Tasks")
    print("-" * 80)

    run_values = async_main(args, chunk_size_bytes)
    
    best_result = max(run_values)
    print(f" >> Best for this config: {best_result:.2f} GB/s\n")

    print("\n" + "=" * 80)
    print(f"FINAL RESULTS TABLE ({args.mode.upper()} - Best of {NUM_RUNS} Runs)")
    print("=" * 80)
    
    header = f"{'CHUNK SIZE':<15} | {'ASYNC TASKS':<15} | THROUGHPUT"
    print(header)
    print("-" * 80)

    row = f"{args.mb}MB".ljust(15) + f" | {args.coro_count} TASKS".ljust(18) + f" | {best_result:.2f} GB/s"
    print(row)
    print("=" * 80)

if __name__ == "__main__":
    logging.basicConfig(level=logging.ERROR)
    main()
```
</details>

<details>
<summary>Numbers with loop.run_in_executor(None, master.close)</summary>

```
(temp) margubur@grpc-team-test:~/scripts$ python3 bench-direct.py --coro-count=32 --mb=1
Starting AsyncMultiRangeDownloader (gRPC Bidi-Stream) Benchmark
Read Mode: SEQUENTIAL
Config: 30s test time + 5s network warmup | 5 Runs | 32 Tasks
--------------------------------------------------------------------------------
Testing: 1 MB chunks
   Warmup run (establishing TCP connections)... Done. (Ignored: 1.26 GB/s)
   Run 1/5 (Benchmarking)... Done. (1.23 GB/s)
   Run 2/5 (Benchmarking)... Done. (1.68 GB/s)
   Run 3/5 (Benchmarking)... Done. (1.26 GB/s)
   Run 4/5 (Benchmarking)... Done. (1.43 GB/s)
   Run 5/5 (Benchmarking)... Done. (1.25 GB/s)
 >> Best for this config: 1.68 GB/s
```
</details>

<details>
<summary>Numbers for master.close</summary>

```
(temp) margubur@grpc-team-test:~/scripts$ python3 bench-direct.py --coro-count=32 --mb=1
Starting AsyncMultiRangeDownloader (gRPC Bidi-Stream) Benchmark
Read Mode: SEQUENTIAL
Config: 30s test time + 5s network warmup | 5 Runs | 32 Tasks
--------------------------------------------------------------------------------
Testing: 1 MB chunks
   Warmup run (establishing TCP connections)... Done. (Ignored: 2.11 GB/s)
   Run 1/5 (Benchmarking)... Done. (1.97 GB/s)
   Run 2/5 (Benchmarking)... Done. (2.01 GB/s)
   Run 3/5 (Benchmarking)... Done. (2.05 GB/s)
   Run 4/5 (Benchmarking)... Done. (2.11 GB/s)
```
</details>

gcsfs/zb_hns_utils.py
googlyrahman: Good catch! This would happen on error path, and it makes sense to keep the fast path out of try-catch.

googlyrahman:
>> Since gRPC typically returns payloads in 2MB chunks, any user operation requesting less than 2MB will only trigger a single write to the buffer.

> Could you elaborate on this? My understanding of gRPC in Python is that we join all chunks, deserialize the data via Protobuf, and then trigger the data_buffer write. The data_buffer write interacts with the Python Protobuf upb memory. How does the 2MB limit help?

Sure. I specifically want to optimize the scenario where a user reads the data in a single pass.

Under the hood, gRPC streams data back to the SDK in 2MB chunks. If a user requests exactly 2MB, the SDK fetches that chunk from gRPC and normally writes it into the buffer. The problem is that the buffer currently allocates a brand new 2MB block of memory just to copy the data over from the gRPC chunk.

That extra allocation and memory copy is completely unnecessary. Since gRPC already gives us a perfectly sized 2MB bytes object, we can optimise this by skipping the new allocation and just exposing those raw gRPC bytes directly to the user. This optimisation is necessary for <=2MB chunk sizes.

zhixiangli:
LGTM! Hi @martindurant, would you mind taking a look if you're interested?

martindurant:
I *think* this is the kind of thing I was suggesting when I mentioned "fast-path" the last time around :) 

First of all, I would change the PR title to indicate that this is a read operation, not write. I understand that to read, bytes might be written into an output buffer, but still.

> This PR also refactors the buffer class to keep its logic in one place.

Glad to see this. 


I don't expect to be able to look at this PR in detail over the next ~24hr.

googlyrahman:
> First of all, I would change the PR title to indicate that this is a read operation, not write. I understand that to read, bytes might be written into an output buffer, but still.

Updated the title.

gcsfs/extended_gcsfs.py
martindurant: @zhixiangli  : are you satisfied?

gcsfs/extended_gcsfs.py
```diff
@@ -461,18 +453,17 @@ async def _download(o, s, b, mrd_or_pool):
...
-                        raise
+            try:
+                master_buffer.close()
+            except Exception:
+                # If we are already handling a network/download exception,
+                # ignore the exception from buffer (which is just a symptom of the drop).
+                # If there's no download error, this means the buffer logic
+                # itself failed, so we must surface the error.
+                if not has_error:
+                    raise
```
martindurant, line 542: Are multiple errors possible?

gcsfs/zb_hns_utils.py
```diff
@@ -182,33 +183,47 @@ class DirectMemmoveBuffer:
...
+
+    Memory allocation is natively deferred. If the payload precisely aligns
+    with expected bounds sequentially, it gracefully overrides manual memmoves
+    using true Zero-Copy payload replacement safely under the hood.
+
+    Note: This class is now strictly Thread-Safe
     """
 
-    def __init__(self, start_address, end_address, executor, max_pending=5):
+    THRESHOLD_BYTES_FOR_SCHEDULING = 128 * 1024
```
martindurant, line 245: I am surprised that this is so small. We had a bigger threshold before?

gcsfs/zb_hns_utils.py
```diff
@@ -219,6 +234,74 @@ def __init__(self, start_address, end_address, executor, max_pending=5):
         self._done_event = threading.Event()
         self._done_event.set()
 
+    class PartialView:
```
martindurant: Why does this class need to be dynamically defined? That seems like it's asking for trouble. It doesn't seem to depend on any closure actually (only instantiation is `self.PartialView(self, offset, size)`).

gcsfs/zb_hns_utils.py
```diff
@@ -219,6 +234,74 @@ def __init__(self, start_address, end_address, executor, max_pending=5):
...
+                raise ValueError(f"Expected bytes, but got {type(data)}")
+            size = len(data)
+            with self._view_lock:
+                if self.current_offset + size > self.expected_size:
+                    error_msg = (
+                        f"Attempted to write {size} bytes "
+                        f"at offset {self.current_offset}. "
+                        f"Max capacity is {self.expected_size} bytes."
+                    )
+                    raise BufferError(error_msg)
```
martindurant:
```suggestion
            if self.current_offset + size > self.expected_size:
                error_msg = (
                    f"Attempted to write {size} bytes "
                    f"at offset {self.current_offset}. "
                    f"Max capacity is {self.expected_size} bytes."
                )
                raise BufferError(error_msg)
            with self._view_lock:
```

gcsfs/zb_hns_utils.py
```diff
@@ -219,6 +234,74 @@ def __init__(self, start_address, end_address, executor, max_pending=5):
...
+                abs_offset = self.start_offset + self.current_offset
+                self.current_offset += size
+
+            return self.parent._submit_write(abs_offset, data, size)
+
+        def close(self):
+            """
+            Validates boundaries enforcing complete local payload consistency.
+            """
+            with self._view_lock:
```
martindurant: Why do you want a lock here if nothing is being mutated?

gcsfs/zb_hns_utils.py
```diff
@@ -219,6 +234,74 @@ def __init__(self, start_address, end_address, executor, max_pending=5):
...
+
+    def get_view(self, offset, size):
+        """Constructs secure mapped offset references correctly handling constraint layouts."""
+        if offset < 0 or offset + size > self.expected_size:
+            raise ValueError("Invalid view requested: exceeds physical boundaries!")
+
+        start = offset
+        end = offset + size
+
+        with self._lock:
```
martindurant, line 292: As elsewhere, the checks should probably not live within the lock. 
Actually, `append()` is atomic, so I don't know why you need it at all.

gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,119 @@ def _decrement_pending(self):
...
-            raise e
+            with self._lock:
+                if self._stop_accepting_writes or self._is_closed:
+                    raise ValueError("I/O operation on closed buffer.")
+
+                if self._error:
+                    raise self._error
+
+                if self._result_bytes is None:
+                    if dest_offset == 0 and size == self.expected_size:
```
martindurant, line 339:
```suggestion
                    # fastpath: return buffer directly
                    if dest_offset == 0 and size == self.expected_size:
```

gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,119 @@ def _decrement_pending(self):
...
+                        self._result_bytes = data_bytes
+                        self.semaphore.release()  # Release because we skip the executor
+                        fut = concurrent.futures.Future()
+                        fut.set_result(None)
```
martindurant, line 344: Returning this future is pretty weird. I suppose it's essentially an Event in this use.

gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,119 @@ def _decrement_pending(self):
...
+                        self._total_bytes_written += size
+                        return fut
+                    else:
+                        self._result_bytes = PyBytes_FromStringAndSize(
+                            None, self.expected_size
+                        )
+                        self._start_address = PyBytes_AsString(self._result_bytes)
```
martindurant:
```suggestion
                    self._result_bytes = PyBytes_FromStringAndSize(
                        None, self.expected_size
                    )
                    self._start_address = PyBytes_AsString(self._result_bytes)
```

gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,119 @@ def _decrement_pending(self):
...
+
+                # Defensive programming: gracefully catch internal overwrite attempts
+                if self._start_address is None:
+                    raise BufferError(
+                        "Attempted to execute standard write over a Zero-Copied payload."
+                    )
+
+                dest = self._start_address + dest_offset
+                if self._pending_count == 0:
+                    self._done_event.clear()
```
martindurant, line 360: Race possible?

gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,119 @@ def _decrement_pending(self):
...
+                self._pending_count += 1
+
+        except BaseException:
+            self.semaphore.release()
+            raise
```
martindurant, line 365:
```suggestion
        finally:
            self.semaphore.release()
```

gcsfs/zb_hns_utils.py
```diff
@@ -227,86 +310,119 @@ def _decrement_pending(self):
...
+            # Fast path, no need to send it to executor
+            try:
+                self._do_memmove(dest, data_bytes, size)
+            except BaseException:
+                # The exception is already captured in self._error by _do_memmove
+                pass
+
+            fut = concurrent.futures.Future()
+            with self._lock:
+                local_err = self._error
```
martindurant: Again, not sure why the lock is needed for an attribute read

martindurant (commented):
I see lots of uses of the (threading) lock for read-only situations, would appreciate reconsidering if this is necessary. Python does not deal well with locks as the number of threads increases.

Is it actually better to _not_ have the buffer/views not be threadsafe: do we have a workflow where they are called from multiple threads?

gcsfs/zb_hns_utils.py
googlyrahman: I can't use finally here. The semaphore limits pending tasks and is released by the executor task upon completion. We only want to release it here if the executor submission itself fails.

gcsfs/zb_hns_utils.py
googlyrahman: I don't think a race is possible here, this entire block is protected by self._lock, let me know otherwise

gcsfs/zb_hns_utils.py
googlyrahman: Returning a pre-completed Future keeps the return type consistent for the fast-path so the caller doesn't have to handle multiple return types.

gcsfs/zb_hns_utils.py
googlyrahman: Added the comment

gcsfs/zb_hns_utils.py
googlyrahman: While append() itself is atomic under the GIL, we are doing a check and act operation here. We have to iterate through the existing intervals to check for overlaps before appending. If we drop the lock, two threads could pass the overlap check at the exact same time and then both append overlapping views. We need to keep the lock to prevent that.

gcsfs/zb_hns_utils.py
googlyrahman: Hmmm, 128KB appears to be switch point. I think it depends whether you use it synchronously or asynchronously. For asynchronous execution 128KB is the switch point. However for synchronous, we require a bigger threshold like 512 KB or 1MB so that threading doesn't become an overhead.

Since our use case is asynchronous, i wrote a benchmarking script similar to our current use case. There 128KB is the switch point irrespective of the number of tasks.

Let me know if i miss anything here, or the benchmarking does not represent our usecase.

<details>
<summary>The Script</summary>

```
import ctypes
import threading
import concurrent.futures
import time
import asyncio

# --- System Under Test Parameters ---
NUM_BUFFERS = 16                # 'X' buffers to create
CONCURRENCY_PER_BUFFER = 4     # 4 asyncio tasks per buffer
SHARED_WORKERS = 8             # Shared thread pool size across ALL buffers

# -------------------------------------------------------------------------
# C-API Mocks 
# -------------------------------------------------------------------------
ctypes.pythonapi.PyBytes_FromStringAndSize.restype = ctypes.py_object
ctypes.pythonapi.PyBytes_FromStringAndSize.argtypes = [ctypes.c_void_p, ctypes.c_ssize_t]
ctypes.pythonapi.PyBytes_AsString.restype = ctypes.c_void_p
ctypes.pythonapi.PyBytes_AsString.argtypes = [ctypes.py_object]

def PyBytes_FromStringAndSize(dummy, size):
    return ctypes.pythonapi.PyBytes_FromStringAndSize(None, size)

def PyBytes_AsString(obj):
    return ctypes.pythonapi.PyBytes_AsString(obj)

# -------------------------------------------------------------------------
# Buffer Prototype 
# -------------------------------------------------------------------------
class PartialView:
    def __init__(self, parent, start_offset, expected_size):
        self.parent = parent
        self.start_offset = start_offset
        self.expected_size = expected_size
        self.current_offset = 0
        self._view_lock = threading.Lock()

    def write(self, data):
        size = len(data)
        with self._view_lock:
            if self.current_offset + size > self.expected_size:
                raise BufferError("Exceeded local view capacity.")
            abs_offset = self.start_offset + self.current_offset
            self.current_offset += size
        return self.parent._submit_write(abs_offset, data, size)

    def close(self):
        if self.current_offset < self.expected_size:
            raise BufferError("Buffer incomplete.")

class DirectMemmoveBuffer:
    THRESHOLD_BYTES_FOR_SCHEDULING = 512 * 1024  

    def __init__(self, expected_size, executor, max_pending=5):
        self.expected_size = expected_size
        self.executor = executor
        self._pending_count = 0
        self._error = None
        self._total_bytes_written = 0
        self._stop_accepting_writes = False
        self._is_closed = False
        self._allocated_intervals = []
        self._result_bytes = None
        self._start_address = None
        
        self.semaphore = threading.Semaphore(max_pending)
        self._lock = threading.Lock()
        self._done_event = threading.Event()
        self._done_event.set()

    def get_view(self, offset, size):
        start, end = offset, offset + size
        with self._lock:
            if self._stop_accepting_writes or self._is_closed:
                raise ValueError("Buffer closed.")
            for a_start, a_end in self._allocated_intervals:
                if max(start, a_start) < min(end, a_end):
                    raise ValueError("Overlapping view.")
            self._allocated_intervals.append((start, end))
        return PartialView(self, offset, size)

    def _decrement_pending(self):
        self.semaphore.release()
        with self._lock:
            self._pending_count -= 1
            if self._pending_count == 0:
                self._done_event.set()

    def _submit_write(self, dest_offset, data_bytes, size):
        self.semaphore.acquire()
        try:
            with self._lock:
                if self._stop_accepting_writes or self._is_closed:
                    raise ValueError("Buffer closed.")
                if self._error:
                    raise self._error
                if self._result_bytes is None:
                    self._result_bytes = PyBytes_FromStringAndSize(None, self.expected_size)
                    self._start_address = PyBytes_AsString(self._result_bytes)
                dest = self._start_address + dest_offset
                if self._pending_count == 0:
                    self._done_event.clear()
                self._pending_count += 1
        except BaseException:
            self.semaphore.release()
            raise

        if size <= self.THRESHOLD_BYTES_FOR_SCHEDULING:
            try:
                self._do_memmove(dest, data_bytes, size)
            except BaseException:
                pass
            fut = concurrent.futures.Future()
            if self._error:
                fut.set_exception(self._error)
            else:
                fut.set_result(None)
            return fut
        else:
            try:
                return self.executor.submit(self._do_memmove, dest, data_bytes, size)
            except BaseException as e:
                with self._lock:
                    self._error = e
                self._decrement_pending()
                raise e

    def _do_memmove(self, dest, data_bytes, size):
        try:
            with self._lock:
                if self._error:
                    return
            ctypes.memmove(dest, data_bytes, size)
            with self._lock:
                self._total_bytes_written += size
        except BaseException as e:
            with self._lock:
                if self._error is None:
                    self._error = e
            raise
        finally:
            self._decrement_pending()

    def close(self):
        with self._lock:
            self._stop_accepting_writes = True
        
        self._done_event.wait()
        
        with self._lock:
            self._is_closed = True
            if self._error:
                raise self._error


# -------------------------------------------------------------------------
# Multi-Buffer Asyncio Benchmark Logic
# -------------------------------------------------------------------------
async def async_producer_worker(buffer, payload, chunk_size, iterations, start_offset):
    """An asyncio task that pushes chunks to a specific buffer."""
    loop = asyncio.get_running_loop()
    offset = start_offset
    view = buffer.get_view(offset, chunk_size * iterations)
    
    for _ in range(iterations):
        view.write(payload)
        
    view.close()


async def execute_async_test_run(chunk_size, is_direct):
    """
    Sets up X buffers, sharing 1 Executor of 8 workers.
    Spawns 4 async tasks per buffer.
    """
    total_chunks_per_buffer = 10000
    chunks_per_producer = total_chunks_per_buffer // CONCURRENCY_PER_BUFFER
    
    total_bytes = total_chunks_per_buffer * chunk_size
    payload = b"x" * chunk_size

    if is_direct:
        DirectMemmoveBuffer.THRESHOLD_BYTES_FOR_SCHEDULING = chunk_size + 1
    else:
        DirectMemmoveBuffer.THRESHOLD_BYTES_FOR_SCHEDULING = -1 

    # Create the SINGLE shared executor for this test run
    with concurrent.futures.ThreadPoolExecutor(max_workers=SHARED_WORKERS) as shared_executor:
        buffers = []
        tasks = []
        
        # 1. Initialize all X buffers
        for _ in range(NUM_BUFFERS):
            buffer = DirectMemmoveBuffer(
                expected_size=total_bytes, 
                executor=shared_executor, 
                max_pending=SHARED_WORKERS  # Tie semaphore backpressure to pool size
            )
            buffers.append(buffer)
            
        # 2. Spawn 4 producers for EACH buffer
        for buffer in buffers:
            for i in range(CONCURRENCY_PER_BUFFER):
                start_offset = i * (chunks_per_producer * chunk_size)
                tasks.append(
                    asyncio.create_task(
                        async_producer_worker(buffer, payload, chunk_size, chunks_per_producer, start_offset)
                    )
                )

        start_time = time.perf_counter()
        
        # 3. Await all (X * 4) async producers
        await asyncio.gather(*tasks)
        
        # 4. Wait for all buffers to drain their background queues
        loop = asyncio.get_running_loop()
        close_tasks = [
            loop.run_in_executor(None, b.close) for b in buffers
        ]
        await asyncio.gather(*close_tasks)
        
        end_time = time.perf_counter()
        
        return end_time - start_time


def run_targeted_threshold_test():
    candidate_sizes_x = [
        32 * 1024,       # 32 KB
        128 * 1024,      # 128 KB
        256 * 1024,      # 256 KB
        512 * 1024,      # 512 KB
        1024 * 1024,     # 1 MB
    ]

    total_tasks = NUM_BUFFERS * CONCURRENCY_PER_BUFFER

    print(f"Running Asyncio Crossover Test (4000 chunks per buffer)")
    print(f"Architecture: {NUM_BUFFERS} Buffers | {CONCURRENCY_PER_BUFFER} Tasks/Buffer ({total_tasks} Total Tasks) | {SHARED_WORKERS} Shared Threads")
    print(f"{'Chunk Size (X)':<15} | {'Direct Time (s)':<18} | {'Threaded Time (s)':<18} | {'Winner & Gain'}")
    print("-" * 80)

    for chunk_size in candidate_sizes_x:
        direct_time = asyncio.run(execute_async_test_run(chunk_size, is_direct=True))
        threaded_time = asyncio.run(execute_async_test_run(chunk_size, is_direct=False))
        
        if direct_time < threaded_time:
            winner = "Direct"
            gain_pct = ((threaded_time - direct_time) / threaded_time) * 100
            gain_str = f"+{gain_pct:.1f}% faster"
        else:
            winner = "Threaded"
            gain_pct = ((direct_time - threaded_time) / direct_time) * 100
            gain_str = f"+{gain_pct:.1f}% faster"
            
        label = f"{chunk_size / 1024:.0f} KB"
        print(f"{label:<15} | {direct_time:<18.4f} | {threaded_time:<18.4f} | {winner} {gain_str}")

if __name__ == "__main__":
    run_targeted_threshold_test()
```

</details>

<details>
<summary>The Numbers</summary>

```
(temp) margubur@grpc-team-test:~/scripts/rahman$ python3 bench-buffer.py
Running Asyncio Crossover Test (4000 chunks per buffer)
Architecture: 1 Buffers | 4 Tasks/Buffer (4 Total Tasks) | 8 Shared Threads
Chunk Size (X)  | Direct Time (s)    | Threaded Time (s)  | Winner & Gain
--------------------------------------------------------------------------------
32 KB           | 0.1184             | 0.1772             | Direct +33.2% faster
128 KB          | 0.2941             | 0.2374             | Threaded +19.3% faster
256 KB          | 0.5543             | 0.2864             | Threaded +48.3% faster
512 KB          | 1.0982             | 0.3963             | Threaded +63.9% faster
1024 KB         | 2.2215             | 0.5787             | Threaded +73.9% faster
(temp) margubur@grpc-team-test:~/scripts/rahman$ python3 bench-buffer.py
Running Asyncio Crossover Test (4000 chunks per buffer)
Architecture: 1 Buffers | 4 Tasks/Buffer (4 Total Tasks) | 8 Shared Threads
Chunk Size (X)  | Direct Time (s)    | Threaded Time (s)  | Winner & Gain
--------------------------------------------------------------------------------
32 KB           | 0.1154             | 0.1821             | Direct +36.7% faster
128 KB          | 0.2937             | 0.2342             | Threaded +20.3% faster
256 KB          | 0.5490             | 0.2842             | Threaded +48.2% faster
512 KB          | 1.0895             | 0.3980             | Threaded +63.5% faster
1024 KB         | 2.2206             | 0.5976             | Threaded +73.1% faster
(temp) margubur@grpc-team-test:~/scripts/rahman$ python3 bench-buffer.py
Running Asyncio Crossover Test (4000 chunks per buffer)
Architecture: 8 Buffers | 4 Tasks/Buffer (32 Total Tasks) | 8 Shared Threads
Chunk Size (X)  | Direct Time (s)    | Threaded Time (s)  | Winner & Gain
--------------------------------------------------------------------------------
32 KB           | 0.9193             | 1.4610             | Direct +37.1% faster
128 KB          | 2.4230             | 1.9067             | Threaded +21.3% faster
256 KB          | 4.5067             | 2.3277             | Threaded +48.4% faster
512 KB          | 8.8955             | 3.1789             | Threaded +64.3% faster
1024 KB         | 17.8854            | 4.5165             | Threaded +74.7% faster
```
</details>

gcsfs/extended_gcsfs.py
googlyrahman: Yeah, That’s especially true during the closing process. If the background workers fail to write background writes to the buffer, they throw an error upon closing.

We should only print these errors if there were no download failures. If both a network error and a buffer error occur, the network error should take priority, followed by the buffer error.

gcsfs/zb_hns_utils.py
googlyrahman: You're right, we don't need a lock here. I've removed the lock.

gcsfs/zb_hns_utils.py
googlyrahman: Done

gcsfs/zb_hns_utils.py
googlyrahman: It's just a nested class, not a dynamic closure, but you're right that it doesn't need to be here. I've moved it to the module level for readabiliy.

gcsfs/zb_hns_utils.py
googlyrahman: This is check-then-act sequence, can't move the condition out of the lock here. Imagine two threads trying to write 10 bytes when you only have 10 bytes left to fill the buffer. If i keep the check outside, both of them would pass this check, and end up writing 20 bytes.

gcsfs/zb_hns_utils.py
googlyrahman: You're right, we don't need the lock here. I removed the lock acquisition on this path.

googlyrahman:
Tried addressing your comments @martindurant, Please take a look again, and let me know if you've any concerns.

Regarding:
> I see lots of uses of the (threading) lock for read-only situations, would appreciate reconsidering if this is necessary. Python does not deal well with locks as the number of threads increases.

I've removed few locks on read-only situations wherever possible based on your comments.

> Is it actually better to not have the buffer/views not be threadsafe: do we have a workflow where they are called from multiple threads?

Our goal is to eventually move this buffer to the SDK, where it will be used in both synchronous [For standard buckets] and asynchronous contexts [For zonal buckets]. Keeping it thread-safe would mainly help PythonSDK in synchronous context where they would use multithreading to achieve gcs reads of standard bucket.

gcsfs/zb_hns_utils.py
```diff
@@ -180,37 +181,95 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+    """A bounded memory writer providing robust overfill/underfill constraint validations."""
+
+    def __init__(self, parent, start_offset, expected_size):
+        self.parent = parent
+        self.start_offset = start_offset
+        self.expected_size = expected_size
+        self.current_offset = 0
+        self._view_lock = threading.Lock()
+
+    def write(self, data):
```
zhixiangli, line 194: When performing a gRPC stream-based download via MRD, the server can emit a final response chunk that is empty (content=b"") with range_end=True to signal completion of the range.

If the zero-copy fast path is taken, the first write directly returns the payload and leaves self._start_address set to None. The subsequent empty write (b"") bypasses the PartialView.write capacity check (since size=0) and calls _submit_write. This hits the defensive guard and raises a BufferError: Attempted to execute standard write over a Zero-Copied payload., causing the valid read to crash.

gcsfs/zb_hns_utils.py
```diff
@@ -180,37 +181,95 @@ async def close_aaow(aaow, finalize_on_close=False):
...
+        """
+        Schedules a write operation to memory mapping.
+        """
+        if not isinstance(data, bytes):
+            raise ValueError(f"Expected bytes, but got {type(data)}")
```
zhixiangli, line 199: It was `data_bytes = bytes(data) if not isinstance(data, bytes) else data` that supports bytearray /memoryview

googlyrahman: It is an intentional change.

gcsfs/zb_hns_utils.py
googlyrahman: This was quite an edge case, specifically when user writes empty bytes on zero copied payloads.

Good catch Zhixiang! I'll let write ignore empty bytes request. We've check in place inside get_value which returns an empty bytes if the address is none.

zhixiangli (approved):

PR #841, opened 2026-05-11
feat: Allow running E2E tests against fsspec HEAD

Mahalaxmibejugam:
This PR adds the capability to run GCSFS end-to-end integration tests against the latest development head of fsspec. This enables setting up a nightly or scheduled pipeline to catch breaking changes in fsspec before they are released.

Added _INSTALL_FSSPEC_HEAD substitution variable (defaults to "false") and passed it to the remote VM setup script. This variable would be set to True from the nightly run trigger. If set to "true", it installs fsspec directly from its git repository using --force-reinstall to ensure the latest head is fetched. Also added a substitution variable to give flexibility to run against GCSFS latest released version or HEAD.

Currently, GCSFS tests run in every fsspec PR but are limited to the emulator and do not cover end-to-end testing against real GCS. Implementing this setup for nightly runs would help detect any fsspec changes that might break GCSFS.

### DIFF
--- a/cloudbuild/e2e-tests-cloudbuild.yaml
+++ b/cloudbuild/e2e-tests-cloudbuild.yaml
@@ -18,6 +18,8 @@ substitutions:
   _SHORT_BUILD_ID: ${BUILD_ID:0:8}
   _SKIP_IF_FAILED: '[[ -f /workspace/FAILED ]] && echo "⚠️ Skipping: Previous failure" && exit 0'
   _TEST_ENV_VARS: "SHORT_BUILD_ID='${_SHORT_BUILD_ID}' PROJECT_ID='${PROJECT_ID}' REGION='${_REGION}' KEY_RING='${_GCSFS_KEY_RING_NAME}' KEY_NAME='${_GCSFS_KEY_NAME}'"
+  _INSTALL_FSSPEC_HEAD: "false"
+  _INSTALL_GCSFS_FROM_PYPI: "false"
 
 steps:
   # Generate a persistent SSH key for this build run.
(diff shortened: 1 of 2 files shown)


martindurant:
This is generally a good idea.
Will a failure send an alert somewhere?

zhixiangli:
Hi! I'm thinking of creating a list of dependencies that we should test nightly, starting with fsspec. I was wondering if we could make this configuration more flexible—perhaps by adding a single toggle to manage fsspec and other dependencies collectively?

Mahalaxmibejugam:
> Will a failure send an alert somewhere?

Yes, I will be setting up an alert based on the build status. It can be received on email or chat.

Mahalaxmibejugam:
> I'm thinking of creating a list of dependencies that we should test nightly, starting with fsspec. I was wondering if we could make this configuration more flexible—perhaps by adding a single toggle to manage fsspec and other dependencies collectively?

Using a single toggle to fetch multiple dependencies from their HEAD may make it difficult to identify which specific dependency is causing a failure. What other dependencies are you planning to include in nightly run?

zhixiangli:
> > I'm thinking of creating a list of dependencies that we should test nightly, starting with fsspec. I was wondering if we could make this configuration more flexible—perhaps by adding a single toggle to manage fsspec and other dependencies collectively?
> 
> Using a single toggle to fetch multiple dependencies from their HEAD may make it difficult to identify which specific dependency is causing a failure. What other dependencies are you planning to include in nightly run?

Ideally, all direct dependencies—such as aiohttp—should have version exclusions configured. My guess is that there are compatibility issues we should detect early. Defer to your decision on which libraries to include, as they generally present similar problems.

```
dependencies = [
    "aiohttp!=4.0.0a0,!=4.0.0a1,>=3.9.0",
    "decorator>4.1.2",
    "fsspec>=2026.3.0",
    "google-auth>=1.2",
    "google-auth-oauthlib",
    "google-cloud-storage>=3.9.0",
    "google-cloud-storage-control",
    "requests",
]
```

Mahalaxmibejugam:
> Ideally, all direct dependencies—such as aiohttp—should have version exclusions configured

In fsspec PRs, we are only running the emulator based runs and don't have e2e tests running against real GCS. We are planning for these nightly runs to fill this gap. So even if some breaking change is introduced in fsspec, we won't block the PRs from getting merged but we will be able to know the issue early. So if a release is planned, we will be able to delay it until the fixes are done.

IMO adding other dependencies to this nightly run won't add much value but would be a noise if there are frequent failures due to multiple dependencies being used from HEAD

zhixiangli (approved):

PR #842, opened 2026-05-11
Adds Pipe function in microbenchmark

ankitaluthra1:
This pull request introduces microbenchmarks for the pipe() operation in gcsfs

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
+++ b/gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
@@ -64,10 +64,11 @@ def _process_worker(
     process_durations_shared,
     index,
 ):
+    """A worker function for each process to pipe files concurrently."""
+
     # Generate data buffer efficiently per process
     data_buffer = b"0" * file_size
 
-    """A worker function for each process to pipe files concurrently."""
     start_time = time.perf_counter()
 
     with ThreadPoolExecutor(max_workers=threads) as executor:
--- a/gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
+++ b/gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import time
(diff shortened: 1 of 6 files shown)


gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
```diff
@@ -0,0 +1,126 @@
...
+    file_size,
+    chunk_size,
+    threads,
+    process_durations_shared,
+    index,
+):
+    """A worker function for each process to pipe files concurrently."""
+
+    # Generate data buffer efficiently per process
+    data_buffer = b"0" * file_size
```
zhixiangli: nit: could we use a random string instead? Since "0" * file_size is trivial to compress, I'm concerned that compression might impact performance, especially if it’s implemented in a current or future hot path.

gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
```diff
@@ -0,0 +1,126 @@
...
+    "gcsfs_benchmark_pipe",
+    single_threaded_cases,
+    indirect=True,
+    ids=lambda p: p.name,
+)
+def test_pipe_single_threaded(benchmark, gcsfs_benchmark_pipe, monitor):
+    gcs, file_paths, params = gcsfs_benchmark_pipe
+
+    # Generate data buffer once efficiently outside the timed benchmark execution
+    data_buffer = b"0" * params.file_size_bytes
```
zhixiangli: nit: use a random string.

zhixiangli (approved):

gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
ankitaluthra1: updated

gcsfs/tests/perf/microbenchmarks/pipe/test_pipe.py
ankitaluthra1: updated

PR #843, opened 2026-05-12
Improve test speed and emulator compatibility

Yonghui-Lee:
This PR improves test execution speed and reliability. It addresses slow credential discovery timeouts, optimizes test file setup, and reduces repeated failing lookups on the emulator.

### Cache Unknown Bucket Type In Tests

- Added a cache_unknown_buckets option to ExtendedGcsFileSystem to allow caching of bucket types even when they are returned as UNKNOWN (which happens with the emulator). This avoids repeated slow lookups.

### ADC Timeout Prevention

- Added an autouse fixture _avoid_adc_timeout to set NO_GCE_CHECK=true and a default GOOGLE_CLOUD_PROJECT="dummy-project" when not running against the real GCS. This prevents slow timeouts attempting to reach the GCE metadata server in environments without GCP access.

### Isolated Multi-threaded Test File

- Removed multi_threaded_test_file from the global all files dictionary. Created a dedicated multi_threaded_test_file fixture to lazily create this file only for the tests that actually require it.


Mahalaxmibejugam:
Thank you for these improvements, these fixes appear to be generic for all test steps. They alone will not resolve the high execution time for the "Run Standard Bucket Tests with default ON extended feature support" step. 

In GitHub CI, we run standard bucket tests with the experimental value set to both True and False. The step with the value set to False takes two minutes, whereas the step with it set to True can take up to 20 minutes in some instances. We need to verify if this discrepancy is due to the additional storage layout call(which is not supported in emulator).

The failure persists even after these changes, as the tests didn't complete within 30 minutes.

Yonghui-Lee:
@Mahalaxmibejugam Thanks for your insight. 

"Run Standard Bucket Tests with default ON extended feature support" was slow for a long time. So I suspect there is some generic issues introduced during recent PRs which cause the test time to exceed 30 mins. But the improvements didn't solve our problem.

Then I spent more time investing the "Run Standard Bucket Tests with default ON extended feature support" specially. As you suspect, the additional storage layout calls significantly increase the test time. I will address it in this PR.

zhixiangli:
As discussed, apart from the root cause, I wish we could introduce parallel runs that significantly reduce the test execution time, maybe a separate PR, as we have very powerful machines, so it should be achievable.

martindurant:
> I wish we could introduce parallel runs 

Would be reasonable, but I also am interested in why they are slow in the first place.

> we have very powerful machines

GHA CI doesn't, thought, and for the future we don't want to assume that local testers do either. (This is assuming we want to/can test using an emulator in the long run)

gcsfs/tests/conftest.py
```diff
@@ -438,7 +456,10 @@ def zonal_write_mocks():
         return
 
     patch_target_get_bucket_type = (
```
zhixiangli, line 460: nit: `patch_target_lookup_bucket_type`

gcsfs/tests/conftest.py
```diff
@@ -82,6 +75,24 @@
 }
 
 
+@pytest.fixture(autouse=True)
+def _avoid_adc_timeout(monkeypatch):
+    """Avoid slow ADC lookups and Metadata Server requests in tests."""
+    # Do not apply if tests are explicitly running against real GCS
+    if os.environ.get("STORAGE_EMULATOR_HOST") == "https://storage.googleapis.com":
```
zhixiangli: nit: could we use a function to determine if it is a valid GCS?

gcsfs/extended_gcsfs.py
```diff
@@ -104,6 +104,7 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
             - retry_multiplier: Multiplier for delay between retries.
             These map to `google.api_core.retry.AsyncRetry` arguments (without 'retry_' prefix).
         """
+        self._cache_unknown_buckets = kwargs.pop("cache_unknown_buckets", False)
```
zhixiangli: Generally, it's better to avoid exposing new fields or configuration options in production code solely for testing use cases.

We can achieve the same result entirely within the test suite by mocking _get_bucket_type in conftest.py to return UNKNOWN instantly when running against the emulator. This keeps the production code clean.

zhixiangli (changes requested):

gcsfs/extended_gcsfs.py
Yonghui-Lee: I'm considering it will also be helpful in production. For example, a user interacts with standard GCS buckets but he does not have the storage.objects.list permission which is required to call the Storage Control API.

gcsfs/tests/conftest.py
Yonghui-Lee: Done

gcsfs/tests/conftest.py
Yonghui-Lee: Done

gcsfs/extended_gcsfs.py
zhixiangli: That's a good point; we should improve the caching logic, but let's address that in a separate PR and keep this one focused on improving the tests. Once we expose the new parameter publicly, we must maintain it, as removing it would be a breaking change.

Yonghui-Lee: Okay. I remove it from this PR and raise https://github.com/fsspec/gcsfs/pull/844

PR #844, opened 2026-05-13
Document intentional non-caching and fallback behavior for UNKNOWN bucket layouts

Yonghui-Lee:


gcsfs/extended_gcsfs.py
```diff
@@ -104,6 +107,7 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
             - retry_multiplier: Multiplier for delay between retries.
             These map to `google.api_core.retry.AsyncRetry` arguments (without 'retry_' prefix).
         """
+        self._cache_unknown_buckets = kwargs.pop("cache_unknown_buckets", False)
```
zhixiangli: If a user sets `_cache_unknown_buckets` to true and has the required permissions, is the result still cached when a transient error occurs? Could we check if it is a permission error / cache with a TTL, allowing for access to be granted later? WDYT?

ankitaluthra1:
Not caching UNKNOWN type is intentional, since the storage_lookup api should work on every bucket. In case bucket type is UNKNOWN it surely is a transient error and should not be cached. I think we should revert caching UNKNOWN in this PR and can update the documentation instead so its clearer that its an intentional change

gcsfs/extended_gcsfs.py
```diff
@@ -96,6 +96,9 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
         finalize_on_close : bool, default False
             By default, files in zonal buckets are left unfinalized to allow appends.
         **kwargs : dict
+            - cache_unknown_buckets : bool, default False
+                Whether to cache UNKNOWN bucket types. Useful when users lack permissions
```
ankitaluthra1: This use case is very rare. storage layout api needs minimal object permissions
https://docs.cloud.google.com/storage/docs/getting-storage-layout#expandable-1

Adding this flag might not be useful as in most cases UNKNOWN might come due to transient api call failure which should not be cached

ankitaluthra1 (changes requested):

zhixiangli (approved):

PR #845, opened 2026-05-13
Run tests concurrently in CI

Yonghui-Lee:
This PR enables concurrent test execution in CI using `pytest-xdist` to speed up the build process. It also addresses several test failures that were exposed by running against the emulator in a parallel environment.

### CI Infrastructure & Parallelization
- **Enabled `pytest-xdist`**: Added `-n auto` to `pytest` calls in `.github/workflows/ci.yml` and added `pytest-xdist` to `environment_gcsfs.yaml` and `pyproject.toml`.
- **Isolated Worker Buckets**: Modified `gcsfs/tests/settings.py` to append the xdist worker ID (e.g., `_gw0`) to bucket names. This ensures that parallel workers do not conflict by operating on the same buckets.
- **Global Emulator Setup**: Moved the GCS emulator startup to a dedicated step in `ci.yml` and set `STORAGE_EMULATOR_HOST` globally. This ensures all workers share a single emulator instance rather than trying to spin up conflicting docker containers.

### Test Fixes & Cleanups
- **Skipped `test_sign` on Emulator**: Skipped this test entirely when running against the emulator. The emulator does not support signed URL validation.
- **Skipped `test_credentials_from_raw_token` on Emulator**: Skipped this test on the emulator because the emulator accepts any token, causing the test to fail when it expects an invalid credentials error.
- **Consolidated Dummy Credentials**: Deleted `gcsfs/tests/fake-secret.json` which was malformed (invalid PEM format) and caused `DefaultCredentialsError` in extended tests. Updated CI to use `fake-service-account-credentials.json` which contains a validly formatted dummy key.

### DIFF
--- a/gcsfs/tests/settings.py
+++ b/gcsfs/tests/settings.py
@@ -1,20 +1,24 @@
 import os
 
-worker_id = os.environ.get("PYTEST_XDIST_WORKER")
-suffix = f"_{worker_id}" if worker_id else ""
 
-TEST_BUCKET = os.getenv("GCSFS_TEST_BUCKET", "gcsfs_test") + suffix
-TEST_VERSIONED_BUCKET = (
-    os.getenv("GCSFS_TEST_VERSIONED_BUCKET", "gcsfs_test_versioned") + suffix
+def _get_bucket_name(env_var: str, default_name: str) -> str:
+    worker_id = os.environ.get("PYTEST_XDIST_WORKER")
+    suffix = f"_{worker_id}" if worker_id else ""
+    return os.getenv(env_var, default_name) + suffix
+
+
+TEST_BUCKET = _get_bucket_name("GCSFS_TEST_BUCKET", "gcsfs_test")
+TEST_VERSIONED_BUCKET = _get_bucket_name(
+    "GCSFS_TEST_VERSIONED_BUCKET", "gcsfs_test_versioned"
 )
-TEST_HNS_BUCKET = os.getenv("GCSFS_HNS_TEST_BUCKET", "gcsfs_hns_test") + suffix
-TEST_ZONAL_BUCKET = os.getenv("GCSFS_ZONAL_TEST_BUCKET", "gcsfs_zonal_test") + suffix
+TEST_HNS_BUCKET = _get_bucket_name("GCSFS_HNS_TEST_BUCKET", "gcsfs_hns_test")
+TEST_ZONAL_BUCKET = _get_bucket_name("GCSFS_ZONAL_TEST_BUCKET", "gcsfs_zonal_test")
 TEST_PROJECT = os.getenv("GCSFS_TEST_PROJECT", "project")
-TEST_REQUESTER_PAYS_BUCKET = (
-    os.getenv("GCSFS_TEST_REQ_PAYS_BUCKET", "gcsfs_test_req_pays") + suffix
+TEST_REQUESTER_PAYS_BUCKET = _get_bucket_name(
+    "GCSFS_TEST_REQ_PAYS_BUCKET", "gcsfs_test_req_pays"
 )
-TEST_HNS_REQUESTER_PAYS_BUCKET = (
-    os.getenv("GCSFS_HNS_TEST_REQ_PAYS_BUCKET", "gcsfs_hns_test_req_pays") + suffix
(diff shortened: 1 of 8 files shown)


gcsfs/tests/test_credentials.py
```diff
@@ -85,8 +85,10 @@ def test_connect_cloud_not_on_google():
 @pytest.mark.parametrize("token", ["", "incorrect.token", "x" * 100])
 def test_credentials_from_raw_token(token):
     with patch.dict(os.environ, {"FETCH_RAW_TOKEN_EXPIRY": "false"}):
+        fs = GCSFileSystem(project="myproject", token=token)
+        if not fs.on_google:
+            pytest.skip("Emulator does not validate tokens")
         with pytest.raises(HttpError, match="Invalid Credentials"):
-            fs = GCSFileSystem(project="myproject", token=token)
```
zhixiangli, line 89: > the emulator accepts any token, causing the test to fail when it expects an invalid credentials error

So, was this test failing? The CIs seemed to pass in the past.

gcsfs/tests/settings.py
```diff
@@ -1,15 +1,20 @@
...
 TEST_PROJECT = os.getenv("GCSFS_TEST_PROJECT", "project")
-TEST_REQUESTER_PAYS_BUCKET = os.getenv(
-    "GCSFS_TEST_REQ_PAYS_BUCKET", "gcsfs_test_req_pays"
+TEST_REQUESTER_PAYS_BUCKET = (
+    os.getenv("GCSFS_TEST_REQ_PAYS_BUCKET", "gcsfs_test_req_pays") + suffix
 )
-TEST_HNS_REQUESTER_PAYS_BUCKET = os.getenv(
-    "GCSFS_HNS_TEST_REQ_PAYS_BUCKET", "gcsfs_hns_test_req_pays"
+TEST_HNS_REQUESTER_PAYS_BUCKET = (
+    os.getenv("GCSFS_HNS_TEST_REQ_PAYS_BUCKET", "gcsfs_hns_test_req_pays") + suffix
```
zhixiangli: nit: consider using a function to generate the bucket name?

gcsfs/tests/test_core.py
```diff
@@ -2399,23 +2399,17 @@ def test_find_maxdepth(gcs):
 
 
 def test_sign(gcs, monkeypatch):
+    if not gcs.on_google:
+        pytest.skip("Emulator does not support signing")
```
zhixiangli, line 2403: Could you explain why this fix is needed? The CIs passed previously.

gcsfs/tests/test_credentials.py
Yonghui-Lee: *   **Before the last commit:** The GCS emulator was started programmatically inside the `docker_gcs` fixture in `conftest.py`, and its URL was passed to tests via the `gcs_factory` fixture.
*   `test_credentials_from_raw_token` does **not** use the `gcs` fixture or `gcs_factory`. It instantiates `GCSFileSystem` directly: `fs = GCSFileSystem(project="myproject", token=token)`.
*   Because `STORAGE_EMULATOR_HOST` was not set in the environment before, this `fs` instance was actually trying to hit the **real GCS API**. Since the tokens were invalid, real GCS rejected them with `HttpError: Invalid Credentials`, and the test passed.
*   **In the last commit:** To support parallel testing with `pytest-xdist`, the emulator startup was moved to `.github/workflows/ci.yml`, and `STORAGE_EMULATOR_HOST` was set globally for the CI job.
*   Now, the `fs` instance in `test_credentials_from_raw_token` automatically picks up `STORAGE_EMULATOR_HOST` and talks to the **emulator**. Since the emulator doesn't validate tokens, the call succeeds, causing the test to fail because it expected an exception.

gcsfs/tests/test_core.py
Yonghui-Lee: Yes, it can pass. The reason to skip it on emulator is that emulator doesn't support signature validation. So it passes but it doesn't validate the correctness of `sign`. Skipping it will make it explicit.

gcsfs/tests/settings.py
Yonghui-Lee: Done

PR #846, opened 2026-05-14
Add mrd pool cache

Yonghui-Lee:
## Summary

This pull request introduces a filesystem-level cache of MRD pools, `MRDPoolCache`. 

Previously, opening a `ZonalFile` or invoking range operations (like `_cat_ranges` or `read_into_memory_chunked`) initialized a new `MRDPool` (and its constituent MRDs) from scratch. This added significant overhead due to connection setups, handshakes, and stream initialization.

With this change, `MRDPoolCache` manages `MRDPool` lifecycle at the filesystem level. It implements an LRU eviction cache for idle pools while using reference counting to protect active pools from eviction. Idle pools hold their underlying gRPC downloaders and streams in a shared queue, enabling reads of the same object generation to reuse existing gRPC connections instantly.

### 1. Core Cache Logic
- **`MRDPoolCache` Class**:
  - Caches MRD pools keyed by `(bucket, object, generation)`.
  - Uses reference counting (`_incref`, `_decref`) to track active pools in use.
  - Leverages an `OrderedDict` for LRU eviction of idle pools once the `max_idle_pools` limit is exceeded.
  - Safely handles pool initialization failures by cleaning up partially initialized downloaders and purging the cache key unconditionally.
- **`MRDPool` Enhancements**:
  - Now accepts an optional `cache` reference.
  - Introduces `_get_or_create_mrd()` to query the cache for an idle MRD before falling back to spawning a new one on-demand.
  - On `close()`, instead of forcefully closing all MRDs, it releases them back to the cache queue (`cache.release()`) if a cache is attached.

### 2. Filesystem-level Management
- **Lifecycle & Finalization**:
  - `ExtendedGcsFileSystem` instantiates a single `_mrd_pool_cache` on construction with customizable size via `mrd_pool_cache_size`.
  - Registers `_finalize_mrd_pool_cache` weakref finalizers to gracefully close all cached streams on GC, handling various event loop scenarios safely.
- **Resource Sharing in Readers**:
  - Updated `_cat_ranges`, `read_into_memory_chunked`, and `_get_file` to fetch `MRDPool` instances via `self._mrd_pool_cache.get(...)` rather than constructing raw `MRDPool` instances.

### 3. Zonal File Integration
- Integrates `ZonalFile` with the new cache, retrieving the pool synchronously via on initialization.
- Removes explicit `mrd_pool.initialize` call during init, since cache fetching automatically handles the initialization transparently.

### 4. Test
- **`test_zb_hns_utils.py`**:
  - Comprehensive unit tests verifying `MRDPoolCache` behaviors
- **`conftest.py`**:
  - Introduced `_close_gcs()` / `_close_gcs_async()` to explicitly release gRPC and cache resources in sync/async fixtures, preventing cross-test resource leakage.


gcsfs/zb_hns_utils.py
```diff
@@ -429,20 +451,185 @@ async def close(self):
...
+        the cache if available, otherwise closes them.
         """
         async with self._lock:
             if self._closed:
                 return
 
-            tasks = []
-            for mrd in self._all_mrds:
-                tasks.append(mrd.close())
+            if self._initialized and self._free_mrds.qsize() != len(self._all_mrds):
```
zhixiangli: With `pool_size = 1` and `mrd_supports_multi_request = True`, 
* Task A takes the only MRD and Task B reuses it via round-robin.
* After Task A returns the MRD, `self._free_mrds.qsize() == len(self._all_mrds)` evaluates to True, even though Task B is still using it.

Is it okay to close in this case?

gcsfs/extended_gcsfs.py
```diff
@@ -1570,48 +1621,46 @@ async def _get_file(self, rpath, lpath, callback=None, **kwargs):
...
-            if size is None:
-                logger.warning(
-                    f"AsyncMultiRangeDownloader (MRD) for {rpath} has no 'persisted_size'. "
-                    "Falling back to _info() to get the file size. "
-                    "This may result in incorrect behavior for unfinalized objects."
-                )
-                size = (await self._info(rpath))["size"]
-            callback.set_size(size)
+            async with mrd_pool.get_mrd() as mrd:
+                size = mrd.persisted_size
```
zhixiangli, line 1638: When an MRD instance is first created, it queries GCS for the object's metadata to determine its size, which is then stored as persisted_size (e.g., X).

When a write operation appends data, the file on GCS grows to a new size, Y (where Y > X). However, because the object's generation ID remains unchanged, the MRDPoolCache yields the old, cached MRD for subsequent reads.

Since this cached MRD still reports persisted_size = X, it is unaware that the file has grown to Y, causing errors in future reads.

Does this make sense?

zhixiangli (changes requested):

gcsfs/zb_hns_utils.py
Yonghui-Lee: The responsibility of MRD Pool is to call close to all MRDs. And MRD itself will throw an error to handle this case. I remove the check here.

gcsfs/extended_gcsfs.py
Yonghui-Lee: That's a good catch. I change it to always create mrd on pool initialization to get the new persisted_size.

zhixiangli: While always creating a new MRD on pool initialization ensures a fresh size for a new pool, the MRDPool instance still stores this persisted_size for its lifetime. Since the generation ID does not change when data is appended, any reuse of the pool instance across operations after an append will operate with a stale persisted_size. This means it will still accept stale MRDs from the cache and fail to read newly appended data.

Yonghui-Lee: It's a design decision when we introduced MRDPool to balance staleness and performance. Users need to reopen a file to see the newly appended data, which I think is also the fsspec semantic.

gcsfs/extended_gcsfs.py
```diff
@@ -130,6 +133,36 @@ def __init__(self, *args, finalize_on_close=False, **kwargs):
...
+        except RuntimeError:
+            current_loop = None
+
+        if loop and loop.is_running():
+            asyncio.run_coroutine_threadsafe(cache.close(), loop)
+        elif current_loop is not None and current_loop.is_running():
+            asyncio.run_coroutine_threadsafe(cache.close(), current_loop)
+        elif asyn.loop[0] is not None and asyn.loop[0].is_running():
+            try:
+                asyn.sync(asyn.loop[0], cache.close, timeout=0.1)
```
zhixiangli: Would 100ms timeout too short to close all MRDs?

gcsfs/zb_hns_utils.py
```diff
@@ -429,20 +453,182 @@ async def close(self):
...
+
+        Args:
+            key (tuple): Cache key (bucket, object, generation).
+            mrds (list): List of MRDs to release.
+        """
+        mrds_to_close = []
+        mrd_queue = self._mrd_queues.get(key)
+        if mrd_queue is not None:
+            for mrd in mrds:
+                mrd_queue.append(mrd)
```
zhixiangli: The `mrd_queue` is unbounded. MRDs should be closed if it exceeds the pool size.

gcsfs/extended_gcsfs.py
```diff
@@ -1570,48 +1621,46 @@ async def _get_file(self, rpath, lpath, callback=None, **kwargs):
             return
         callback = callback or NoOpCallback()
 
-        mrd = None
+        mrd_pool = await self._mrd_pool_cache.get(bucket, key, generation, pool_size=1)
```
zhixiangli, line 1635: With a pool size of 1, the pool cache initializes a new pool every time, resulting in no performance improvements and added overhead.

Do you have any measurements showing the improvements made?

zhixiangli (changes requested):

gcsfs/extended_gcsfs.py
```diff
@@ -8,6 +8,7 @@
 from enum import Enum
 from glob import has_magic
 
+import fsspec
```
ankitaluthra1, line 11: can we make this targetted import ?

gcsfs/extended_gcsfs.py
ankitaluthra1: IIUC this assumes the data is read only like in training phase, this should be clearly documented with default OFF

gcsfs/extended_gcsfs.py
ankitaluthra1: +1 to this
@googlyrahman @Yonghui-Lee do we have any macro benchmark results which confirm not just this change but overall MRD pooling is significantly improving performance ? it ll be good to add in description

gcsfs/extended_gcsfs.py
Yonghui-Lee: CMIIM, it's fsspec semantic. For any subclasses of AbstractBufferedFile, we will not be able to read newly appended data after opening the file. Because the file size is stored in self.size when init. This size is never refreshed or updated during subsequent reads or seeks.

gcsfs/extended_gcsfs.py
Yonghui-Lee: It's for `fsspec.FSTimeoutError`. We use the same import in the core.py.

gcsfs/extended_gcsfs.py
zhixiangli: gcsfs does not support read-write modes per https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L1926-L1927

gcsfs/zb_hns_utils.py
Yonghui-Lee: Pool size is not supposed to limit the mrd pool cache. It's the limit for mrd pool. Multiple pools sharing the same mrd pool cache will have different pool size. The size of `mrd_queue` is nearly maximum of the sum of the numbers of MRDs opened by all pools with the same key.

zhixiangli: I suspect there are some leaks:

1. cache.get() creates mrd_1, and release() appends mrd_1 to mrd_queue.
2. cache.get() creates mrd_2, and release() appends mrd_2 to mrd_queue.
3. cache.get() creates mrd_3, and release() appends mrd_3 to mrd_queue.
...

gcsfs/extended_gcsfs.py
Yonghui-Lee: > gcsfs does not support read-write modes per https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L1926-L1927

Yes. It doesn't support. So the previous discussion is based on 2 opened files with "rb" and "wb".

gcsfs/zb_hns_utils.py
Yonghui-Lee: I add a new parameter for the limit of the queue size.

gcsfs/extended_gcsfs.py
Yonghui-Lee: Changed to 5s.

gcsfs/extended_gcsfs.py
Yonghui-Lee: I added a micro benchmark to show the improvement of read throughput. For the current tessellation macro benchmark, I think pool cache may not improve it because we don't open the same file multiple times in the same process.

image

gcsfs/zb_hns_utils.py
```diff
@@ -429,20 +453,182 @@ async def close(self):
...
+            MRDPool: An initialized MRDPool instance.
+        """
+        if self._closed:
+            raise RuntimeError("MRDPoolCache is closed.")
+        fs = self._gcsfs()
+        if fs is None:
+            raise RuntimeError("ExtendedGcsFileSystem has been garbage collected.")
+
+        if generation is None:
+            info = await fs._info(f"{bucket_name}/{object_name}")
```
zhixiangli, line 570: Will this extra round-trip negatively impact performance?

gcsfs/zb_hns_utils.py
```diff
@@ -429,20 +453,182 @@ async def close(self):
...
+            generation = info.get("generation")
```
zhixiangli, line 571: Does calling `fs._info()` on an unfinalized object may return a previous finalized generation (causing the read stream to connect to stale data) or fail entirely if no finalized version exists?

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
```diff
@@ -30,3 +30,19 @@ scenarios:
   - name: "read_rand_fixed_duration_multi_thread"
     pattern: "rand"
     threads: [32]
+
+  - name: "read_repeatedly_open_same_file_fixed_duration"
+    pattern: "repeatedly_open_same_file"
```
zhixiangli: Would `reopen` be cleaner?

zhixiangli (changes requested):

gcsfs/zb_hns_utils.py
Yonghui-Lee: `fs._info()` always returns the unfinalized version.

gcsfs/zb_hns_utils.py
Yonghui-Lee: Info calls are fast. And extra round-trip is also measured in the newly added benchmark. The amortized cost will be little.

gcsfs/tests/perf/microbenchmarks/read/configs.yaml
Yonghui-Lee: Done

zhixiangli (approved):

PR #847, opened 2026-05-14
Fix Storage Control endpoint resolution for TPC

Mahalaxmibejugam:
This PR fixes the endpoint resolution for the Storage Control API in ExtendedGcsFileSystem to support TPC environments correctly.

Changes
- gcsfs/extended_gcsfs.py: Refactored _get_control_plane_client to use self._location for deriving secure or insecure channel endpoints dynamically
- gcsfs/tests/test_extended_hns_gcsfs.py: Added test_get_control_plane_client_endpoint to cover explicit endpoint, universe domain, emulator, and default fallback scenarios.
- cloudbuild/setup_vm.sh: Added a log to print the fsspec version after installing it from head, to help debug CI issues.
- Improved is_real_gcs() to dynamically detect custom universes using _location().
- Configured gcs_factory to inject default project and location parameters to prevent location and project validation errors in private clouds. In TPC universe, providing a location is must while creating the buckets unlike standard production where it has a default region

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -8,6 +8,7 @@ from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
 from glob import has_magic
 
+import grpc
 from fsspec import asyn
 from fsspec.callbacks import NoOpCallback
 from google.api_core import exceptions as api_exceptions
@@ -188,7 +189,14 @@ class ExtendedGcsFileSystem(GCSFileSystem):
                 endpoint = self._location.split("://")[-1]
                 channel_kwargs["host"] = endpoint
 
-            channel = transport_cls.create_channel(**channel_kwargs)
+            if self._location and self._location.startswith("http://"):
+                host = channel_kwargs["host"]
+                channel = grpc.aio.insecure_channel(
+                    host, options=channel_kwargs.get("options")
+                )
+            else:
+                channel = transport_cls.create_channel(**channel_kwargs)
+
             transport = transport_cls(channel=channel)
             self._storage_control_client = storage_control_v2.StorageControlAsyncClient(
                 transport=transport
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -160,8 +160,8 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         if self._grpc_client is None:
             client_options = ClientOptions(quota_project_id=self._user_project)
             if self._location:
-                # client_options expects only the host:port, without the protocol.
-                endpoint = self._location.split("://")[-1]
+                # client_options expects only the host:port, without any protocol or path components.
+                endpoint = self._location.split("://")[-1].split("/")[0]
                 client_options.api_endpoint = endpoint
             self._grpc_client = AsyncGrpcClient(
                 credentials=self.credential,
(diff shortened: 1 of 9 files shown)


zhixiangli (approved):

gcsfs/extended_gcsfs.py
```diff
@@ -179,11 +180,23 @@ async def _get_control_plane_client(self):
...
-                quota_project_id=self._user_project,
-            )
+            channel_kwargs = {
+                "credentials": self.credential,
+                "options": [("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
+                "quota_project_id": self._user_project,
+            }
+            if self._location:
+                endpoint = self._location.split("://")[-1]
+                channel_kwargs["host"] = endpoint
```
ankitaluthra1, line 190: in case of self_location being empty, channel is not getting initialised, is that expected ?

gcsfs/extended_gcsfs.py
```diff
@@ -179,11 +180,23 @@ async def _get_control_plane_client(self):
...
+
+            if self._location and self._location.startswith("http://"):
+                host = channel_kwargs["host"]
+                channel = grpc.aio.insecure_channel(
+                    host, options=channel_kwargs.get("options")
```
ankitaluthra1: quota project in insecure grpc channels is intentionally skipped ?

ankitaluthra1 (changes requested):

gcsfs/extended_gcsfs.py
```diff
@@ -179,11 +180,23 @@ async def _get_control_plane_client(self):
...
-                options=[("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
-                quota_project_id=self._user_project,
-            )
+            channel_kwargs = {
+                "credentials": self.credential,
+                "options": [("grpc.primary_user_agent", f"{USER_AGENT}/{version}")],
+                "quota_project_id": self._user_project,
+            }
+            if self._location:
+                endpoint = self._location.split("://")[-1]
```
ankitaluthra1: can self_location contain string with more than domain name like https://storage.googleapis.com/storage/v1/, in this case endpoint will have more than domain name, is that okay ?

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: If self._location is empty, channel is initialised in the else block without passing the endpoint. In such cases default storage endpoint would be used in the control client. 

Additionally, self._location() will always have a default standard endpoint if it is not specified in the environment variables or during FileSystem creation.

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: Addressed for storage control client and grpc client. 

However, self._location is intended to contain only protocol://host:port, as strictly documented in the GCSFileSystem documentation. Having anything like https://storage.googleapis.com/storage/v1/ would break core.py.

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: grpc.aio.insecure_channel() function does not have support for concept of Google-specific options like quota_project_id

We can pass quota_project_id in transport_cls() but passing it along with channel would completely ignore the quota_project_id. Since insecure channels are only used for local emulators(starting with http://) that lack any quota or billing systems, the emulator would completely ignore these GCP billing headers even if they were sent.

gcsfs/tests/utils.py
```diff
@@ -44,4 +44,13 @@ def tmpfile(extension="", dir=None):
...
+    if (
+        "STORAGE_EMULATOR_HOST" not in os.environ
+        and "GOOGLE_CLOUD_UNIVERSE_DOMAIN" not in os.environ
+    ):
+        return False
+
+    from gcsfs.core import _location
+
+    host = _location()
+    return host.startswith("https://")
```
ankitaluthra1, line 56: Can custom endpoint for real GCS start with http ?

Mahalaxmibejugam: I am checking with the TPC team regarding this. If http:// endpoints are supported, we will need to update both the production code and the test utilities. I will include the necessary changes in my next PR if http is confirmed to be supported.

PR #849, opened 2026-05-25
Fix BigQuery ingestion errors and improve Cloud Build robustness

Mahalaxmibejugam:
This PR fixes BigQuery ingestion failures caused by schema shifts and improves the robustness of the Cloud Build pipelines.

Key Changes:
- BigQuery Ingestion: Refactored ingest.sql to use dynamic SQL with explicit, alphabetically ordered column lists. This prevents type/count mismatches when schema evolves.
- Fail-Fast Validations: Added checks for required substitution variables in both benchmarks and ingestion pipelines to prevent wasted resources.
- Cleanup Fallback: Added TTLs (1h/4h) to OS Login SSH keys in E2E and benchmarks pipelines as a safety fallback.
- Bug Fix: Fixed a bug in benchmarks-cloudbuild.yaml that added a trailing hyphen to sanitized branch names (e.g., main-).

### DIFF
--- a/cloudbuild/benchmarks/ingest.sql
+++ b/cloudbuild/benchmarks/ingest.sql
@@ -1,47 +1,57 @@
--- 1. Variable declarations must be at the top
-DECLARE alter_stmt STRING;
+BEGIN
+  -- 1. Variable declarations must be at the top
+  DECLARE alter_stmt STRING;
+  DECLARE columns_list STRING;
+  DECLARE insert_query STRING;
 
--- 2. Ensure the history table exists with metadata columns
-CREATE TABLE IF NOT EXISTS `@PROJECT_ID@.@DATASET_NAME@.history`
-(
-  run_date DATE,
-  build_id STRING,
-  run_timestamp TIMESTAMP,
-  source_uri STRING,
-  branch_name STRING
-)
-PARTITION BY run_date;
+  -- 2. Ensure the history table exists with metadata columns
+  CREATE TABLE IF NOT EXISTS `@PROJECT_ID@.@DATASET_NAME@.history`
+  (
+    run_date DATE,
+    build_id STRING,
+    run_timestamp TIMESTAMP,
+    source_uri STRING,
+    branch_name STRING
+  )
+  PARTITION BY run_date;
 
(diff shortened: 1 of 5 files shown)


cloudbuild/benchmarks/ingest.sql
```diff
@@ -1,47 +1,57 @@
...
+    FROM `@PROJECT_ID@.@DATASET_NAME@.INFORMATION_SCHEMA.COLUMNS`
+    WHERE table_name = 'staging'
+  );
+
+  SET insert_query = CONCAT(
+    "INSERT INTO `@PROJECT_ID@.@DATASET_NAME@.history` (run_date, build_id, run_timestamp, source_uri, branch_name, ",
+    columns_list,
+    ") SELECT PARSE_DATE('%d%m%Y', REGEXP_EXTRACT(_FILE_NAME, r'/(\\d{8})/')) as run_date, ",
+    "REGEXP_EXTRACT(_FILE_NAME, r'/([0-9a-fA-F-]{36})/') as build_id, ",
+    "PARSE_TIMESTAMP('%d%m%Y-%H%M%S', REGEXP_EXTRACT(_FILE_NAME, r'/(\\d{8}-\\d{6})/')) as run_timestamp, ",
```
zhixiangli, line 49: Could we change the format to %Y%m%d so that it's directly comparable as a string after sorting? This would be helpful for automatically comparing branch results with the main branch, specifically for pulling and aggregating results from the last X days (sort, compare, filter).

Mahalaxmibejugam: Changing the format to %Y%m%d would be a breaking change, as existing objects in the results bucket use the ddmmyyyy format and would be incompatible with %Y%m%d. While we could format the date after parsing, that would result in inconsistent date formats within the history table. Since run_date is a DATE type rather than a string, sorting should function correctly regardless of the display format.

zhixiangli: Thanks for the explanation. I considered using bash to list the files and sort them to filter the results from the past X days directly from the buckets. However, since this involves a breaking change, we will likely need to use BigQuery to handle the filtering.

zhixiangli (approved):

PR #851, opened 2026-05-29
Optimize benchmarks pipeline, fix quota leaks and handle manual builds

Mahalaxmibejugam:
This PR improves the robustness and maintainability of the benchmarks Cloud Build pipeline in benchmarks-cloudbuild.yaml

- Guaranteed teardown: Moved benchmark fixture teardown to a finally block to guarantee GCS resource cleanup even if the setup phase fails which would improve the pipeline execution time by an hour. Earlier if setup fails, teardown was not handled and delete-buckets step in cloudBuild took almost an hour to delete objects inside the bucket.
- Optimized Empty File Handling: Updated logic to use pipe() instead of open() + info() for empty files, reducing overhead and improving performance of read benchmarks.
- Fixed Quota Issues: Added a cleanup-leaked-resources step at the beginning of the build to find and delete leaked VMs (older than 5 hours) with the prefix ${_INFRA_PREFIX}-vm-. This prevents failed or cancelled builds from exhausting project quotas.
- Reduced Code Duplication: Combined the create-io-vm and create-metadata-vm steps into a single reusable create-vm step. It uses bash conditions to apply high-performance flags (like hyperdisk and GVNIC) only when _BENCHMARK_TYPE is IO.
- Fixed Manual Build Failures: Added a conditional check during the installation of gcsfs on the VM. If the .git directory is missing (which happens during manual gcloud builds submit runs), it sets a dummy version 0.0.0 to prevent hatch-vcs from failing.

### DIFF
--- a/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
+++ b/cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
@@ -63,6 +63,39 @@ steps:
         echo "export VM_NAME='${_INFRA_PREFIX}-vm-$${SHORT_BUILD_ID}'" >> /workspace/build_vars.env
     waitFor: ["-"]
 
+  # 2.5 Clean up leaked resources from previous failed builds.
+  - name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
+    id: "cleanup-leaked-resources"
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        source /workspace/build_vars.env
+        echo "Cleaning up leaked resources with prefix: ${_INFRA_PREFIX}"
+
+        CURRENT_TIME=$$(date +%s)
+        # 5 hours threshold
+        THRESHOLD=18000
+
+        VMS=$$(gcloud compute instances list --project="${PROJECT_ID}" --filter="name~'${_INFRA_PREFIX}-vm-'" --format="value(name,zone,creationTimestamp)")
+
+        while read -r name zone creation_time; do
+          if [ -z "$$name" ]; then continue; fi
+
+          CREATION_SECONDS=$$(date -d "$$creation_time" +%s)
+          AGE=$$(($$CURRENT_TIME - $$CREATION_SECONDS))
+
+          if [ $$AGE -gt $$THRESHOLD ]; then
+            echo "Deleting leaked VM: $$name in zone $$zone (Age: $$AGE seconds)"
+            gcloud compute instances delete "$$name" --project="${PROJECT_ID}" --zone="$$zone" --quiet || true
+          else
+            echo "Skipping VM: $$name (Age: $$AGE seconds)"
+          fi
+        done <<< "$$VMS"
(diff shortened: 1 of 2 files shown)


cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
```diff
@@ -176,7 +169,8 @@ steps:
             --service-account="${_VM_SERVICE_ACCOUNT}" \
             --tags="allow-ssh" \
             --labels="build-id=$$RUN_ID" \
-            --quiet
+            --quiet \
+            "${EXTRA_ARGS[@]}"
```
zhixiangli: $${EXTRA_ARGS[@]}?

gcsfs/tests/perf/microbenchmarks/conftest.py
```diff
@@ -41,20 +41,22 @@ def _write_file(gcs, path, file_size, chunk_size):
 
 
 def _prepare_files(gcs, file_paths, file_size=0):
-    if file_size > 0:
+    if file_size == 0:
+        try:
+            gcs.pipe({path: b"" for path in file_paths})
+        except Exception as e:
+            pytest.fail(f"Failed to pipe files: {e}")
+    else:
```
zhixiangli: If you return early from the if block, an else block is not needed.

gcsfs/tests/perf/microbenchmarks/conftest.py
```diff
@@ -41,20 +41,22 @@ def _write_file(gcs, path, file_size, chunk_size):
 
 
 def _prepare_files(gcs, file_paths, file_size=0):
-    if file_size > 0:
+    if file_size == 0:
+        try:
+            gcs.pipe({path: b"" for path in file_paths})
```
zhixiangli, line 46: This removes the multiprocessing pool and serializes empty-object creation. Will this significantly increase benchmark setup time when creating a massive number of objects in the metadata benchmark?

Mahalaxmibejugam: No, this will likely be faster and more efficient. While it removes the Python multiprocessing pool, gcs.pipe with a dictionary of paths leverages fsspec's async capabilities to execute the file creations concurrently within a single process event loop. For empty objects, the actual work is minimal, and avoiding the overhead of spawning and managing a pool of 200 processes will actually speed up the setup phase when creating empty objects.

cloudbuild/benchmarks/benchmarks-cloudbuild.yaml
Mahalaxmibejugam: Changed it to $${EXTRA_ARGS[@]} to be consistent with other bash variables and to distinguish it from CloudBuild substitution variables.

gcsfs/tests/perf/microbenchmarks/conftest.py
Mahalaxmibejugam: Done

PR #854, opened 2026-06-02
Fix concurrent MRD reuse by tracking in-flight MRDs on close

zhixiangli:
Fix concurrent MRD reuse by tracking in-flight MRDs on close

MRDPool.close() previously extracted all instantiated MRDs (`self._all_mrds`), including ones actively yielded by `get_mrd()`, and donated them back to the MRDPoolCache. A concurrent task could acquire the active MRD, driving it from two contexts simultaneously and causing corrupted reads or gRPC errors.

This change drains only `_free_mrds` on `close()`. The pool is marked closed so any in-flight MRDs eventually execute `await close_mrd(mrd)` in their `finally` block instead of being donated to the cache, preventing reuse.

### DIFF
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -446,8 +446,11 @@ class MRDPool:
             # Only return the MRD to the free queue if we were the ones who took it out
             # or if we just spawned it. This prevents duplicate entries in the queue
             # when multiple concurrent tasks share the same MRD via round-robin.
-            if (create_new or used_from_queue) and not self._closed:
-                self._free_mrds.put_nowait(mrd)
+            if create_new or used_from_queue:
+                if self._closed:
+                    await close_mrd(mrd)
+                else:
+                    self._free_mrds.put_nowait(mrd)
 
     async def close(self):
         """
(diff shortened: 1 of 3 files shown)


gcsfs/zb_hns_utils.py
```diff
@@ -459,15 +462,19 @@ async def close(self):
         async with self._lock:
             if self._closed:
                 return
+            self._closed = True
+
+            free_mrds = []
+            while not self._free_mrds.empty():
```
Yonghui-Lee, line 468: In the first version, we have the following check. I remove it to let MRD handle by itself.

```
if self._initialized and self._free_mrds.qsize() != len(self._all_mrds):
                raise RuntimeError("Cannot close pool with active MRDs.")
```

My idea is that closing when using is probably a bug. If mrd itself can handle the concurrent cases, we can let it handle. Otherwise, maybe we should raise an error. If we defer the close and the read hangs indefinitely, the resource will be leaked.

zhixiangli: This PR change checks if the MRDPool is closed. If it is, the MRD is closed after use; otherwise, it is returned to the `free_mrd`. May I know your suggestions on this approach?

Yonghui-Lee: We have 2 ways to handle the concurrent cases.

First is avoiding closing when using in our codes. MRD pool is only internally used. If there are some unexpected cases, let MRD handles them, which will raise errors. It's the current design.

Second is handling them in MRD pool. If we want to do it, we need to consider more cases. For example, when mrd_supports_multi_request is true, even if mrd is in free queue, it might still be in use

zhixiangli: I understand your point about systematically fixing the entire problem at once. However, if you look at the commit, it is co-authored with Jules as part of an experiment; I believe we can rely on that tool to handle the rest, and this change should not be blocked since it improves the current situation.

Yonghui-Lee (approved):

PR #856, opened 2026-06-02
Refcount in-flight MRDs in MRDPool

zhixiangli:
Fixes an issue where shared MRDs could be closed or requeued by the first caller to finish while other callers were still driving them. This replaces the old flag logic with a per-MRD refcount to ensure an MRD is only requeued/closed by its last active holder.

### DIFF
--- a/gcsfs/tests/test_zb_hns_utils.py
+++ b/gcsfs/tests/test_zb_hns_utils.py
@@ -19,14 +19,9 @@ generation = "12345"
 
 @pytest.mark.asyncio
 async def test_shared_mrd_closed_only_by_last_holder():
-    from unittest import mock
-
-    from gcsfs.zb_hns_utils import MRDPool
-
     class FakeMRD:
         def __init__(self):
             self.persisted_size = 0
-            self.bucket_name, self.object_name = "b", "o"
             self.close_count = 0
 
         async def close(self):
--- a/gcsfs/tests/test_zb_hns_utils.py
+++ b/gcsfs/tests/test_zb_hns_utils.py
@@ -18,6 +18,39 @@ generation = "12345"
 
 
 @pytest.mark.asyncio
+async def test_shared_mrd_closed_only_by_last_holder():
+    from unittest import mock
+    from gcsfs.zb_hns_utils import MRDPool
+
+    class FakeMRD:
+        def __init__(self):
+            self.persisted_size = 0
+            self.bucket_name, self.object_name = "b", "o"
+            self.close_count = 0
+        async def close(self):
+            self.close_count += 1
+
+    pool = MRDPool(mock.Mock(), "b", "o", 1, pool_size=1, cache=None)
+    pool.mrd_supports_multi_request = True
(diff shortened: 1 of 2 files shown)


gcsfs/tests/test_zb_hns_utils.py
```diff
@@ -17,6 +17,43 @@
 generation = "12345"
 
 
+@pytest.mark.asyncio
+async def test_shared_mrd_closed_only_by_last_holder():
+    from unittest import mock
+
+    from gcsfs.zb_hns_utils import MRDPool
```
Yonghui-Lee: We can remove these imports.

gcsfs/tests/test_zb_hns_utils.py
```diff
@@ -17,6 +17,43 @@
...
+
+    class FakeMRD:
+        def __init__(self):
+            self.persisted_size = 0
+            self.bucket_name, self.object_name = "b", "o"
```
Yonghui-Lee: Do we need these fields?

gcsfs/zb_hns_utils.py
```diff
@@ -421,32 +444,36 @@ async def get_mrd(self):
...
-                    mrd = await self._get_or_create_mrd()
-                except BaseException as e:
-                    self._active_count -= 1
-                    raise e
-            elif mrd is None:
-                # We did not spawn a new one and we did not grab one via round-robin.
-                # This means we must wait for a free one from the queue.
+            if mrd is None:
+                # Either the queue was non-empty, or the pool is full and sharing
+                # is disabled: block until a holder returns an MRD. NOTE: the lock
```
Yonghui-Lee: We won't block when queue is non-empty?

gcsfs/tests/test_zb_hns_utils.py
```diff
@@ -17,6 +17,38 @@
 generation = "12345"
 
 
+@pytest.mark.asyncio
+async def test_shared_mrd_closed_only_by_last_holder():
+    class FakeMRD:
+        def __init__(self):
+            self.persisted_size = 0
```
Yonghui-Lee, line 24: We can also remove persisted_size

zhixiangli: Inside  MRDPool.initialize()  (in  gcsfs/zb_hns_utils.py ), the pool creates the first MRD and attempts to read its  persisted_size.

```
    412:             if not self._initialized and self._active_count == 0:                                                                                                                                                                                 
    413:                 # Always create a new MRD on initialization to get the up-to-date persisted_size                                                                                                                                                  
    414:                 mrd = await self._create_mrd()                                                                                                                                                                                                    
    415:                 self._all_mrds.append(mrd)                                                                                                                                                                                                        
    416:                 self.persisted_size = mrd.persisted_size # <--- Accesses mrd.persisted_size 
```

If we remove  self.persisted_size = 0  from  FakeMRD , the test will fail with an  AttributeError

Yonghui-Lee (approved):

PR #857, opened 2026-06-02
Fix teardown order in close_resources()

zhixiangli:
Reordered resource teardown in `close_resources()` so that the MRD pool cache is closed before the gRPC transport. MRDs ride on the gRPC channel, so draining the pool cache after closing the channel was raising exceptions against a dead transport.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -241,7 +241,7 @@ class ExtendedGcsFileSystem(GCSFileSystem):
             )
         return self._storage_control_client
 
-    async def close_resources(self):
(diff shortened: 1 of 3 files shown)


gcsfs/extended_gcsfs.py
```diff
@@ -242,24 +242,30 @@ async def _get_control_plane_client(self):
         return self._storage_control_client
 
     async def close_resources(self):
```
Yonghui-Lee: Please help rename it to `_close_resources`, it's only used for tests

Yonghui-Lee (approved):

Yonghui-Lee (approved):

PR #858, opened 2026-06-02
Route rm paths independently per bucket to support mixed HNS/flat deletions

zhixiangli:
Previously, `ExtendedGcsFileSystem.rm` determined whether to use HNS-specific deletion logic based solely on the bucket of the first path in the input list. This led to incorrect routing and failures when the input list contained a mix of paths from both HNS-enabled and non-HNS (flat) buckets.

This change:
- Groups input paths in `rm` by bucket and processes each group independently.
- Routes HNS buckets to HNS-specific deletion and flat buckets to standard `super()._rm`.
- Ensures that multiple paths within the same bucket (HNS or flat) are still grouped efficiently for batch operations.

Includes comprehensive unit and integration tests to verify:
- Correct independent routing and execution order of mixed HNS/flat paths.
- Efficiency is maintained for same-bucket multi-path deletions.
- Expected error handling for empty input lists.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -1192,8 +1192,11 @@ class ExtendedGcsFileSystem(GCSFileSystem):
         Deletes files and directories.
 
         This method overrides the parent `_rm` to correctly handle directory
-        deletion in HNS-enabled buckets. For non-HNS buckets, it falls back
-        to the parent implementation.
+        deletion in HNS-enabled buckets. Input paths are grouped by bucket and
+        each bucket is routed independently: HNS buckets use HNS-specific
+        deletion, while non-HNS buckets fall back to the parent implementation.
+        This means a single call may mix paths from HNS and flat buckets, and
+        the per-bucket groups are deleted concurrently.
 
         For HNS buckets, it first expands the path to get a list of all files
         and directories, then categorizes them. It deletes files in batches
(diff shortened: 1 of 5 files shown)


gcsfs/extended_gcsfs.py
```diff
@@ -1199,16 +1199,53 @@ async def _rm(self, path, recursive=False, maxdepth=None, batchsize=20):
...
-            # Fall back to the parent's async rm implementation for non-HNS buckets.
+        grouped = {}
+        for p in path:
+            bucket, _, _ = self.split_path(p)
+            grouped.setdefault(bucket, []).append(p)
+
+        results = []
+        succeeded = False
+
+        for bucket, bucket_paths in grouped.items():
```
Mahalaxmibejugam: Can we delete the paths from different buckets concurrently?

gcsfs/extended_gcsfs.py
```diff
@@ -1217,11 +1254,9 @@ async def _rm(self, path, recursive=False, maxdepth=None, batchsize=20):
             path, recursive=recursive, maxdepth=maxdepth, detail=True
         )
 
-        # Separate files and directories based on their type.
-        # Directories must be deleted from the deepest first.
```
Mahalaxmibejugam, line 1227: Should we keep these code comments?

gcsfs/extended_gcsfs.py
```diff
@@ -1199,16 +1199,53 @@ async def _rm(self, path, recursive=False, maxdepth=None, batchsize=20):
...
+            return await self._rm_bucket_paths(
+                bucket,
+                path,
+                recursive=recursive,
+                maxdepth=maxdepth,
+                batchsize=batchsize,
+            )
 
-        is_hns = await self._is_bucket_hns_enabled(bucket)
+        if not path:
```
Mahalaxmibejugam, line 1216: Can we move this edge case to the top of the method?

gcsfs/extended_gcsfs.py
```diff
@@ -1199,16 +1199,53 @@ async def _rm(self, path, recursive=False, maxdepth=None, batchsize=20):
             maxdepth (int, optional): The maximum depth to traverse for deletion.
             batchsize (int): The number of files to delete in a single batch request.
         """
-        if isinstance(path, list):
-            # For HNS check, we can check for bucket type from the first path.
-            bucket, _, _ = self.split_path(path[0]) if path else (None, None, None)
-        else:
+        if not isinstance(path, list):
```
Mahalaxmibejugam: Can we flatten this logic by handling all inputs as a list—converting the path from a string to a list and then calling rm_bucket_paths only once?

gcsfs/tests/integration/test_extended_hns.py
```diff
@@ -968,6 +968,48 @@ def test_rm_wildcards_non_recursive(self, gcs_hns):
         assert not gcsfs.exists(files[0])
         assert gcsfs.exists(files[1])  # subdir file still exists
 
+    @pytest.mark.parametrize("order", ["flat_first", "hns_first"])
+    def test_rm_mixed_buckets_order(self, gcs_hns, order):
+        """Test mixed flat and HNS bucket deletes in different orders."""
+        gcsfs = gcs_hns
+        unique_id = uuid.uuid4().hex
+
+        flat_file = f"{TEST_BUCKET}/mixed_rm_{unique_id}/file.txt"
```
Mahalaxmibejugam: in CI pipeline, TEST_BUCKET and TEST_HNS_BUCKET both point to HNS bucket while running HNS test suite, do we want to create a separate standard bucket for this test and add it in the global buckets to be deleted?

Mahalaxmibejugam (changes requested):

gcsfs/tests/settings.py
```diff
@@ -12,6 +12,10 @@ def _get_bucket_name(env_var: str, default_name: str) -> str:
     "GCSFS_TEST_VERSIONED_BUCKET", "gcsfs_test_versioned"
 )
 TEST_HNS_BUCKET = _get_bucket_name("GCSFS_HNS_TEST_BUCKET", "gcsfs_hns_test")
+# A dedicated standard (non-HNS) bucket. In the HNS test suite GCSFS_TEST_BUCKET
+# may point at an HNS-enabled bucket, so tests that need a genuinely flat bucket
+# (e.g. mixed HNS/flat deletions) should use this one.
```
Mahalaxmibejugam: These details are very specific to cloudbuild pipeline and I think we can avoid them in the global settings.py file. We can add generic description about the bucket type that needs to be used for this environment variable.

Mahalaxmibejugam (approved):

gcsfs/tests/settings.py
zhixiangli: Reworded — settings.py now just notes that GCSFS_FLAT_TEST_BUCKET should point at a standard (non-HNS / flat) bucket. The suite-specific rationale (TEST_BUCKET possibly being HNS in the HNS suite) stays in the flat_bucket fixture docstring in conftest.py.

PR #859, opened 2026-06-03
Fix multiprocessing emulator _get_bucket_type calls

yuxin00j:
This PR resolves hangs when executing the microbenchmark suite locally against the GCS emulator (`fake-gcs-server`).

### The Problem:
Python's `multiprocessing` library (spawn context) launches child processes that do not inherit active mock patches from the parent process. When these child processes instantiate `ExtendedGcsFileSystem`, they attempt to call the real `_get_bucket_type` method. Since the GCS emulator does not support the Storage Control API, this call fails or hangs, leading to timeouts in multi-process test cases.

### Changes:
- **Pool Initialization:** Added `_init_pool_worker` in `gcsfs/tests/perf/microbenchmarks/conftest.py` as an initializer for spawned Pool workers to explicitly bypass `_get_bucket_type` calls on emulator.
- **Process Target Wrapper:** Introduced `_multiprocess_worker_wrapper` in `runner.py` to wrap process target functions and apply the `_get_bucket_type` mock in all spawned child processes.

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/conftest.py
+++ b/gcsfs/tests/perf/microbenchmarks/conftest.py
@@ -42,15 +42,11 @@ def _write_file(gcs, path, file_size, chunk_size):
 
 def _init_pool_worker():
     """Initializer for spawned pool workers to bypass _get_bucket_type calls on emulator."""
-    from gcsfs.tests.utils import is_real_gcs
+    from gcsfs.tests.utils import _patch_get_bucket_type_for_emulator
 
-    if not is_real_gcs():
-        from gcsfs.extended_gcsfs import BucketType, ExtendedGcsFileSystem
-
-        async def mock_get_bucket_type(self, bucket):
-            return BucketType.UNKNOWN
-
-        ExtendedGcsFileSystem._get_bucket_type = mock_get_bucket_type
+    patch = _patch_get_bucket_type_for_emulator()
+    if patch:
+        patch.start()
 
 
 def _prepare_files(gcs, file_paths, file_size=0):
--- a/gcsfs/tests/perf/microbenchmarks/conftest.py
+++ b/gcsfs/tests/perf/microbenchmarks/conftest.py
@@ -40,6 +40,19 @@ def _write_file(gcs, path, file_size, chunk_size):
         )
 
 
+def _init_pool_worker():
+    """Initializer for spawned pool workers to bypass _get_bucket_type calls on emulator."""
+    from gcsfs.tests.utils import is_real_gcs
+
+    if not is_real_gcs():
+        from gcsfs.extended_gcsfs import BucketType, ExtendedGcsFileSystem
+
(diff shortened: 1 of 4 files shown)


gcsfs/tests/perf/microbenchmarks/runner.py
```diff
@@ -118,6 +121,18 @@ def run_multi_threaded_fixed_duration(
     benchmark.pedantic(lambda: None, rounds=1, iterations=1, warmup_rounds=0)
 
 
+def _multiprocess_worker_wrapper(worker_target, args):
```
zhixiangli, line 121: Where is this used?

zhixiangli (changes requested):

yuxin00j:
Good catch by @zhixiangli! I had created `_multiprocess_worker_wrapper` but forgot to actually use it in `run_multi_process`. I have just pushed a commit that applies the wrapper to the spawned child processes.

gcsfs/tests/perf/microbenchmarks/conftest.py
```diff
@@ -40,6 +40,19 @@ def _write_file(gcs, path, file_size, chunk_size):
...
 
 
+def _init_pool_worker():
+    """Initializer for spawned pool workers to bypass _get_bucket_type calls on emulator."""
+    from gcsfs.tests.utils import is_real_gcs
+
+    if not is_real_gcs():
+        from gcsfs.extended_gcsfs import BucketType, ExtendedGcsFileSystem
+
+        async def mock_get_bucket_type(self, bucket):
```
zhixiangli: Have a common helper for this logic

```
  def _patch_get_bucket_type_for_emulator():
      """Patch bucket type detection in spawned workers when running against fake-gcs-server."""
      from gcsfs.tests.utils import is_real_gcs

      if is_real_gcs():
          return None

      from unittest import mock

      from gcsfs.extended_gcsfs import BucketType

      return mock.patch(
          "gcsfs.extended_gcsfs.ExtendedGcsFileSystem._get_bucket_type",
          return_value=BucketType.UNKNOWN,
      )
```

zhixiangli (changes requested):

gcsfs/tests/perf/microbenchmarks/conftest.py
yuxin00j: I've added the `_patch_get_bucket_type_for_emulator` helper in `utils.py` and updated `_init_pool_worker`, `_multiprocess_worker_wrapper`, and `_mock_get_bucket_type_on_emulator` to use it.

gcsfs/tests/perf/microbenchmarks/runner.py
yuxin00j: It is used below in `run_multi_process` (around line 192) as the target for `ctx.Process`.

zhixiangli (approved):

PR #862, opened 2026-06-03
Introduce concurrency in _get_file method

googlyrahman:
## 1. Implementation of _get_file_concurrent in gcsfs/core.py

A new asynchronous method, _get_file_concurrent, has been added to handle multi-threaded downloads ([ source](https://github.com/fsspec/gcsfs/compare/main...ankitaluthra1:gcsfs:diskx?expand=1&content_ref=async+def+_get_file_concurrent+self+rpath+lpath+concurrency+headers+none+callback+none+kwargs)). Key features include:

- Parallel Fetching: Splits the file into parts based on the requested concurrency and fetches them using HTTP Range headers.
- Non-blocking Disk I/O: Sequential disk writes are offloaded to background threads using run_in_executor to keep the event loop unblocked.
- Consistency Verification: Since on-the-fly checksumming is not feasible for out-of-order concurrent downloads, integrity is verified post-download if a consistency check is requested.
- Generation Pinning: The exact object generation is pinned during the process to prevent silent corruption if the source object is updated mid-download.
- Robustness: Includes retry logic via @retry_request and explicit task cancellation on failure.

## 2. Integration with _get_file
The primary _get_file entry point has been updated to automatically route requests to the concurrent implementation when the concurrency keyword argument is greater than 1.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1936,19 +1936,94 @@ class GCSFileSystem(asyn.AsyncFileSystem):
             checker.validate_http_response(r)  # validate file consistency
             return r.status, r.headers, r.request_info, data
 
+    def _init_local_file(self, lpath, total_size):
+        """Creates the target directory and pre-allocates the file size."""
+        os.makedirs(os.path.dirname(lpath) or os.curdir, exist_ok=True)
+        if total_size == 0:
+            open(lpath, "wb").close()
+        else:
+            with open(lpath, "wb") as f:
+                f.truncate(total_size)
+
+    @retry_request(retries=retries)
+    async def _download_chunk_to_file(
+        self,
+        url,
+        rpath,
+        lpath,
+        start_offset,
+        end_offset,
+        chunk_size,
+        headers,
+        callback,
+        **kwargs,
+    ):
+        """Fetches a specific byte range of a file and writes it to disk."""
+        req_headers = {"Range": f"bytes={start_offset}-{end_offset - 1}"}
+        if headers:
+            req_headers.update(headers)
+
+        bytes_written = 0
+        await self._set_session()
+
+        async with self.session.get(
+            url,
+            headers=self._get_headers(req_headers),
+            params=self._get_params(kwargs),
+            timeout=self.requests_timeout,
+        ) as r:
+            validate_response(r.status, None, rpath)
+            local_f = None
+            try:
+                local_f = open(lpath, "rb+")
+                local_f.seek(start_offset)
+                while True:
+                    chunk = await r.content.read(chunk_size)
+                    if not chunk:
+                        break
+
+                    local_f.write(chunk)
+                    callback.relative_update(len(chunk))
+                    bytes_written += len(chunk)
+
+                expected_bytes = end_offset - start_offset
+                if bytes_written != expected_bytes:
+                    raise RuntimeError(
+                        f"Connection closed early. Expected {expected_bytes} bytes, got {bytes_written}."
+                    )
+            except BaseException:
+                callback.relative_update(-bytes_written)
+                raise
+            finally:
+                if local_f is not None:
+                    local_f.close()
+
+    async def _verify_file_consistency(self, lpath, details, consistency):
+        """Verifies the integrity of the downloaded file in a background thread."""
+        loop = asyncio.get_running_loop()
+        checker = get_consistency_checker(consistency)
+
+        def verify_file():
+            with open(lpath, "rb") as f:
+                while True:
+                    data = f.read(1024 * 1024)
+                    if not data:
+                        break
+                    checker.update(data)
+            checker.validate_json_response(details)
+
+        await loop.run_in_executor(None, verify_file)
+
     async def _get_file_concurrent(
         self, rpath, lpath, concurrency, headers=None, callback=None, **kwargs
     ):
-        """
-        Concurrent write the gcs object into the disk.
-        """
+        """Main orchestrator for concurrent file downloads."""
         consistency = kwargs.pop("consistency", self.consistency)
-        chunk_size = kwargs.pop(
-            "chunk_size", 128 * 1024
-        )  # 128KB micro-chunks for streaming
+        chunk_size = kwargs.pop("chunk_size", 128 * 1024)
 
         details = await self._info(rpath, **kwargs)
         total_size = details.get("size", 0)
+
         if total_size < self.MIN_CHUNK_SIZE_FOR_CONCURRENCY:
             concurrency = 1
 
@@ -1962,9 +2037,9 @@ class GCSFileSystem(asyn.AsyncFileSystem):
         callback = callback or NoOpCallback()
         callback.set_size(total_size)
 
-        os.makedirs(os.path.dirname(lpath) or os.curdir, exist_ok=True)
+        self._init_local_file(lpath, total_size)
+
         if total_size == 0:
-            open(lpath, "wb").close()
             if check_consistency:
                 checker = get_consistency_checker(consistency)
                 checker.update(b"")
@@ -1972,57 +2047,26 @@ class GCSFileSystem(asyn.AsyncFileSystem):
             return
 
         part_size = (total_size + concurrency - 1) // max(1, concurrency)
-        with open(lpath, "wb") as f:
-            f.truncate(total_size)
-
-        loop = asyncio.get_running_loop()
         url = self.url(rpath)
-
-        @retry_request(retries=self.retries)
-        async def stream_and_write(start_offset, end_offset):
-            req_headers = {"Range": f"bytes={start_offset}-{end_offset - 1}"}
-            if headers:
-                req_headers.update(headers)
-
-            bytes_written = 0
-            await self._set_session()
-            async with self.session.get(
-                url,
-                headers=self._get_headers(req_headers),
-                params=self._get_params(kwargs),
-                timeout=self.requests_timeout,
-            ) as r:
-                validate_response(r.status, None, rpath)
-                local_f = None
-                try:
-                    local_f = open(lpath, "rb+")
-                    local_f.seek(start_offset)
-                    while True:
-                        chunk = await r.content.read(chunk_size)
-                        if not chunk:
-                            break
-
-                        local_f.write(chunk)
-
-                        callback.relative_update(len(chunk))
-                        bytes_written += len(chunk)
-
-                    expected_bytes = end_offset - start_offset
-                    if bytes_written != expected_bytes:
-                        raise RuntimeError(
-                            f"Connection closed early. Expected {expected_bytes} bytes, got {bytes_written}."
-                        )
-                except BaseException:
-                    callback.relative_update(-bytes_written)
-                    raise
-                finally:
-                    if local_f is not None:
-                        local_f.close()
-
         tasks = []
+
         for offset in range(0, total_size, part_size):
             end = min(offset + part_size, total_size)
-            tasks.append(asyncio.create_task(stream_and_write(offset, end)))
+            tasks.append(
+                asyncio.create_task(
+                    self._download_chunk_to_file(
+                        url,
+                        rpath,
+                        lpath,
+                        offset,
+                        end,
+                        chunk_size,
+                        headers,
+                        callback,
+                        **kwargs,
+                    )
+                )
+            )
 
         try:
             await asyncio.gather(*tasks)
@@ -2038,19 +2082,7 @@ class GCSFileSystem(asyn.AsyncFileSystem):
                 "On-the-fly checksum disabled for concurrent out-of-order downloads. "
                 "Verifying file integrity post-download."
             )
-            checker = get_consistency_checker(consistency)
-
-            def verify_file():
-                with open(lpath, "rb") as f:
-                    while True:
-                        data = f.read(1024 * 1024)
-                        if not data:
-                            break
-                        checker.update(data)
-                checker.validate_json_response(details)
-
-            # Execute the hashing in a background thread to keep the event loop unblocked
-            await loop.run_in_executor(None, verify_file)
+            await self._verify_file_consistency(lpath, details, consistency)
 
     async def _get_file(self, rpath, lpath, callback=None, **kwargs):
         if os.path.isdir(lpath):
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1940,7 +1940,8 @@ class GCSFileSystem(asyn.AsyncFileSystem):
         """Creates the target directory and pre-allocates the file size."""
         os.makedirs(os.path.dirname(lpath) or os.curdir, exist_ok=True)
         if total_size == 0:
-            open(lpath, "wb").close()
+            with open(lpath, "wb"):
+                pass
         else:
             with open(lpath, "wb") as f:
                 f.truncate(total_size)
@@ -1959,9 +1960,10 @@ class GCSFileSystem(asyn.AsyncFileSystem):
         **kwargs,
     ):
         """Fetches a specific byte range of a file and writes it to disk."""
-        req_headers = {"Range": f"bytes={start_offset}-{end_offset - 1}"}
+        req_headers = {}
         if headers:
             req_headers.update(headers)
+        req_headers["Range"] = f"bytes={start_offset}-{end_offset - 1}"
 
         bytes_written = 0
         await self._set_session()
@@ -1988,8 +1990,9 @@ class GCSFileSystem(asyn.AsyncFileSystem):
 
                 expected_bytes = end_offset - start_offset
                 if bytes_written != expected_bytes:
-                    raise RuntimeError(
-                        f"Connection closed early. Expected {expected_bytes} bytes, got {bytes_written}."
+                    raise aiohttp.client_exceptions.ClientPayloadError(
+                        f"Connection closed early. Expected {expected_bytes} bytes, "
+                        f"got {bytes_written} at offset {start_offset}."
                     )
             except BaseException:
                 callback.relative_update(-bytes_written)
@@ -2027,6 +2030,11 @@ class GCSFileSystem(asyn.AsyncFileSystem):
         if total_size < self.MIN_CHUNK_SIZE_FOR_CONCURRENCY:
             concurrency = 1
 
+        if concurrency == 1:
+            return await self._get_file_request(
+                self.url(rpath), lpath, headers=headers, callback=callback, **kwargs
+            )
+
         check_consistency = consistency not in ("none", None)
 
         # Prevent silent corruption by pinning the exact object generation
@@ -2046,7 +2054,7 @@ class GCSFileSystem(asyn.AsyncFileSystem):
                 checker.validate_json_response(details)
             return
 
-        part_size = (total_size + concurrency - 1) // max(1, concurrency)
+        part_size = max(chunk_size, (total_size + concurrency - 1) // concurrency)
         url = self.url(rpath)
         tasks = []
 
@@ -2070,15 +2078,15 @@ class GCSFileSystem(asyn.AsyncFileSystem):
 
         try:
             await asyncio.gather(*tasks)
-        except BaseException as e:
+        except BaseException:
             for t in tasks:
                 if not t.done():
                     t.cancel()
             await asyncio.gather(*tasks, return_exceptions=True)
-            raise e
+            raise
 
         if check_consistency:
-            logger.debug(
+            logger.warning(
                 "On-the-fly checksum disabled for concurrent out-of-order downloads. "
                 "Verifying file integrity post-download."
             )
@@ -2091,13 +2099,32 @@ class GCSFileSystem(asyn.AsyncFileSystem):
         callback = callback or NoOpCallback()
 
         concurrency = kwargs.pop("concurrency", DEFAULT_CONCURRENCY)
-        if concurrency > 1:
-            await self._get_file_concurrent(
-                rpath, lpath, concurrency, callback=callback, **kwargs
+
+        # If the user provides a custom Range header, fall back to the sequential path.
+        # Parsing range headers client-side to spawn concurrent tasks is messy because
+        # formats vary widely (e.g., bytes=-500, bytes=100-, bytes=0-50, 100-150).
+        # Building a flawless parser for this edge case introduces unnecessary risk.
+        # Also, if users want concurrent range downloads, gcsfs already handles that
+        # natively through cat_file(start=..., end=...).
+        headers = kwargs.get("headers", {})
+        if headers and any(k.lower() == "range" for k in headers):
+            logger.debug(
+                "Custom Range header detected. Falling back to sequential download."
             )
-        else:
-            u2 = self.url(rpath)
-            await self._get_file_request(u2, lpath, callback=callback, **kwargs)
+            concurrency = 1
+
+        try:
+            if concurrency > 1:
+                await self._get_file_concurrent(
+                    rpath, lpath, concurrency, callback=callback, **kwargs
+                )
+            else:
+                u2 = self.url(rpath)
+                await self._get_file_request(u2, lpath, callback=callback, **kwargs)
+        except BaseException:
+            if os.path.exists(lpath):
+                os.remove(lpath)
+            raise
 
     def _open(
         self,
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1936,12 +1936,136 @@ class GCSFileSystem(asyn.AsyncFileSystem):
             checker.validate_http_response(r)  # validate file consistency
             return r.status, r.headers, r.request_info, data
 
+    async def _get_file_concurrent(
+        self, rpath, lpath, concurrency, headers=None, callback=None, **kwargs
+    ):
+        """
+        Concurrent write the gcs object into the disk.
+        """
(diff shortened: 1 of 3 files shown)


gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+
+        # Prevent silent corruption by pinning the exact object generation
+        generation = details.get("generation")
+        if generation and "generation" not in kwargs:
+            kwargs["generation"] = generation
+
+        callback = callback or NoOpCallback()
+        callback.set_size(total_size)
+
+        self._init_local_file(lpath, total_size)
```
martindurant, line 2050: We should only do this if there will be multiple chunks. On posix, the file init should be fast (but not free), but on windows I believe it still allocates and zero-fills all the space up front, which can be costly.

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+            for t in tasks:
+                if not t.done():
+                    t.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
+            raise e
+
+        if check_consistency:
+            logger.debug(
+                "On-the-fly checksum disabled for concurrent out-of-order downloads. "
+                "Verifying file integrity post-download."
```
martindurant, line 2093: I don't remember how this works (maybe it's only s3), but isn't one of the checksums available calculated directly from the checksums of the parts - as opposed to a running checksum where you need the continuous stream of bytes?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+            await asyncio.gather(*tasks)
+        except BaseException as e:
+            for t in tasks:
+                if not t.done():
+                    t.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
+            raise e
+
+        if check_consistency:
+            logger.debug(
```
martindurant: There is an argument to make this a higher priority logging or even a warning. Perhaps not, since consistency is explicitly requested by the user.

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+        self._init_local_file(lpath, total_size)
+
+        if total_size == 0:
+            if check_consistency:
+                checker = get_consistency_checker(consistency)
+                checker.update(b"")
+                checker.validate_json_response(details)
+            return
+
+        part_size = (total_size + concurrency - 1) // max(1, concurrency)
```
martindurant: Shouldn't there be a minimum part size, perhaps the same as the default block size? (Above, we only check whether the total size is below the threshold to go concurrent; here I mean that perhaps we have fewer chunks than requested)

gcsfs/core.py
martindurant: Consistency _could_ be checked normally if there is only one chunk. Perhaps not worth writing a separate code path for that.

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+                        break
+
+                    local_f.write(chunk)
+                    callback.relative_update(len(chunk))
+                    bytes_written += len(chunk)
+
+                expected_bytes = end_offset - start_offset
+                if bytes_written != expected_bytes:
+                    raise RuntimeError(
+                        f"Connection closed early. Expected {expected_bytes} bytes, got {bytes_written}."
```
martindurant: Include offset in this message - maybe a full download of the file can be avoided on the next try.

gcsfs/tests/test_core.py
```diff
@@ -1468,6 +1428,134 @@ def test_errors(gcs):
         assert "bucket" in str(e)
 
 
+def test_get_file_request_direct(gcs):
```
martindurant, line 1432: We should have at least one check setting concurrency= on the top-level get_file ( or get() ) rather than the specific methods.

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+            )
+
+        try:
+            await asyncio.gather(*tasks)
+        except BaseException as e:
+            for t in tasks:
+                if not t.done():
+                    t.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
+            raise e
```
zhixiangli: Should we remove the lpath before raising an error?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+
+        if check_consistency:
+            logger.debug(
+                "On-the-fly checksum disabled for concurrent out-of-order downloads. "
+                "Verifying file integrity post-download."
+            )
+            await self._verify_file_consistency(lpath, details, consistency)
```
zhixiangli, line 2095: Is it possible for `_verify_file_consistency` to raise an error? Should we also remove the lpath if an exception occurs?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+        end_offset,
+        chunk_size,
+        headers,
+        callback,
+        **kwargs,
+    ):
+        """Fetches a specific byte range of a file and writes it to disk."""
+        req_headers = {"Range": f"bytes={start_offset}-{end_offset - 1}"}
+        if headers:
+            req_headers.update(headers)
```
zhixiangli, line 1965: Is it possible for a user "Range" header to overwrite the chunk range?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+                    if not chunk:
+                        break
+
+                    local_f.write(chunk)
+                    callback.relative_update(len(chunk))
+                    bytes_written += len(chunk)
+
+                expected_bytes = end_offset - start_offset
+                if bytes_written != expected_bytes:
+                    raise RuntimeError(
```
zhixiangli: It looks like a retryable error; should we raise a retryable exception instead?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+
+        bytes_written = 0
+        await self._set_session()
+
+        async with self.session.get(
+            url,
+            headers=self._get_headers(req_headers),
+            params=self._get_params(kwargs),
+            timeout=self.requests_timeout,
+        ) as r:
```
zhixiangli, line 1976: Should we explicitly check if r.status is 206, not just >= 400?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+            validate_response(r.status, None, rpath)
+            local_f = None
+            try:
+                local_f = open(lpath, "rb+")
```
zhixiangli, line 1980: Offload the blocking open, seek, write, etc. to the loop executor?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+                local_f.seek(start_offset)
+                while True:
+                    chunk = await r.content.read(chunk_size)
+                    if not chunk:
+                        break
+
+                    local_f.write(chunk)
```
zhixiangli, line 1989: nit: consider `os.pwrite(fd, chunk, offset)`?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
...
+    ):
+        """Main orchestrator for concurrent file downloads."""
+        consistency = kwargs.pop("consistency", self.consistency)
+        chunk_size = kwargs.pop("chunk_size", 128 * 1024)
+
+        details = await self._info(rpath, **kwargs)
+        total_size = details.get("size", 0)
+
+        if total_size < self.MIN_CHUNK_SIZE_FOR_CONCURRENCY:
+            concurrency = 1
```
zhixiangli, line 2033: If concurrency is 1, should go to `_get_file_request` path?

gcsfs/core.py
```diff
@@ -1936,12 +1936,168 @@ async def _get_file_request(
             checker.validate_http_response(r)  # validate file consistency
             return r.status, r.headers, r.request_info, data
 
+    def _init_local_file(self, lpath, total_size):
+        """Creates the target directory and pre-allocates the file size."""
+        os.makedirs(os.path.dirname(lpath) or os.curdir, exist_ok=True)
+        if total_size == 0:
+            open(lpath, "wb").close()
```
zhixiangli: nit: with open(lpath, "wb"): pass

gcsfs/core.py
zhixiangli: nit: use bare raise preserves the original traceback.

zhixiangli (changes requested):

gcsfs/tests/test_core.py
googlyrahman: Added the test.

gcsfs/core.py
googlyrahman: Done.

gcsfs/core.py
googlyrahman: I don't expect user to come at this method if they're passing the Range Headers, See the newly added check in `_get_file`, but i think it still make sense to assign the header after putting the user header, so i've refactored it.

gcsfs/core.py
googlyrahman: I'm just keeping it consistent with the existing method. If we need explicit check, we need it at both the places.  However i agree, that we should explicitly check for 206 here. I think it is more of defensive programming, and doesn't matter much here.

gcsfs/core.py
googlyrahman: I actually did try it, and it was not giving any gains because of threading overhead, and it was counter productive. I'll see if i can find out the script with the numbers.

gcsfs/core.py
googlyrahman: Keeping it similar to existing code given i ran the perf test, and don't see any performance gains in using `os.pwrtie` as it's a sequential writes. Let me know if there are any benefits of the same.

gcsfs/core.py
googlyrahman: Good catch, it should be a retryable error.

gcsfs/core.py
googlyrahman: Done

gcsfs/core.py
googlyrahman: Hmm, i think, we can fallback to `_get_file_request` here too.

gcsfs/core.py
googlyrahman: I think, now we only enter this path if concurrency>1, so this should be fine now.

gcsfs/core.py
googlyrahman: Doesn't matter much as it only enters this path if file size > 5MB and 5MB is always beneficial for concurrent downloads, but i've kept a minimum part size of chunk size which user can specify.

gcsfs/core.py
googlyrahman: Good catch, and we should. The existing code doesn't remove it, we should add a try-catch method in _get_file method. I'll do this.

googlyrahman: Done

gcsfs/core.py
googlyrahman: Made it warning

googlyrahman: Yeah, this was the rationale, With the recent changes, we don't enter into this path with concurrency=1, we only come in this path if concurrency>1

gcsfs/core.py
googlyrahman: MD5 requires data in strict sequential order, so we cannot compute it in parts. CRC32C can be combined from parts, but adding that custom handling would introduce specific branching for a consistency check that is not the default in gcsfs.

I would consider adding it if CRC32C was the default, but since the default consistency is none (as seen [here](https://github.com/fsspec/gcsfs/blob/main/gcsfs/core.py#L313)), I left it out to keep the code simple.

Let me know if you think it is worth adding the branching logic just for CRC32C. If we go that route, the flow would look like this:

- Each chunk download computes the checksum if the user requested crc32c and returns it to the main task.
- The main task checks if the checksums are returned and if they can be combined. If so, it combines them.
- If the returned checksum is None or the algorithm cannot be computed in parts, the main task falls back to computing the checksum via disk reads.

gcsfs/core.py
googlyrahman: Addressed in previous comment.

gcsfs/core.py
zhixiangli: Got it. It would be great to include comments stating that it doesn't help, just in case others explore the offload in the future.

gcsfs/core.py
zhixiangli: `pwrite` is atomic with respect to the positioning (seeking) and writing operations combined. Defer to your judgment if you think it is necessary.

gcsfs/core.py
googlyrahman: Added a comment linking this conversation, This was the script and numbers

<details>
<summary>Click to expand</summary>

```
margubur@seventhsky:~/gcsfsPRs/scripts$ python3 check.py 
=== BENCHMARK: N Tasks x Task Size (MB) x Chunk Size (KB) ===
-------------------------------------------------------------------------------------
Scenario 1: Mid Concurrency (N=4), Fixed Task Size (1 GB)
Tasks (N):  4 | Task Size: 1024 MB | Chunk:    128 KB | Direct: 4.0146s | Threaded: 5.3792s
Tasks (N):  4 | Task Size: 1024 MB | Chunk:   1024 KB | Direct: 1.1491s | Threaded: 1.3054s
Tasks (N):  4 | Task Size: 1024 MB | Chunk:  10240 KB | Direct: 1.1074s | Threaded: 1.2374s
Tasks (N):  4 | Task Size: 1024 MB | Chunk:  51200 KB | Direct: 1.0996s | Threaded: 1.1090s
Tasks (N):  4 | Task Size: 1024 MB | Chunk: 102400 KB | Direct: 1.0823s | Threaded: 1.1071s

Scenario 2: High Concurrency (N=16), Fixed Task Size (1GB)
Tasks (N): 16 | Task Size: 1024 MB | Chunk:    128 KB | Direct: 6.3792s | Threaded: 24.5034s
Tasks (N): 16 | Task Size: 1024 MB | Chunk:   1024 KB | Direct: 5.9888s | Threaded: 15.1990s
Tasks (N): 16 | Task Size: 1024 MB | Chunk:  10240 KB | Direct: 6.3070s | Threaded: 13.3107s
Tasks (N): 16 | Task Size: 1024 MB | Chunk:  51200 KB | Direct: 5.9474s | Threaded: 17.2603s
Tasks (N): 16 | Task Size: 1024 MB | Chunk: 102400 KB | Direct: 6.8089s | Threaded: 15.3526s

Scenario 3: Low Concurrency (N=1), Fixed Task Size (1GB)
Tasks (N):  1 | Task Size: 1024 MB | Chunk:    128 KB | Direct: 0.4166s | Threaded: 1.8194s
Tasks (N):  1 | Task Size: 1024 MB | Chunk:   1024 KB | Direct: 0.3213s | Threaded: 0.5454s
Tasks (N):  1 | Task Size: 1024 MB | Chunk:  10240 KB | Direct: 0.2969s | Threaded: 0.3629s
Tasks (N):  1 | Task Size: 1024 MB | Chunk:  51200 KB | Direct: 0.2703s | Threaded: 0.2900s
Tasks (N):  1 | Task Size: 1024 MB | Chunk: 102400 KB | Direct: 0.2679s | Threaded: 0.2868s

margubur@seventhsky:~/gcsfsPRs/scripts$ cat check.py 
import asyncio
import os
import shutil
import time

TEST_DIR = "bench_file_io_concurrent"

def init_local_file_sync(lpath, total_size):
    """Pre-allocates the file exactly as your _init_local_file does."""
    os.makedirs(os.path.dirname(lpath) or os.curdir, exist_ok=True)
    if total_size == 0:
        open(lpath, "wb").close()
    else:
        with open(lpath, "wb") as f:
            f.truncate(total_size)

async def benchmark_scenario(n_tasks, task_size_mb, chunk_size_kb):
    task_size_bytes = task_size_mb * 1024 * 1024
    chunk_bytes = chunk_size_kb * 1024
    total_size = n_tasks * task_size_bytes
    
    # Create a dummy payload in memory
    dummy_data = b"x" * chunk_bytes
    
    path_direct = os.path.join(TEST_DIR, f"direct_{n_tasks}N_{task_size_mb}MB_{chunk_size_kb}KB.bin")
    path_threaded = os.path.join(TEST_DIR, f"threaded_{n_tasks}N_{task_size_mb}MB_{chunk_size_kb}KB.bin")

    init_local_file_sync(path_direct, total_size)
    init_local_file_sync(path_threaded, total_size)

    # --- 1. Direct Async Worker ---
    async def worker_direct(start_offset):
        bytes_written = 0
        with open(path_direct, "rb+") as f:
            f.seek(start_offset)
            while bytes_written < task_size_bytes:
                write_len = min(chunk_bytes, task_size_bytes - bytes_written)
                f.write(dummy_data[:write_len])
                bytes_written += write_len
                
                # Simulate the event-loop yield that happens during `await r.content.read()`
                await asyncio.sleep(0) 

    # --- 2. Threaded Async Worker ---
    async def worker_threaded(start_offset):
        bytes_written = 0
        with open(path_threaded, "rb+") as f:
            f.seek(start_offset)
            while bytes_written < task_size_bytes:
                write_len = min(chunk_bytes, task_size_bytes - bytes_written)
                
                # Offload just the blocking write to the thread pool executor
                await asyncio.to_thread(f.write, dummy_data[:write_len])
                bytes_written += write_len
                
                await asyncio.sleep(0)

    # Run Direct Benchmark
    t0 = time.perf_counter()
    await asyncio.gather(*(worker_direct(i * task_size_bytes) for i in range(n_tasks)))
    t_direct = time.perf_counter() - t0

    # Run Threaded Benchmark
    t0 = time.perf_counter()
    await asyncio.gather(*(worker_threaded(i * task_size_bytes) for i in range(n_tasks)))
    t_threaded = time.perf_counter() - t0

    print(f"Tasks (N): {n_tasks:>2} | Task Size: {task_size_mb:>3} MB | Chunk: {chunk_size_kb:>6} KB | Direct: {t_direct:.4f}s | Threaded: {t_threaded:.4f}s")
    os.remove(path_direct)
    os.remove(path_threaded)


async def main():
    if os.path.exists(TEST_DIR):
        shutil.rmtree(TEST_DIR)
    os.makedirs(TEST_DIR)

    try:
        print("=== BENCHMARK: N Tasks x Task Size (MB) x Chunk Size (KB) ===")
        print("-" * 85)
        
        print("Scenario 1: Mid Concurrency (N=4), Fixed Task Size (1 GB)")
        await benchmark_scenario(n_tasks=4, task_size_mb=1024, chunk_size_kb=128) # gcsfs default default
        await benchmark_scenario(n_tasks=4, task_size_mb=1024, chunk_size_kb=1024)
        await benchmark_scenario(n_tasks=4, task_size_mb=1024, chunk_size_kb=10*1024)
        await benchmark_scenario(n_tasks=4, task_size_mb=1024, chunk_size_kb=50*1024)
        await benchmark_scenario(n_tasks=4, task_size_mb=1024, chunk_size_kb=100*1024)
        
        
        
        print("\nScenario 2: High Concurrency (N=16), Fixed Task Size (1GB)")
        await benchmark_scenario(n_tasks=16, task_size_mb=1024, chunk_size_kb=128) # gcsfs default default
        await benchmark_scenario(n_tasks=16, task_size_mb=1024, chunk_size_kb=1024)
        await benchmark_scenario(n_tasks=16, task_size_mb=1024, chunk_size_kb=10*1024)
        await benchmark_scenario(n_tasks=16, task_size_mb=1024, chunk_size_kb=50*1024)
        await benchmark_scenario(n_tasks=16, task_size_mb=1024, chunk_size_kb=100*1024)
        
        print("\nScenario 3: Low Concurrency (N=1), Fixed Task Size (1GB)")
        await benchmark_scenario(n_tasks=1, task_size_mb=1024, chunk_size_kb=128) # gcsfs default default
        await benchmark_scenario(n_tasks=1, task_size_mb=1024, chunk_size_kb=1024)
        await benchmark_scenario(n_tasks=1, task_size_mb=1024, chunk_size_kb=10*1024)
        await benchmark_scenario(n_tasks=1, task_size_mb=1024, chunk_size_kb=50*1024)
        await benchmark_scenario(n_tasks=1, task_size_mb=1024, chunk_size_kb=100*1024)

    finally:
        if os.path.exists(TEST_DIR):
            shutil.rmtree(TEST_DIR)

if __name__ == "__main__":
```

</details>

gcsfs/core.py
googlyrahman: Good point on the atomicity! but since each worker opens its own file descriptor here and then write sequentially, they don't share a file cursor, so we don't have to worry about race conditions between the seek and write operations, or reducing the overhead of (`seek`+`write`) - I am going to keep the current approach so it stays consistent with the rest of the codebase.

googlyrahman:
@zhixiangli, please hold off on reviewing or merging this PR for now. I believe there is a better architectural approach that I initially deferred.

Currently, our gcsfs code duplicates the download logic already present in our streaming (cat_file) implementation. To keep the codebase DRY, it makes more sense to reuse the existing streaming and prefetching logic, and have _get_file simply write that streamed data to disk. This will significantly simplify the code and avoid maintaining two separate download paths for streaming and non-streaming workloads.

I initially hesitated because this approach requires additional refactoring. Specifically, the streaming logic (GCSFile/ZonalFile) currently only exposes a synchronous API, which will not work with the asynchronous _get_file method. Additionally, the streaming implementation doesn't handle headers the same way _get_file does. Despite these challenges, I now think unifying the download logic is worth the effort.

I am not currently adding async support directly to GCSFile or ZonalFile, as that may require broader work from the team. Instead, to fast-track disk download performance, I plan to have _get_file interface directly with the BackgroundPrefetcher. To support this, I have raised a separate PR that introduces the necessary async API for the prefetcher.

See the change i'm proposing in the prefetcher: https://github.com/fsspec/gcsfs/pull/877

martindurant:
> To keep the codebase DRY, it makes more sense to reuse the existing streaming and prefetching logic, and have _get_file simply write that streamed data to disk. This will significantly simplify the code and avoid maintaining two separate download paths for streaming and non-streaming workloads.

We might want to have some conversation around this. The download logic is different in that:
- we know beforehand everything we will be reading, so no need to measure the read pattern (you could pre-set the read record queue)
- there should be no need for buffer copies, everything can be written as it comes
- the order of arriving data is unimportant (caveat: the pre-building of a zero-filled file on windows versus sparse files on posix)

So the only feature we reuse is the worker offloading.

You do make a very good point about code complexity. AbstractFileSystem.get_file (sync version, of course) essentially does what you want.

The proof will be in the benchmarks: can you really saturate the stream, and would using the file/prefetcher even end up faster?

googlyrahman:
> we know beforehand everything we will be reading, so no need to measure the read pattern (you could pre-set the read record queue)

The read pattern logic in the prefetcher is lightweight. It just calculates a rolling average, so it does not act as a bottleneck.

> there should be no need for buffer copies, everything can be written as it comes

Users can set the chunk_size based on their needs. The call eventually goes to _cat_file, which returns the bytes object that we then write to disk. If the chunk_size is too low, opening the stream repeatedly will backfire. If the chunk_size is high, buffer copies are unavoidable. The ZonalFile handles this pretty good by preallocating the space in `DirectMemmoveBuffer`, and re-using the same streams.

> the order of arriving data is unimportant (caveat: the pre-building of a zero-filled file on windows versus sparse files on posix)

While I agree the order is unimportant, the prefetcher produces the data sequentially. I think this will also make it easier to compute the checksum since we will not need a second pass after the data is written to disk, unlike the approach for this PR. To prevent head-of-line blocking, we process four chunks at a time. You can check the code to see exactly how this is handled.

Overall, the only downside I see with the new approach is memory usage. Earlier, we wrote data as soon as it arrived in 128KB chunks. Now, we are handling it in 16MB chunks. This increases the memory footprint slightly, but I think the trade-off is worth it for the concurrency > 1 path since it boosts throughput and keeps the code simple.

The concurrency = 1 path remains exactly the same as before.

This PR:
- Lower memory footprint
- Introduce repetitive logic of download

The alternative PR:
- Comparatively higher memory usage
- No repeated logic (if you look at the zonal integration, it is just a few lines of change)

The throughput and memory details are shared in the PR description: https://github.com/fsspec/gcsfs/pull/877/. I am seeing 10x gains for zonal buckets and up to 2-3x gains for standard buckets compared to gcsfs HEAD.

martindurant:
> make it easier to compute the checksum since we will not need a second pass after the data is written to disk,

Good point.

>  I am seeing 10x gains for zonal buckets and up to 2-3x gains for standard buckets compared to gcsfs HEAD.

That's proof enough for me. Is this actually near saturating the bandwidth?

googlyrahman:
> That's proof enough for me. Is this actually near saturating the bandwidth?

Yes, it almost saturates the bandwidth for both Standard and Zonal buckets. 

- The Standard buckets provide approximately 600–700 MB/s of RAM-based download throughput, and the disk download speed is comparable. 
- The Zonal bucket provides around 2.2 GB/s of throughput, and the disk-based download speed is close to that at approximately 1.9 GB/s

googlyrahman:
Closing this in favour of https://github.com/fsspec/gcsfs/pull/877

PR #864, opened 2026-06-05
Bypass redundant fs.info() HTTP calls during ZonalFile initialization

yuxin00j:
This PR optimizes the initialization of `ZonalFile` (Zonal Buckets) by eliminating a redundant `HTTP GET` request that was being fired to fetch object metadata.

Previously, instantiating a `ZonalFile` for a read stream would trigger an `fs.info(self.path)` call during the parent `GCSFile` instantiation. This PR patches this behavior by directly mapping the `object_metadata` protobuf received from the underlying gRPC stream into `self._details` prior to invoking the `super().__init__()` method.

**Changes:**
- Preemptively populates `self._details` in `ZonalFile.__init__` by extracting the raw gRPC `object_metadata`.
- Introduces `_extract_metadata_dict` using `google.protobuf.json_format.MessageToDict` to map the gRPC protobuf into a dictionary that perfectly mimics the standard REST API JSON response.
- Implements strict format parity to ensure seamless `fsspec` compatibility (base64 encoding the gRPC `crc32c` integer, flattening `checksums`, and mapping legacy timestamp names).
- Adds robust unit test coverage to verify metadata extraction, checksum conversion, and fallback logic if protobuf parsing fails.

**Impact:**
- Reduces latency on zonal bucket reads by trimming an entire network round-trip.
- Ensures downstream applications reading Zonal files still have access to full object metadata (e.g., `storageClass`, `timeCreated`) without needing secondary HTTP calls.

### DIFF
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -39,9 +39,14 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
     Wraps Google API errors into standard Python exceptions.
     """
     try:
-        return await AsyncMultiRangeDownloader.create_mrd(
+        mrd = await AsyncMultiRangeDownloader.create_mrd(
             grpc_client, bucket_name, object_name, generation
         )
+
+        # Extract metadata from the updated Python SDK
+        read_obj_str = getattr(mrd, "read_obj_str", None)
+        mrd.object_metadata = getattr(read_obj_str, "object_metadata", None)
+        return mrd
     except NotFound:
         # We wrap the error here to match standard Python error handling
         # and avoid leaking Google API exceptions to users.
@@ -479,6 +484,7 @@ class MRDPool:
         self._active_count = 0
         self._lock = asyncio.Lock()
         self.persisted_size = None
+        self.object_metadata = None
         self._initialized = False
         self._closed = False
 
@@ -541,6 +547,7 @@ class MRDPool:
                 mrd = await self._create_mrd()
                 self._all_mrds.append(mrd)
                 self.persisted_size = mrd.persisted_size
+                self.object_metadata = getattr(mrd, "object_metadata", None)
                 self._free_mrds.put_nowait(mrd)
                 self._active_count += 1
 
--- a/gcsfs/zonal_file.py
+++ b/gcsfs/zonal_file.py
@@ -68,6 +68,7 @@ class ZonalFile(GCSFile):
         self.gcsfs = gcsfs
         self.pool_size = pool_size
         object_size = None
+        self._details = None
         if "r" in self.mode:
             self.mrd_pool = asyn.sync(
                 self.gcsfs.loop,
@@ -78,6 +79,12 @@ class ZonalFile(GCSFile):
                 self.pool_size,
             )
             object_size = self.mrd_pool.persisted_size
+
+            if getattr(self.mrd_pool, "object_metadata", None):
+                raw_details = _extract_metadata_dict(
+                    self.mrd_pool.object_metadata, bucket, key
+                )
+                self._details = self.gcsfs._process_object(bucket, raw_details)
 
             if object_size is None:
                 logger.warning(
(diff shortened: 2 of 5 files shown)


gcsfs/zonal_file.py
```diff
@@ -79,6 +80,12 @@ def __init__(
             )
             object_size = self.mrd_pool.persisted_size
 
+            if (
+                hasattr(self.mrd_pool, "object_metadata_dict")
+                and self.mrd_pool.object_metadata_dict
+            ):
+                self._details = self.mrd_pool.object_metadata_dict
```
zhixiangli: Any in-place mutation of one file's _details corrupts the pool's copy and every sibling file. Copy on assignment: self._details = dict(self.mrd_pool.object_metadata_dict).

gcsfs/zb_hns_utils.py
```diff
@@ -541,6 +559,8 @@ async def initialize(self):
                 mrd = await self._create_mrd()
                 self._all_mrds.append(mrd)
                 self.persisted_size = mrd.persisted_size
+                if hasattr(mrd, "object_metadata_dict"):
```
zhixiangli: Initialize self.object_metadata_dict = None in __init__ to avoid hasattr

gcsfs/zonal_file.py
```diff
@@ -79,6 +80,12 @@ def __init__(
             )
             object_size = self.mrd_pool.persisted_size
 
+            if (
+                hasattr(self.mrd_pool, "object_metadata_dict")
```
zhixiangli: Initialize object_metadata_dict in MRDPool to avoid hasattr

gcsfs/zb_hns_utils.py
```diff
@@ -39,9 +39,27 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
...
+        read_obj_str = getattr(mrd, "read_obj_str", None)
+        obj_meta = getattr(read_obj_str, "object_metadata", None)
+        if obj_meta:
+            mrd.object_metadata_dict = {
+                "name": getattr(obj_meta, "name", None),
+                "size": getattr(obj_meta, "size", None),
+                "generation": (
+                    str(getattr(obj_meta, "generation", None))
+                    if getattr(obj_meta, "generation", None)
+                    else None
```
zhixiangli:
```
str(gen) if (gen := getattr(obj_meta, "generation", None)) is not None else None

File: gcsfs/zb_hns_utils.py [hunk 5]
```diff
@@ -39,9 +39,27 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
...
+        mrd = await AsyncMultiRangeDownloader.create_mrd(
             grpc_client, bucket_name, object_name, generation
         )
+
+        # Extract metadata from the updated Python SDK
+        read_obj_str = getattr(mrd, "read_obj_str", None)
+        obj_meta = getattr(read_obj_str, "object_metadata", None)
+        if obj_meta:
+            mrd.object_metadata_dict = {
+                "name": getattr(obj_meta, "name", None),
```
zhixiangli: What's the name? Should it be object key or full path?

gcsfs/zb_hns_utils.py
```diff
@@ -39,9 +39,27 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
     Wraps Google API errors into standard Python exceptions.
     """
     try:
-        return await AsyncMultiRangeDownloader.create_mrd(
+        mrd = await AsyncMultiRangeDownloader.create_mrd(
             grpc_client, bucket_name, object_name, generation
         )
+
+        # Extract metadata from the updated Python SDK
```
zhixiangli, line 46: Is the Python SDK released yet? If yes, we can bump the Python SDK dependency in gcsfs to >= [version number].

zhixiangli (changes requested):

gcsfs/zonal_file.py
yuxin00j: Done.

gcsfs/zb_hns_utils.py
yuxin00j: Updated dependencies to "google-cloud-storage>=3.11.0"

gcsfs/zb_hns_utils.py
yuxin00j: name is the bucket_name/path_to_object.

gcsfs/zb_hns_utils.py
yuxin00j: Done.

gcsfs/zb_hns_utils.py
yuxin00j: Done. This change is removed.

gcsfs/zonal_file.py
yuxin00j: This change is removed.

PR #869, opened 2026-06-08
Refactor: Create flat bucket dynamically in tests

zhixiangli:
### Summary

Refactors the test setup for mixed HNS/flat bucket integration tests. Instead of relying on a pre-created standard (flat) bucket passed via the `GCSFS_FLAT_TEST_BUCKET` environment variable, the tests now dynamically create a temporary flat bucket during the test session.

### Details

*   **Dynamic Bucket Creation**: Introduced a session-scoped fixture `flat_bucket_name` in `gcsfs/tests/conftest.py` that creates a standard bucket with a unique name (`gcsfs-test-flat-{uuid}`) and registers it for deletion at the end of the session.
*   **Fixture Update**: Updated `flat_bucket` fixture to use `flat_bucket_name` and handle cleanup before/after each test.
*   **Cleanup**: Removed `TEST_FLAT_BUCKET` from `gcsfs/tests/settings.py` and its usages, as it is no longer needed.
*   **CI Configuration**: Removed the need to manage dedicated flat buckets in `cloudbuild/create_resources.sh`, `cloudbuild/cleanup.sh`, and `cloudbuild/run_tests.sh`.

### DIFF
--- a/gcsfs/tests/conftest.py
+++ b/gcsfs/tests/conftest.py
@@ -20,7 +20,6 @@ from gcsfs import GCSFileSystem
 from gcsfs.extended_gcsfs import BucketType
 from gcsfs.tests.settings import (
     TEST_BUCKET,
-    TEST_FLAT_BUCKET,
     TEST_HNS_BUCKET,
     TEST_HNS_REQUESTER_PAYS_BUCKET,
     TEST_PROJECT,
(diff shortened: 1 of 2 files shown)


cloudbuild/run_tests.sh
```diff
@@ -58,6 +59,7 @@ case "$TEST_SUITE" in
     export GCSFS_HNS_TEST_BUCKET="gcsfs-test-hns-${SHORT_BUILD_ID}"
     export GCSFS_TEST_REQ_PAYS_BUCKET="gcsfs-test-hns-req-pay-${SHORT_BUILD_ID}"
     export GCSFS_HNS_TEST_REQ_PAYS_BUCKET="gcsfs-test-hns-req-pay-${SHORT_BUILD_ID}"
+    export GCSFS_FLAT_TEST_BUCKET="gcsfs-test-standard-${SHORT_BUILD_ID}"
```
Mahalaxmibejugam: We run standard and HNS test suites in parallel in the cloudBuild pipeline, using the same bucket for both these test suites would lead to race conditions and failures.

Mahalaxmibejugam (changes requested):

cloudbuild/run_tests.sh
```diff
@@ -38,6 +38,7 @@ case "$TEST_SUITE" in
     export GCSFS_ZONAL_TEST_BUCKET="gcsfs-test-zonal-${SHORT_BUILD_ID}"
     export GCSFS_HNS_TEST_BUCKET="gcsfs-test-zonal-${SHORT_BUILD_ID}"
     export GCSFS_HNS_TEST_REQ_PAYS_BUCKET="gcsfs-test-zonal-${SHORT_BUILD_ID}"
+    export GCSFS_FLAT_TEST_BUCKET="gcsfs-test-standard-for-zonal-flat-${SHORT_BUILD_ID}"
```
Mahalaxmibejugam: Can we use the existing gcsfs-test-standard-for-zonal-${SHORT_BUILD_ID} bucket here? This is standard bucket which works for mixed flat/HNS test and also used only within the zonal test suite, so there would be no issues like race conditions which can happen in HNS tests as mentioned in the other comment.

Mahalaxmibejugam:
Considering these changes, I believe it is more appropriate to create the bucket within the test itself rather than introducing a new environment variable. There are many tests already which create their own bucket if not required globally across the test suite. 

This avoids increasing the CI pipeline's complexity by requiring additional bucket creation and deletion for a single edge-case test. Furthermore, running the integration test suite now requires the unnecessary extra step of creating a bucket and exporting it via an additional environment variable.

gcsfs/tests/conftest.py
```diff
@@ -500,28 +499,31 @@ def gcs_hns(gcs_factory, buckets_to_delete):
...
+    return bucket_name
 
-    In the HNS test suite TEST_BUCKET and TEST_HNS_BUCKET may both point at
-    HNS-enabled buckets, so tests that need to exercise genuinely mixed
-    HNS/flat behavior cannot rely on TEST_BUCKET being flat. This fixture
-    creates a standard bucket (HNS disabled), registers it for end-of-session
-    cleanup, and cleans its contents before and after the test.
+
+@pytest.fixture
+def flat_bucket(flat_bucket_name, gcs_factory):
```
Mahalaxmibejugam, line 518: As flat_bucket_name is brand new bucket for each test, do we need to perform the cleanup before and after the test?

zhixiangli: It's session scope,
```
  @pytest.fixture(scope="session")
  def flat_bucket_name(...):

Mahalaxmibejugam (approved):

PR #870, opened 2026-06-09
Refactor directory cache updates and optimize caching logic

zhixiangli:
This PR refactors the directory cache (`dircache`) update logic in `gcsfs` to centralize caching strategies and optimizes cache invalidation.

### Motivation and Context
Previously, cache invalidation logic was interleaved with I/O operations in `gcsfs/core.py` and `gcsfs/extended_gcsfs.py`, leading to code duplication. Additionally, the default behavior of invalidating all ancestor directories up to the bucket root is redundant for HNS-enabled buckets (where directories are first-class persistent objects) and can be optimized for write operations on all buckets.

This PR addresses these issues by:
1. Centralizing cache update logic into a new module `gcsfs/_dircache.py`.
2. Optimizing cache updates for write operations across all bucket types (standard and HNS) to avoid broad ancestor invalidation when the parent is already cached.
3. Optimizing HNS cache updates for deletion and move operations to perform targeted in-place mutations instead of broad ancestor invalidations.
4. Fixing batch deletion (`_rm_files`) to only update the cache for successfully deleted or confirmed absent files, preventing eviction of failed deletes.

#### 1. Centralized Cache Update Strategies (`gcsfs/_dircache.py`)
- **`DirCacheUpdater`**: The base strategy (used by `GCSFileSystem`).
  - **Writes**: Invalidates only the immediate parent listing if it is already cached (since we know the parent exists, adding a file cannot implicitly create new intermediate directories that would be hidden by a stale ancestor cache). If the parent is not cached, it falls back to broad ancestor invalidation.
  - **Deletes**: Broadly invalidates the affected parent and all of its ancestors.
  - **Moves**: Broadly invalidates the affected parent and all of its ancestors for both source and destination.
- **`HnsDirCacheUpdater`** (subclass of `DirCacheUpdater` used by `ExtendedGcsFileSystem`): A targeted strategy for HNS buckets.
  - **Writes**: Inherits the optimized write behavior from `DirCacheUpdater`.
  - **Deletes**: Mutates parent listings in place by removing deleted entries, avoiding ancestor invalidation entirely. Grouped by parent for efficiency.
  - **Moves**: For intra-bucket moves, removes the source from the parent cache and upserts the destination into the parent cache in place.
  - Non-HNS paths fallback to `DirCacheUpdater` behavior (broad invalidation).

#### 2. Filesystem Refactoring
- **`gcsfs/core.py` (`GCSFileSystem`)**:
  - Inherits from `DirCacheUpdater`.
  - Replaced direct `invalidate_cache` calls in `_mv_file`, `_rm_file`, `_pipe_file`, and `_put_file` with calls to the new cache updater hooks (`_write_file_cache_update`, `_rm_file_cache_update`).
  - Updated `_rm_files` (batch delete) to only update the cache for successfully deleted or confirmed absent (404) files, preventing eviction of failed deletes.
- **`gcsfs/extended_gcsfs.py` (`ExtendedGcsFileSystem`)**:
  - Inherits from `HnsDirCacheUpdater`.
  - Removed overridden `_mv_file_cache_update` and `_update_dircache_after_rename` as they are now handled by `HnsDirCacheUpdater`.
  - Refactored `_mkdir` and `_rmdir` to use low-level cache primitives (`_cache_add_entry`, `_cache_drop_entries`) defined in `HnsDirCacheUpdater`.

#### 3. Testing
- Added comprehensive unit and integration tests to verify:
  - Targeted HNS deletion cache updates (in-place mutation, no ancestor invalidation).
  - Optimized write cache updates (invalidating only immediate parent when cached, falling back to broad invalidation when uncached) for both HNS and standard buckets.
  - Low-level cache helpers (`_cache_drop_entries`, `_cache_add_entry`).
- Updated existing tests to align with the new caching behavior.

### DIFF
--- a/gcsfs/_dircache.py
+++ b/gcsfs/_dircache.py
@@ -6,14 +6,15 @@ consistent with the bucket. This module groups that concern into two mixins so
 the bookkeeping lives in one place rather than being interleaved with the I/O
 in :mod:`gcsfs.core` and :mod:`gcsfs.extended_gcsfs`:
 
-- :class:`DirCacheUpdater` -- the default strategy. Broadly invalidates the
-  affected parent directory and all of its ancestors, so the next listing
-  re-reads them from the bucket.
+- :class:`DirCacheUpdater` -- the default strategy. For deletes and moves it
+  broadly invalidates the affected parent directory and all of its ancestors, so
+  the next listing re-reads them from the bucket. For writes it invalidates only
+  the immediate parent when that parent is already cached (and therefore known to
+  exist), falling back to broad ancestor invalidation otherwise.
 - :class:`HnsDirCacheUpdater` -- a targeted strategy for Hierarchical Namespace
   (HNS) buckets, where directories are first-class persistent objects. It
-  mutates parent listings in place for deletes and moves, and invalidates only
-  the immediate parent listing for writes, falling back to the broad strategy
-  for non-HNS paths.
+  mutates parent listings in place for deletes and moves, falling back to the
+  broad strategy for non-HNS paths.
 
 Both are mixed into a filesystem class and rely on the host class for
 ``dircache``, ``split_path``, ``_parent``, ``_strip_protocol``,
@@ -24,7 +25,12 @@ import asyncio
 
 
 class DirCacheUpdater:
-    """Default ``dircache`` update strategy: broad ancestor invalidation.
+    """Default ``dircache`` update strategy.
+
+    Deletes and moves broadly invalidate the affected parent and all of its
+    ancestors; writes invalidate only the immediate parent when it is already
+    cached (and therefore known to exist), otherwise they fall back to broad
+    ancestor invalidation.
 
     Mixed into :class:`gcsfs.core.GCSFileSystem`. The methods are the hooks that
     the mutating I/O paths (``_rm_file``, ``_rm_files``, ``_put_file``, ...) call
@@ -47,9 +53,28 @@ class DirCacheUpdater:
             self.invalidate_cache(parent)
 
     async def _write_file_cache_update(self, path):
-        # A file was created or overwritten at ``path`` (via put/pipe/cp). The
-        # default behavior invalidates the parent and all of its ancestors.
-        self.invalidate_cache(self._parent(path))
+        # A file was created or overwritten at ``path`` (via put/pipe/cp).
+        #
+        # When the immediate parent's listing is already cached, we have listed
+        # it before, so it already exists and adding a file inside it cannot
+        # create a new directory in any ancestor's listing. Invalidate only that
+        # immediate parent so the new file is picked up on its next listing.
+        #
+        # When the parent is not cached, this write may have implicitly created
+        # it (and intermediate directories) -- flat buckets simulate directories
+        # from object prefixes, and HNS buckets auto-create missing parents on
+        # object write. Leaving a cached ancestor untouched would hide the new
+        # directory, so fall back to invalidating the parent and all ancestors.
+        #
+        # Like the rest of this module, this relies on the dircache being
+        # consistent with the bucket (this client is the sole mutator between
+        # listings); concurrent external mutations are reconciled only by
+        # ``invalidate_cache`` / listings expiry, not here.
+        parent = self._parent(path)
+        if parent in self.dircache:
+            self.dircache.pop(parent, None)
+        else:
+            self.invalidate_cache(parent)
 
 
 class HnsDirCacheUpdater(DirCacheUpdater):
@@ -57,10 +82,10 @@ class HnsDirCacheUpdater(DirCacheUpdater):
 
     Mixed into :class:`gcsfs.extended_gcsfs.ExtendedGcsFileSystem`. For HNS
     buckets directories are persistent objects, so deletes and moves usually
-    only change the contents of the immediate parent's listing. Writes
-    invalidate the immediate parent listing without walking every ancestor up to
-    the bucket root. Non-HNS / cross-bucket paths defer to
-    :class:`DirCacheUpdater` via ``super()``.
+    only change the contents of the immediate parent's listing. Non-HNS /
+    cross-bucket paths defer to :class:`DirCacheUpdater` via ``super()``. The
+    write path is bucket-type-agnostic and handled entirely by
+    :meth:`DirCacheUpdater._write_file_cache_update`.
     """
 
     def _cache_drop_entries(self, parent, names):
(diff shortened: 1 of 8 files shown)


martindurant:
Question for this and #871: in the hierarchical case, don't we know exactly what changes should happen with the directory cache? We should be able to make those change without having to invalidate anything at all.

(Side note: where we have the full set of listings above some key in the standard buckets case, we also don't have to invalidate all the way, we should know which levels are affcted; but this is more logic/work to implement.)

zhixiangli:
> Question for this and #871: in the hierarchical case, don't we know exactly what changes should happen with the directory cache? We should be able to make those change without having to invalidate anything at all.
> 
> (Side note: where we have the full set of listings above some key in the standard buckets case, we also don't have to invalidate all the way, we should know which levels are affcted; but this is more logic/work to implement.)

- For HNS buckets (hierarchical case): We now avoid invalidating parent/ancestor cache directories entirely.
rm (this branch): _rm_files_cache_update updates the immediate parent's cached directory listing in place by removing the deleted entries. It doesn't trigger any invalidations.

- For standard buckets (flat case): We fall back to the base GCSFileSystem behavior of invalidating the parent and its ancestors. Because directory structures in standard buckets are simulated (using empty placeholders or common prefixes), deleting the last file under a prefix destroys the pseudo-directory. Updating these in place client-side is much more complex since we would have to query GCS or maintain local state to determine if other files still share the same prefix.

Mahalaxmibejugam:
Within the FileSystem class, numerous methods such as put_file, cp_file, pipe_file, and rm_file currently trigger a full cache invalidation.

Rather than repeating logic by adding custom directory cache mutations directly within individual deletion and move methods, can we centralize the low-level cache operations. Since cache update logic exists for mv_file and some parts of it is duplicated in this PR for rm_file, can we keep the implementation DRY by maximizing code reuse?

martindurant:
No opposition to that

zhixiangli:
> Within the FileSystem class, numerous methods such as put_file, cp_file, pipe_file, and rm_file currently trigger a full cache invalidation.
> 
> Rather than repeating logic by adding custom directory cache mutations directly within individual deletion and move methods, can we centralize the low-level cache operations. Since cache update logic exists for mv_file and some parts of it is duplicated in this PR for rm_file, can we keep the implementation DRY by maximizing code reuse?

@Mahalaxmibejugam @martindurant That would be scope creep. I think it's safer and cleaner to implement this method by method, as it keeps the PR small and easier to review. We can still keep the code DRY through future refactoring.

Mahalaxmibejugam:
> @Mahalaxmibejugam @martindurant That would be scope creep. I think it's safer and cleaner to implement this method by method, as it keeps the PR small and easier to review. We can still keep the code DRY through future refactoring.

We can create the cache utility for HNS buckets in this PR and apply it to the existing rm_file and mv_file methods, leaving other such as pipe_file and put_file for a future PR. This avoids code duplication while keeping the scope of the additional methods for future PRs.

gcsfs/tests/integration/test_extended_hns.py
Mahalaxmibejugam, line 1: Should we add a test for write method's cache invalidation as we have modified the logic for the same?

gcsfs/_dircache.py
```diff
@@ -0,0 +1,242 @@
...
+
+    async def _rm_files_cache_update(self, paths):
+        parents = set(self._parent(p) for p in paths) | set(paths)
+        for parent in parents:
+            self.invalidate_cache(parent)
+
+    async def _write_file_cache_update(self, path):
+        # A file was created or overwritten at ``path`` (via put/pipe/cp). The
+        # default behavior invalidates the parent and all of its ancestors.
+        self.invalidate_cache(self._parent(path))
```
Mahalaxmibejugam: For standard buckets, I think we don't have to invalidate all ancestors during a write operation. We currently perform a full invalidation because we cannot determine if the affected file is the last one in the directory path. If it is, the entire directory structure might be removed from bucket, necessitating a full cache clear.

However, for writes, couldn't we simply update the parent cache instead of invalidating the entire ancestor chain?

gcsfs/_dircache.py
```diff
@@ -0,0 +1,242 @@
...
+        This involves three main steps:
+        1. Removing the source folder and all its descendants from the cache.
+        2. Removing the source folder's entry from its parent's listing.
+        3. Adding the new destination folder's entry to its parent's listing.
+
+        Args:
+            path1 (str): The source path that was renamed.
+            path2 (str): The destination path.
+        """
+        # 1. Find and remove all descendant paths of the source from the cache.
```
Mahalaxmibejugam, line 152: Let's strip the protocol from these paths at the start of the method, otherwise, since the key in the dircache doesn't store the protocol, subsequent calls to pop and startswith will fail. Should we also add a test case covering this use case?

gcsfs/_dircache.py
Mahalaxmibejugam: Or are we not adding an entry in the cache after writes(looks like same for HNS and standard) because of insufficient information?

Mahalaxmibejugam (changes requested):

gcsfs/_dircache.py
Mahalaxmibejugam, line 1: Should we add a test file for this newly added dircache.py?

gcsfs/tests/integration/test_extended_hns.py
```diff
@@ -1024,6 +1024,119 @@ def test_rm_mixed_buckets_order(self, gcs_hns, flat_bucket, order):
...
+            base_dir in gcsfs.dircache
+        ), "Ancestor directory should not be invalidated on write."
+        assert (
+            sibling_dir in gcsfs.dircache
+        ), "Sibling directory cache should not be invalidated on write."
+
+        # The new file is visible once the parent is re-listed.
+        assert new_file in gcsfs.ls(parent_dir, detail=False)
+
+    def test_write_file_cache_invalidates_only_immediate_parent_standard(
```
Mahalaxmibejugam: Can we move this test to test_core.py which is specific for the standard buckets keeping this file for HNS related tests?

gcsfs/_dircache.py
Mahalaxmibejugam: Adding a reply here:

Changes are made in latest commit for standard buckets to keep the logic same as HNS buckets to invalidate the parent if it exists in the dircache(which confirms that parent_dir exists from before) and invalidates all ancestors if the parent doesn't have any entry in dircache.

gcsfs/tests/test_extended_hns_gcsfs.py
```diff
@@ -671,6 +671,476 @@ async def test_mv_file_cache_update_fallback(self, gcs_hns, gcs_hns_mocks):
...
+
+        await fs._write_file_cache_update(path)
+
+        # Broad invalidation: cached ancestors are popped so a re-list picks up
+        # any directory the write implicitly created.
+        assert grandparent not in fs.dircache
+        assert TEST_HNS_BUCKET not in fs.dircache
+
+    @pytest.mark.asyncio
+    async def test_pipe_file_zonal_routes_through_write_cache_update(self):
```
Mahalaxmibejugam: Can we move this to a zonal-specific test file? We currently maintain separate test files for standard, zonal, and HNS buckets. Common tests for all bucket types typically reside in test_core.py, while Zonal and HNS-specific tests are kept in their respective files.

Same for other zonal tests in this file

Mahalaxmibejugam (approved):

PR #872, opened 2026-06-09
Bypass fs.info during file open by extracting metadata from gRPC

yuxin00j:
This PR significantly reduces latency for Zonal Bucket reads by bypassing redundant HTTP REST API calls (`fs.info`) during `fs.open()`.

### Key Changes
* **gRPC Metadata Extraction:** Instead of triggering an extra HTTP request, we use `_extract_metadata_dict` to convert the gRPC object protobuf directly into the exact JSON dictionary format expected by `fsspec` internals.
* **Timestamp Normalization:** Includes defensive handling for older Python versions (< 3.11) where `datetime.fromisoformat` fails when encountering timestamp strings with nanosecond precision returned by the gRPC API.
* **Flaky Test Fix:** Resolves a GC timing flake in `test_extended_gcsfs_unit.py` where unclosed mocked caches leaked into subsequent tests.

### ⚠️ Dependency Note: `google-cloud-storage`
This PR intentionally omits bumping the `google-cloud-storage` dependency to `>=3.12.0`. 

The upstream PR (`googleapis/google-cloud-python#17261`) that exposes the required `_details` attribute on the gRPC file object was merged *after* the `3.11.0` release. Forcing a dependency bump now would break environments until Google cuts a new release. 

To ensure backwards compatibility without blocking this optimization, we gracefully degrade to the traditional HTTP `fs.info` metadata fetch if `f._details` is unavailable on the client's current `google-cloud-storage` version. **We will bump the dependency to officially support this feature exclusively once the next version of `google-cloud-storage` is released.**

### DIFF
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -43,9 +43,6 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
             grpc_client, bucket_name, object_name, generation
         )
 
-        # Extract metadata from the updated Python SDK
-        read_obj_str = getattr(mrd, "read_obj_str", None)
-        mrd.object_metadata = getattr(read_obj_str, "object_metadata", None)
         return mrd
     except NotFound:
         # We wrap the error here to match standard Python error handling
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -39,9 +39,14 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
     Wraps Google API errors into standard Python exceptions.
     """
     try:
-        return await AsyncMultiRangeDownloader.create_mrd(
+        mrd = await AsyncMultiRangeDownloader.create_mrd(
             grpc_client, bucket_name, object_name, generation
         )
+
+        # Extract metadata from the updated Python SDK
+        read_obj_str = getattr(mrd, "read_obj_str", None)
+        mrd.object_metadata = getattr(read_obj_str, "object_metadata", None)
+        return mrd
     except NotFound:
         # We wrap the error here to match standard Python error handling
         # and avoid leaking Google API exceptions to users.
@@ -479,6 +484,7 @@ class MRDPool:
         self._active_count = 0
         self._lock = asyncio.Lock()
         self.persisted_size = None
+        self.object_metadata = None
         self._initialized = False
         self._closed = False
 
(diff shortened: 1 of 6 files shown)


martindurant:
Why the calls to `gc.collect()` ?

gcsfs/zonal_file.py
```diff
@@ -79,6 +80,12 @@ def __init__(
             )
             object_size = self.mrd_pool.persisted_size
 
+            if getattr(self.mrd_pool, "object_metadata", None):
```
martindurant, line 83: What happens if false?

gcsfs/zb_hns_utils.py
```diff
@@ -39,9 +39,14 @@ async def init_mrd(grpc_client, bucket_name, object_name, generation=None):
     Wraps Google API errors into standard Python exceptions.
     """
     try:
-        return await AsyncMultiRangeDownloader.create_mrd(
+        mrd = await AsyncMultiRangeDownloader.create_mrd(
             grpc_client, bucket_name, object_name, generation
         )
+
+        # Extract metadata from the updated Python SDK
+        read_obj_str = getattr(mrd, "read_obj_str", None)
```
chandra-siri: nit: mrd will always have `read_obj_str` attribute.

gcsfs/zonal_file.py
```diff
@@ -393,3 +400,90 @@ def close(self):
...
+    """
+    try:
+        from google.protobuf.json_format import MessageToDict
+
+        # Use preserving_proto_field_name=False to output camelCase JSON keys
+        details = MessageToDict(obj_meta._pb, preserving_proto_field_name=False)
+    except Exception:
+        # If MessageToDict fails (e.g. proto version mismatch), fallback to an empty
+        # dict and rely on the manual population below for minimum required fields.
+        details = {}
```
zhixiangli, line 430: Add a logger.debug(...)/logger.warning(...) in the handler so  the fallback is observable.

gcsfs/zb_hns_utils.py
zhixiangli, line 738: For the typical fs.open("bucket/key") (no explicit generation), this fires an fs._info HTTP request before the MRD is created. Is this reusable?

This logic is newly introduced ~2 weeks ago.

gcsfs/zonal_file.py
yuxin00j: gcsfs handles this by triggering an HTTP info() call later in GCSFile initialization, which is the current behaviour.

gcsfs/zb_hns_utils.py
yuxin00j: Got it, thanks for pointing this out! Since we have to call _info anyway, this PR will not improve the performance. I'll close this then. I've created #874 to do an optimization to avoid the duplicate metadata call.  cc @chandra-siri

PR #874, opened 2026-06-10
perf: cache metadata details from MRDPoolCache info lookup to avoid redundant network calls

yuxin00j:
When `MRDPoolCache.get()` is called without a `generation` (e.g., during `ZonalFile` initialization), it performs a network request via `fs._info()` to discover the object's latest generation. 

Previously, this fetched metadata was only used to extract the `generation` integer and the rest of the dictionary was discarded. This PR updates `MRDPoolCache.get()` to accept an optional `file_obj` argument. When provided, it directly assigns the fetched info dictionary to `file_obj._details`. 

This optimization ensures that the `file_obj` caches the object metadata immediately, avoiding a redundant metadata network request later in the file's lifecycle.

## Changes
- **`gcsfs.zb_hns_utils.MRDPoolCache.get`**: Added an optional `file_obj` parameter. If an info fetch is required to determine the generation, the full info dictionary is cached on `file_obj._details`.
- **`gcsfs.zonal_file.ZonalFile`**: Passes `self` to `mrd_pool_cache.get()` during initialization.
- **`gcsfs.tests.test_zonal_file.py`**: Added the `test_mrd_pool_cache_sets_file_details` unit test to verify this behavior.

### DIFF
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -478,6 +478,7 @@ class MRDPool:
         self._free_mrds = asyncio.Queue(maxsize=pool_size)
         self._active_count = 0
         self._lock = asyncio.Lock()
+        self.details = None
         self.persisted_size = None
         self._initialized = False
         self._closed = False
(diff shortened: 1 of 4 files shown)


gcsfs/zb_hns_utils.py
```diff
@@ -711,7 +711,7 @@ def _decref(self, key):
             mrds_to_close.extend(_drain_queue(self._mrd_queues.pop(evict_key, None)))
         return mrds_to_close
 
-    async def get(self, bucket_name, object_name, generation, pool_size):
+    async def get(self, bucket_name, object_name, generation, pool_size, file_obj=None):
```
Yonghui-Lee: Instead of passing file_obj, will it better to 
return details or set details as a field of MRD Pool?

Yonghui-Lee (approved):

PR #875, opened 2026-06-10
Fix MRDPool.close() blocking when pool is exhausted

zhixiangli:
Fix MRDPool.close() blocking when pool is exhausted

This PR fixes a bug where `MRDPool.close()` could block indefinitely if there are tasks waiting for a Multi-Request Device (MRD) to become available in an exhausted pool.

### Cause of the bug
Previously, `MRDPool.get_mrd()` would block on `await self._free_mrds.get()` while holding the pool's lock (`self._lock`). Because the lock was held during the await, any call to `MRDPool.close()` (which also requires the lock) would block waiting for the lock to be released. If no active MRDs were released to unblock the waiter, `close()` would hang.

### Solution
We replaced the blocking `asyncio.Queue.get()` call with a condition variable (`asyncio.Condition`).
1.  Waiters now wait on `self._mrd_available.wait()` when the pool is exhausted. This releases the lock while waiting, allowing other tasks (like `close()`) to acquire the lock.
2.  `MRDPool.close()` now acquires the lock, sets `self._closed = True`, clears `_all_mrds`, and notifies all waiters using `self._mrd_available.notify_all()`.
3.  Woken waiters check if the pool has been closed and raise a `RuntimeError("MRDPool is closed.")` instead of hanging.

#### Before
```mermaid
sequenceDiagram
    participant Waiter as Task 1 (Waiter)
    participant Lock as MRDPool._lock
    participant Queue as MRDPool._free_mrds
    participant Closer as Task 2 (Closer)

    Waiter->>Lock: Acquire Lock
    Note over Waiter, Lock: Lock Held
    Waiter->>Queue: get() (Blocks, Queue Empty)
    Note over Waiter: Blocked inside Lock
    Closer->>Lock: Acquire Lock (Blocks)
    Note over Closer: Blocked waiting for Lock
```

#### After
```mermaid
sequenceDiagram
    participant Waiter as Task 1 (Waiter)
    participant Lock as MRDPool._lock
    participant Cond as MRDPool._mrd_available
    participant Queue as MRDPool._free_mrds
    participant Closer as Task 2 (Closer)

    Waiter->>Lock: Acquire Lock
    Waiter->>Queue: get_nowait() (Empty)
    Waiter->>Cond: wait()
    Note over Waiter, Cond: Releases Lock & Blocks
    Closer->>Lock: Acquire Lock (Success)
    Closer->>Closer: Set _closed = True
    Closer->>Cond: notify_all()
    Closer->>Lock: Release Lock
    Note over Waiter: Wakes up, re-acquires Lock
    Waiter->>Waiter: Checks _closed (True) -> Raises RuntimeError
    Waiter->>Lock: Release Lock
```

### DIFF
--- a/gcsfs/tests/test_zb_hns_utils.py
+++ b/gcsfs/tests/test_zb_hns_utils.py
@@ -425,7 +425,7 @@ async def test_mrd_pool_close_with_exceptions(create_mrd_mock, mock_gcsfs):
 
 
 @pytest.mark.asyncio
-async def test_mrd_pool_queue_filled_during_lock_wait(mock_gcsfs):
+async def test_mrd_pool_get_mrd_reuses_free_mrd(mock_gcsfs):
     pool = MRDPool(mock_gcsfs, "bucket", "obj", "123", pool_size=1)
     mrd_mock = mock.AsyncMock()
 
@@ -436,10 +436,8 @@ async def test_mrd_pool_queue_filled_during_lock_wait(mock_gcsfs):
     with mock.patch.object(pool, "_create_mrd", side_effect=fake_create_mrd):
         await pool.initialize()
 
-        side_effects = [True] + [False] * 10
-        with mock.patch.object(pool._free_mrds, "empty", side_effect=side_effects):
-            async with pool.get_mrd() as mrd:
-                assert mrd == mrd_mock
+        async with pool.get_mrd() as mrd:
+            assert mrd == mrd_mock
 
         # We should not have spawned a new MRD
         assert pool._active_count == 1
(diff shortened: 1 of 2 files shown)


Yonghui-Lee:
We may not need to consider this case because `MRDPoolCache.close` is only called via `weakref.finalize` when ExtendedGcsFileSystem is being garbage collected. The close cannot run concurrently with other operations.

zhixiangli:
> We may not need to consider this case because `MRDPoolCache.close` is only called via `weakref.finalize` when ExtendedGcsFileSystem is being garbage collected. The close cannot run concurrently with other operations.

It is not MRDPoolCache; rather, it is MRDPool that is closed in several places.

It's in draft mode, but I don't plan to proceed with it as I've found a simpler/cleaner path.

PR #877, opened 2026-06-10
Introduce async API in Prefetcher and Integrate it with Disk Reads

googlyrahman:
## Description

**Introduce async API in Prefetcher and Integration with Disk Reads**

- This PR refactors the `BackgroundPrefetcher` to natively support asynchronous operations. 
- This PR integrates those asynchronous API of prefetcher to _get_file in Zonal and Standard Buckets.

## Zonal (Concurrency Enabled for > 5MB)

| Payload | Before Avg Throughput | After Avg Throughput | Before Memory Peak | After Memory Peak | Mem Delta |
|---------|-----------------------|----------------------|--------------------|-------------------|-----------|
| 64KB    | 0.69                  | 0.31                 | 103.7              | 103.9             | +0.2      |
| 1MB     | 11.53                 | 5.46                 | 108.5              | 109.1             | +0.6      |
| 10MB    | 39.94                 | 75.94                | 140.2              | 150.9             | +10.7     |
| 50MB    | 65.09                 | 334.74               | 189.2              | 216.6             | +27.4     |
| 100MB   | 73.12                 | 522.81               | 266.0              | 278.5             | +12.5     |
| 150MB   | 89.88                 | 613.78               | 329.3              | 346.4             | +17.0     |
| 1GB     | 67.89                 | 1346.88              | 419.5              | 533.1             | +113.5    |
| 5GB     | 62.88                 | 1885.74              | 564.1              | 635.0             | +70.9     |
| 10GB    | 75.92                 | 1884.96              | 584.3              | 644.9             | +60.6     |

## Regional (Concurrency Enabled for > 100MB)

| Payload | Before Avg Throughput | After Avg Throughput | Before Memory Peak | After Memory Peak | Mem Delta |
|---------|-----------------------|----------------------|--------------------|-------------------|-----------|
| 64KB    | 1.61                  | 1.10                 | 87.2               | 87.2              | +0.0      |
| 1MB     | 19.67                 | 15.26                | 87.6               | 87.6              | +0.1      |
| 10MB    | 142.04                | 102.54               | 87.7               | 87.7              | +0.0      |
| 50MB    | 225.54                | 200.85               | 88.4               | 88.6              | +0.2      |
| 100MB   | 235.33                | 268.08               | 89.3               | 89.3              | +0.0      |
| 150MB   | 256.19                | 254.82               | 164.7              | 209.0             | +44.3     |
| 1GB     | 295.30                | 527.34               | 212.1              | 355.6             | +143.5    |
| 5GB     | 219.51                | 702.02               | 250.3              | 367.9             | +117.6    |
| 10GB    | 252.99                | 784.28               | 262.2              | 374.2             | +112.0    |

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -305,11 +305,6 @@ class GCSFileSystem(asyn.AsyncFileSystem):
     async_impl = True
     MIN_CHUNK_SIZE_FOR_CONCURRENCY = 5 * 1024 * 1024
 
-    # This threshold applies to the standard bucket, whereas the zonal bucket
-    # uses a 5MB threshold. This difference exists because the standard bucket
-    # lacks the `DirectMemmoveBuffer` implementation used in the zonal bucket.
-    THRESOLD_FOR_DISK_READS = 100 * 1024 * 1024
-
     def __init__(
         self,
         project=DEFAULT_PROJECT,
@@ -382,6 +377,9 @@ class GCSFileSystem(asyn.AsyncFileSystem):
     def project(self):
         return self.credentials.project
 
+    # This threshold applies to the standard bucket, whereas the zonal bucket
+    # uses a 5MB threshold. This difference exists because the standard bucket
+    # lacks the `DirectMemmoveBuffer` implementation used in the zonal bucket.
     async def _get_threshold_for_disk_reads(self, bucket):
         return 100 * 1024 * 1024
 
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -10,6 +10,7 @@ import mimetypes
 import os
 import posixpath
 import re
+import sys
 import uuid
 import warnings
 import weakref
@@ -304,6 +305,11 @@ class GCSFileSystem(asyn.AsyncFileSystem):
     async_impl = True
     MIN_CHUNK_SIZE_FOR_CONCURRENCY = 5 * 1024 * 1024
 
+    # This threshold applies to the standard bucket, whereas the zonal bucket
+    # uses a 5MB threshold. This difference exists because the standard bucket
+    # lacks the `DirectMemmoveBuffer` implementation used in the zonal bucket.
+    THRESOLD_FOR_DISK_READS = 100 * 1024 * 1024
+
     def __init__(
         self,
         project=DEFAULT_PROJECT,
@@ -375,6 +381,9 @@ class GCSFileSystem(asyn.AsyncFileSystem):
     @property
     def project(self):
         return self.credentials.project
+
+    async def _get_threshold_for_disk_reads(self, bucket):
+        return 100 * 1024 * 1024
 
     # Clean up the aiohttp session
     #
@@ -1905,6 +1914,7 @@ class GCSFileSystem(asyn.AsyncFileSystem):
     async def _get_file_request(
         self, rpath, lpath, *args, headers=None, callback=None, **kwargs
     ):
+        rpath = self.url(rpath)
         consistency = kwargs.pop("consistency", self.consistency)
         await self._set_session()
         async with self.session.get(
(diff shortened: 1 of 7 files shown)


gcsfs/prefetcher.py
```diff
@@ -708,108 +720,111 @@ async def _restart_producer(self, new_offset: int):
...
+                requested_size = end - start
+                self.read_tracker.add(requested_size)
+
+                chunk = await self.consumer.consume(requested_size)
+                self.user_offset += len(chunk)
+
+                logger.debug("Completed _async_fetch. Returned %d bytes.", len(chunk))
+                return chunk
+            except asyncio.CancelledError as e:
+                self._error = e
```
zhixiangli, line 794: Should a cancelled read auto-restart the producer on the following read?

gcsfs/prefetcher.py
```diff
@@ -708,108 +720,111 @@ async def _restart_producer(self, new_offset: int):
...
+        async with self._async_lock:
+            if self.is_stopped:
+                return
+
+            self.is_stopped = True
+            logger.debug("Acquired async lock. Tearing down producer and buffers.")
 
-        chunk = await self.consumer.consume(requested_size)
-        self.user_offset += len(chunk)
+            if self.producer:
```
zhixiangli, line 814: Is there any reason the producer might be None when closed?

gcsfs/prefetcher.py
```diff
@@ -263,6 +262,8 @@ async def stop(self):
             )
             await asyncio.gather(*tasks_to_wait, return_exceptions=True)
 
+        self.wakeup_event.clear()
```
zhixiangli, line 265: It's unnecessary, as the restart func start() already clears the event.

googlyrahman: I agree that it is unnecessary because
- If someone calls `restart` -> it gets cleared
- if someone calls `stop` and then `start` -> it gets cleared because the clear happens in start as pro-active measure.

But if someone just calls `stop` and never starts the loop again, the wakeup event might be left in this scenario. While this does not harm us today because the state of wakeup event does not matter if the loop is not gonna start again in future, but it just isn't right or logical to leave the wakeup event set after the loop is stopped. That is why I added it.  Additionally, I think we should continue keeping the clear in start as well for defensive programming.

gcsfs/prefetcher.py
googlyrahman: Ideally yes! If a read gets cancelled mid-flight it can change some of the internal states before cancellation actually takes effect, which may lead the product to behave in unpredictable way for subsequent reads.

gcsfs/prefetcher.py
googlyrahman: No, i don't think the producer would ever be `None` at this line of code.

I think the reason for keeping the check of `if self.producer` is to satisfy the  `lint` complaints (Those red wavy underlines 😁 ) in my IDE where it says that self.producer can be uninitialised.

gcsfs/prefetcher.py
```diff
@@ -649,37 +650,62 @@ def __init__(self, fetcher, size: int, concurrency: int, max_prefetch_size=None)
...
-        fsspec.asyn.sync(self.loop, _start)
+        try:
+            current_loop = asyncio.get_running_loop()
+        except RuntimeError:
+            current_loop = None
+
+        if current_loop is self.loop and current_loop.is_running():
+            _start()
+        elif current_loop is not None and current_loop != self.loop:
+            raise RuntimeError(
```
zhixiangli: Why does this happen, and can we avoid it in the code?

gcsfs/core.py
```diff
@@ -1936,12 +1944,208 @@ async def _get_file_request(
...
+                    f.write(chunk)
+
+        pending_writes = set()
+        try:
+            if hasattr(os, "pwrite"):
+                fd = os.open(lpath, os.O_WRONLY | getattr(os, "O_BINARY", 0))
+
+            async with prefetcher:
+                offset = 0
+                while offset < total_size:
```
zhixiangli, line 2060: Should we verify if offset == total_size after the loop?

googlyrahman: Added this logic.

gcsfs/prefetcher.py
googlyrahman: Updated the logic, and added comments.

zhixiangli (approved):

googlyrahman:
@zhixiangli@google.com

I was investigating why zonal disk download throughput does not match zonal RAM throughput and found that the issue was related to `MRDPoolCache`. The previous version assumed that repetitive calls to `_cat_file` without passing an explicit `MRDPool` would be efficiently handled by `MRDPoolCache`. However, that does not appear to be the case.

It turns out that explicitly passing the `MRDPool` is the best approach. Doing so improved the disk download throughput to 1.9 GB/s, which is close to the RAM download throughput of 2.2 GB/s.

The changes have been added as a separate commit, so you don’t need to re-review all of PR again. I’ve also created a separate issue for `MRDPoolCache` and will sync with Yonghui further on that. In the meantime, I’ve removed the dependency on `MRDPoolCache` to unblock this PR.

The issue = https://github.com/fsspec/gcsfs/issues/895

googlyrahman:
@martindurant, can you also take a look if you're interested, The before vs after comparision is listed in PR description for both standard, and zonal buckets.

martindurant:
Thanks for putting the (final) numbers up top, that's a good pattern.

PR #878, opened 2026-06-11
Fix multiple `_info()` calls in `_process_limits_to_offset_and_length`

zhixiangli:
Resolves an issue where `ExtendedGcsFileSystem._process_limits_to_offset_and_length` could issue up to three separate `_info()` requests to GCS when calculating limits for range reads with negative or `None` offsets. It does this by fetching the size lazily via an internal async function and memoizing it.

### DIFF
--- a/gcsfs/tests/test_extended_gcsfs_unit.py
+++ b/gcsfs/tests/test_extended_gcsfs_unit.py
@@ -696,17 +696,23 @@ async def test_get_all_folders_start_dir_calculation(
 async def test_process_limits_to_offset_and_length_info_calls():
     extended_gcsfs = ExtendedGcsFileSystem(token="anon", skip_instance_cache=True)
     path = "bucket/file"
-    with mock.patch.object(extended_gcsfs, "_info", new_callable=mock.AsyncMock) as mock_info:
+    with mock.patch.object(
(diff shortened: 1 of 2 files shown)


gcsfs/tests/test_extended_gcsfs_unit.py
```diff
@@ -692,6 +692,46 @@ async def test_get_all_folders_start_dir_calculation(
...
+        # Test case where size is fetched (None, None)
+        await extended_gcsfs._process_limits_to_offset_and_length(
+            path, start=None, end=None
+        )
+        mock_info.assert_awaited_once_with(path)
+
+        mock_info.reset_mock()
+
+        # Test case where size is not fetched because start and end are positive and end > size
+        # wait, if end > size, it actually WILL fetch size to clamp length.
```
yuxin00j: nit: clean up comments

yuxin00j (changes requested):

gcsfs/tests/test_extended_gcsfs_unit.py
zhixiangli: @jules

martindurant (approved):

PR #879, opened 2026-06-11
Fix: Route GCSMap through dynamically resolved GCSFileSystem

zhixiangli:
Fixes `GCSMap` in `gcsfs/mapping.py` to route through the environment-selected `GCSFileSystem` class (e.g. `ExtendedGcsFileSystem` for Zonal/HNS) instead of hardcoding `.core.GCSFileSystem`. This ensures mapping users (such as Dask/Zarr) correctly pick up the experimental features when they are enabled via the environment variable. Also updated mapping tests to verify this behavior dynamically.

### DIFF
--- a/gcsfs/__init__.py
+++ b/gcsfs/__init__.py
@@ -28,6 +28,4 @@ if os.getenv("GCSFS_EXPERIMENTAL_ZB_HNS_SUPPORT", "true").lower() in ("true", "1
         )
         # Fallback to core GCSFileSystem, do not register here
 
-# TODO: GCSMap still refers to the original GCSFileSystem. This will be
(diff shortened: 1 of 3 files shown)


Yonghui-Lee:
We can remove the comment in `__init__.py`.
```
# TODO: GCSMap still refers to the original GCSFileSystem. This will be
# addressed in a future update.
```

zhixiangli:
> We can remove the comment in `__init__.py`.
> 
> ```
> # TODO: GCSMap still refers to the original GCSFileSystem. This will be
> # addressed in a future update.
> ```

@jules

Yonghui-Lee (approved):

PR #883, opened 2026-06-11
Fix multi-process benchmark child process crash handling

zhixiangli:
Fixes a critical issue where multi-process benchmarks could silently pass and publish stale or partial results if worker processes crashed.

- Resets the shared data array before each round.
- Inspects process exit codes after joining; terminates remaining processes and raises an error on failure.
- Includes tests in `test_runner.py` for both the failure handling and array reset logic.

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/runner.py
+++ b/gcsfs/tests/perf/microbenchmarks/runner.py
@@ -167,6 +167,9 @@ def run_multi_process(
             logging.info(
                 f"Multi-process {benchmark_group} benchmark: Starting round {round_num + 1}/{params.rounds}."
             )
+            for i in range(params.processes):
+                process_data_shared[i] = 0.0
+
             processes = []
 
             affinity_set = False
@@ -195,8 +198,14 @@ def run_multi_process(
                 if affinity_set:
                     os.sched_setaffinity(0, old_affinity)
 
-            for p in processes:
+            for i, p in enumerate(processes):
                 p.join()
+                if p.exitcode != 0:
+                    for remaining_p in processes:
+                        if remaining_p.is_alive():
+                            remaining_p.terminate()
+                            remaining_p.join()
+                    raise RuntimeError(f"Worker process {i} exited with code {p.exitcode}")
 
             if getattr(params, "runtime", None):
                 results.append(int(sum(process_data_shared[:])))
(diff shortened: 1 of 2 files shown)


yuxin00j (commented):
1. **Conceptual mismatch in test** (`test_runner.py`)

Conceptually, a process that just crashed and was `join()`ed would not still be alive. While this setup in `test_run_multi_process_child_fails` successfully triggers the `terminate()` block for testing purposes, it might be more realistic to simulate `mock_process1` as the one that crashed and `mock_process2` as the sibling process that is still running.

For example, something like this tests the same termination logic but reflects the real-world scenario better:
```python
    mock_process1 = mock.Mock()
    mock_process1.exitcode = 1
    mock_process1.is_alive.return_value = False

    mock_process2 = mock.Mock()
    mock_process2.exitcode = None
    mock_process2.is_alive.return_value = True

    mock_ctx.Process.side_effect = [mock_process1, mock_process2]
```

2. **Duplicate assignment** (`test_runner.py`)

In `test_run_multi_process_resets_shared_data`, there is a redundant assignment you can remove:
```python
    mock_process.exitcode = 0
    mock_process.exitcode = 0
```

zhixiangli:
> 1. **Conceptual mismatch in test** (`test_runner.py`)
> 
> Conceptually, a process that just crashed and was `join()`ed would not still be alive. While this setup in `test_run_multi_process_child_fails` successfully triggers the `terminate()` block for testing purposes, it might be more realistic to simulate `mock_process1` as the one that crashed and `mock_process2` as the sibling process that is still running.
> 
> For example, something like this tests the same termination logic but reflects the real-world scenario better:
> 
> ```python
>     mock_process1 = mock.Mock()
>     mock_process1.exitcode = 1
>     mock_process1.is_alive.return_value = False
> 
>     mock_process2 = mock.Mock()
>     mock_process2.exitcode = None
>     mock_process2.is_alive.return_value = True
> 
>     mock_ctx.Process.side_effect = [mock_process1, mock_process2]
> ```
> 
> 2. **Duplicate assignment** (`test_runner.py`)
> 
> In `test_run_multi_process_resets_shared_data`, there is a redundant assignment you can remove:
> 
> ```python
>     mock_process.exitcode = 0
>     mock_process.exitcode = 0
> ```

@jules

yuxin00j (approved):

PR #890, opened 2026-06-12
Update Codecov config

zhixiangli:
This commit updates `.github/.codecov.yml` to:
1. Add `require_ci_to_pass: true` so Codecov checks wait for CI to finish and pass.
2. Allow up to a `0.5%` code coverage drop by updating the threshold to `0.5%` instead of `0%` to avoid false positve.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,12 +1,15 @@
 ignore:
   - "gcsfs/tests/**/*"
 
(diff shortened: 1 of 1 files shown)


yuxin00j:
Can I check why are we making this change to codecov?

yuxin00j (approved):

PR #891, opened 2026-06-12
Perf: Fix CPU and Memory aggregation logic in ResourceMonitor

yuxin00j:
### Description
This PR fixes two major bugs in `ResourceMonitor` that were causing the performance metrics for the microbenchmark's process tree to report incorrectly.

#### 1. Fixed CPU Usage Drop for Child Processes
`psutil.cpu_percent()` requires two calls to establish an interval. The previous logic recreated the `psutil.Process()` objects on every interval iteration, causing it to constantly evaluate as the "first call" and return `0.0` for all child processes. This is fixed by caching the `Process` objects in a dictionary indexed by PID.

#### 2. Fixed Memory Usage Inflation (Shared Memory Double-Counting)
The previous logic summed `rss` (Resident Set Size) across the parent and all child processes. Because `multiprocessing` uses Copy-on-Write (`fork()`), child processes share a significant portion of physical memory pages with the parent. Summing `rss` double/triple counts these shared pages. This is fixed by using `pss` (Proportional Set Size) from `memory_full_info()`, which correctly apportions shared pages across the processes sharing them, providing an accurate representation of the benchmark's true, unique memory footprint.

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/resource_monitor.py
+++ b/gcsfs/tests/perf/microbenchmarks/resource_monitor.py
@@ -37,29 +37,54 @@ class ResourceMonitor:
         self.stop()
 
     def _monitor(self):
-        psutil.cpu_percent(interval=None)
         current_process = psutil.Process()
+        known_procs = {
+            (current_process.pid, current_process.create_time()): current_process
+        }
+        current_process.cpu_percent(interval=None)
+
         while not self._stop_event.is_set():
             try:
                 # CPU and Memory tracking for current process tree
-                total_cpu = current_process.cpu_percent(interval=None)
-                current_mem = current_process.memory_info().rss
-                for child in current_process.children(recursive=True):
+                all_procs = [current_process] + current_process.children(recursive=True)
+
+                total_cpu = 0.0
+                current_mem = 0.0
+                alive_keys = set()
+
+                for p in all_procs:
                     try:
-                        total_cpu += child.cpu_percent(interval=None)
-                        current_mem += child.memory_info().rss
+                        current_mem += p.memory_info().rss
+
+                        # CPU: Track processes across loops to properly calculate cpu_percent
+                        proc_key = (p.pid, p.create_time())
+                        alive_keys.add(proc_key)
+
+                        if proc_key not in known_procs:
+                            known_procs[proc_key] = p
+                            # Prime newly discovered process. As per psutil docs:
+                            # "the first time this is called it will return a meaningless
+                            # 0.0 value which you are supposed to ignore."
+                            p.cpu_percent(interval=None)
+                        else:
+                            total_cpu += known_procs[proc_key].cpu_percent(
+                                interval=None
+                            )
                     except (psutil.NoSuchProcess, psutil.AccessDenied):
                         continue
+
+                # Remove dead processes from tracking dictionary
+                known_procs = {
+                    k: proc for k, proc in known_procs.items() if k in alive_keys
+                }
 
                 # Normalize CPU by number of vcpus
                 global_cpu = total_cpu / self.vcpus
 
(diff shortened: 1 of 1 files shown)


gcsfs/tests/perf/microbenchmarks/resource_monitor.py
```diff
@@ -37,29 +37,52 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.stop()
 
     def _monitor(self):
-        psutil.cpu_percent(interval=None)
         current_process = psutil.Process()
+        known_procs = {current_process.pid: current_process}
```
zhixiangli: The key should be pid and create_time as the same pid may be reused by OS.

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
```diff
@@ -37,29 +37,52 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                current_mem = 0.0
+
+                for p in all_procs:
                     try:
-                        total_cpu += child.cpu_percent(interval=None)
-                        current_mem += child.memory_info().rss
+                        # Memory: Prefer PSS (Proportional Set Size) to avoid double-counting shared
+                        # memory between parent and children. Fallback to RSS if PSS isn't available.
+                        try:
+                            mem = p.memory_full_info().pss
```
zhixiangli: Is memory_full_info slow?

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
```diff
@@ -37,29 +37,52 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                        except (AttributeError, psutil.AccessDenied):
+                            mem = p.memory_info().rss
```
zhixiangli: Will switching from RSS to PSS break past metrics if we compare them?

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
```diff
@@ -37,29 +37,52 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                        current_mem += mem
+
+                        # CPU: Track processes across loops to properly calculate cpu_percent
+                        if p.pid not in known_procs:
+                            known_procs[p.pid] = p
+                            p.cpu_percent(
+                                interval=None
+                            )  # prime newly discovered process
+                        else:
+                            total_cpu += known_procs[p.pid].cpu_percent(interval=None)
```
zhixiangli: > That means the first time this is called it will return a meaningless 0.0 value which you are supposed to ignore.

https://psutil.readthedocs.io/stable/#psutil.cpu_percent

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
yuxin00j: Updated.

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
yuxin00j: Reverted this change.

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
yuxin00j: Reverted this change.

gcsfs/tests/perf/microbenchmarks/resource_monitor.py
yuxin00j: Yes, we ignore the first call. In subsequent loops, we added it to total_cpu.

zhixiangli (approved):

PR #892, opened 2026-06-12
Optimize E2E tests: parallelize resource creation and enable parallel test execution

zhixiangli:
This PR optimizes the E2E test workflow by parallelizing resource creation and enabling parallel test execution using `pytest-xdist`, significantly reducing the overall build time.

### Changes:

*   **Resource Provisioning**:
    *   Split the former `create_resources.sh` into `create_vm.sh` and `create_buckets.sh` to allow VM provisioning + setup to run in parallel with bucket creation.
    *   Updated `create_buckets.sh` to create base buckets and worker-specific bucket variants (e.g., `bucket-gw0`, `bucket-gw1`) in parallel.
    *   Updated `cleanup.sh` to clean up all bucket variants.
    *   Parameterized VM machine type in resource creation.

*   **Test Execution**:
    *   Installed `pytest-xdist` on the test VM.
    *   Configured `pytest` to run with multiple workers if `PYTEST_XDIST_WORKERS` is set.
    *   Updated test settings to append worker suffixes to bucket names, allowing parallel workers to use isolated buckets.
    *   Added support for a custom worker separator (`GCSFS_TEST_BUCKET_WORKER_SEPARATOR`), set to `-` for Cloud Build.

*   **CI Configuration**:
    *   Updated `e2e-tests-cloudbuild.yaml` to run VM creation/setup and bucket creation steps in parallel.
    *   Configured different worker counts for different test suites (4 for standard, 8 for HNS, 2 for zonal/zonal-core) to optimize resource usage and speed.

**The time for e2e tests has dropped from 20 minutes to 6 minutes.**


cloudbuild/run_tests.sh
```diff
@@ -13,6 +13,7 @@ export STORAGE_EMULATOR_HOST=https://storage.googleapis.com
 export GCSFS_TEST_PROJECT=${PROJECT_ID}
 export GCSFS_TEST_KMS_KEY=projects/${PROJECT_ID}/locations/${REGION}/keyRings/${KEY_RING}/cryptoKeys/${KEY_NAME}
 export GOOGLE_CLOUD_PROJECT=${PROJECT_ID}
+export GCSFS_TEST_BUCKET_WORKER_SEPARATOR="${GCSFS_TEST_BUCKET_WORKER_SEPARATOR:--}"
```
Yonghui-Lee: There is inconsistency in the default value for  GCSFS_TEST_BUCKET_WORKER_SEPARATOR .
- Bash scripts ( create_buckets.sh ,  cleanup.sh ,  run_tests.sh ): Default to  `-`
- Python config ( gcsfs/tests/settings.py ): Defaults to  `_`

cloudbuild/create_buckets.sh
```diff
@@ -0,0 +1,114 @@
...
+# Use a separate bucket for running core tests to avoid exceeding object rate limit
+echo "--- Creating Zonal bucket for running core tests ---"
+create_bucket_variants "${WORKERS_ZONAL_CORE}" "gcsfs-test-zonal-core-${SHORT_BUILD_ID}" \
+    --placement="${ZONE}" \
+    --default-storage-class=RAPID \
+    --enable-hierarchical-namespace \
+    --uniform-bucket-level-access
+
+# Wait for all background bucket creation jobs to finish
+wait
```
Yonghui-Lee: In bash,  wait without arguments returns  0  even if some of the background jobs failed. Because  `set -e`  is enabled, it only triggers on non-zero exit codes of commands in the main thread. Since  wait  returns  0 , the script will exit with  0  (success) even if some jobs failed.

Proposed Fix: 
Collect the PIDs of the background jobs and wait for them individually. This ensures that if any job fails, the  wait  command for that job will return a non-zero exit code, triggering  `set -e`  and failing the script immediately.

Yonghui-Lee (changes requested):

cloudbuild/create_buckets.sh
```diff
@@ -0,0 +1,132 @@
...
+        done
+    fi
+}
+
+create_bucket_variants() {
+    local workers="$1"
+    local bucket_base="$2"
+    shift 2
+
+    gcloud storage buckets create "gs://${bucket_base}" \
```
Yonghui-Lee: nit: creating both base and worker-specific variant buckets might be redundant.

Yonghui-Lee (approved):

PR #893, opened 2026-06-15
Optimize _rm batchsize and speed up tests

zhixiangli:
- gcsfs/core.py: Increase batchsize in _rm to 100 (GCS limit) to minimize round-trips when deleting many objects. Speed up tests clean up.
- gcsfs/tests/test_credentials.py: Speed up test_credentials_from_raw_token by patching asyncio.sleep to avoid backoff delays on invalid tokens.
- gcsfs/tests/test_fuse.py: Speed up test_fuse by reducing wait times, as it is xfail.

TAG=agy
CONV=b864cf2a-71de-497b-afb3-0ec348546d1e

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1591,7 +1591,10 @@ class GCSFileSystem(asyn.AsyncFileSystem):
                 [self._rm_file(f) for f in files], return_exceptions=True, batch_size=5
             )
 
-    async def _rm(self, path, recursive=False, maxdepth=None, batchsize=20):
+    async def _rm(self, path, recursive=False, maxdepth=None, batchsize=100):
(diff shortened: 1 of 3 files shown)


martindurant:
I am interested if you know how much of a performance difference this makes.

zhixiangli:
> I am interested if you know how much of a performance difference this makes.

It is 3 minutes faster, decreasing from 23 minutes to 20 minutes.

yuxin00j (approved):

Yonghui-Lee:
Do we want to also increase for zonal file _rm method?

zhixiangli:
> Do we want to also increase for zonal file _rm method?

No, zonal test sets are already fast enough within the scope of this PR.

PR #896, opened 2026-06-17
Check finalized state to optimize MRD pool initialization

### LINKED ISSUE #895: Unexpected Creating/Closing with MRDPoolCache lowers the throughput.
I noticed a significant performance drop (a difference of around 800 MB/s) when making concurrent `_cat_file` calls without passing an explicit `MRDPool`. I originally expected `MRDPoolCache` to handle the pooling and caching automatically, but it is actually much faster to provide the pool explicitly.

#### Initial Analysis

After looking into it, I found that new MRDs are being created and closed repeatedly. This seems to stem from `MRDPoolCache` calling `MRDPool.initialize()` [on this line](https://github.com/fsspec/gcsfs/blob/main/gcsfs/zb_hns_utils.py#L750). Currently, `initialize()` does not check the cache first. It just creates a new MRD every time.

Because of this, every get request from `MRDPoolCache` spins up a new MRD. Once the total exceeds the global limit and tasks finish their work, the system sees the limit is exceeded and immediately starts closing the older MRDs due to the condition [listed here](https://github.com/fsspec/gcsfs/blob/main/gcsfs/zb_hns_utils.py#L780).

#### Proposed Fix

Do not cache MRD for unfinalized objects, and update `MRDPool.initialize()` to check the cache for an idle MRD before requesting a new one should resolve this constant creation and teardown cycle.

Yonghui-Lee:
Optimize MRD pool initialization and cache reuse based on the object's finalization state.

Close #895

## Key Changes
*   **Finalization Detection**: `MRDPoolCache` now detects if an object is finalized by checking for `timeFinalized` in the metadata.
*   **Initialization**:
    *   **Finalized**: `MRDPool.initialize()` reuses an idle MRD from the cache.
    *   **Unfinalized**: `MRDPool.initialize()` always creates a new MRD to fetch the up-to-date `persisted_size`.

### DIFF
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -537,10 +537,12 @@ class MRDPool:
                 raise RuntimeError("Cannot initialize a closed MRDPool.")
 
             if not self._initialized and self._active_count == 0:
-                if not self.finalized:
-                    mrd = await self._create_mrd()
-                else:
+                if self.finalized:
                     mrd = await self._get_or_create_mrd()
+                else:
+                    # Always create a new MRD for unfinalized objects to get the up-to-date persisted_size
+                    mrd = await self._create_mrd()
+                    self._all_mrds.append(mrd)
                 self.persisted_size = mrd.persisted_size
                 self._free_mrds.put_nowait(mrd)
                 self._active_count += 1
--- a/gcsfs/zb_hns_utils.py
+++ b/gcsfs/zb_hns_utils.py
@@ -465,6 +465,7 @@ class MRDPool:
         bucket_name,
         object_name,
         generation,
+        finalized,
         pool_size,
         cache=None,
     ):
@@ -480,6 +481,7 @@ class MRDPool:
         self._lock = asyncio.Lock()
         self.details = None
         self.persisted_size = None
+        self.finalized = finalized
         self._initialized = False
         self._closed = False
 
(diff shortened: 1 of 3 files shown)


gcsfs/zb_hns_utils.py
```diff
@@ -535,9 +537,10 @@ async def initialize(self):
                 raise RuntimeError("Cannot initialize a closed MRDPool.")
 
             if not self._initialized and self._active_count == 0:
-                # Always create a new MRD on initialization to get the up-to-date persisted_size
```
zhixiangli, line 538: Let's keep this comment as this information is important.

gcsfs/zb_hns_utils.py
```diff
@@ -535,9 +537,10 @@ async def initialize(self):
...
-                mrd = await self._create_mrd()
-                self._all_mrds.append(mrd)
```
zhixiangli, line 540: Why no longer append to all_mrds?

gcsfs/zb_hns_utils.py
```diff
@@ -465,6 +465,7 @@ def __init__(
         bucket_name,
         object_name,
         generation,
+        finalized,
```
zhixiangli, line 468: Will it break other constructors? Ideally it needs to be in the bottom.

zhixiangli (changes requested):

gcsfs/zb_hns_utils.py
Yonghui-Lee: It won't break because Iit's only used internally. In logic, it's better in this position.

gcsfs/zb_hns_utils.py
Yonghui-Lee: Done

gcsfs/zb_hns_utils.py
Yonghui-Lee: Done

gcsfs/zb_hns_utils.py
zhixiangli: A PR was recently merged that affects the MRD; could you please confirm that nothing is broken?

Yonghui-Lee: I rebase and check that it works fine.

zhixiangli (approved):

PR #900, opened 2026-06-18
Run e2e-tests in parallel with ci block in release.yml

zhixiangli:
Removes `needs: ci` from the `e2e-tests` block so it can run in parallel with the `ci` block in `.github/workflows/release.yml`. This decreases CI time for release builds as the build block already waits for both CI and e2e-tests to finish.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,6 @@ jobs:
 
   e2e-tests:
     name: Wait for E2E Integration Tests (Cloud Build)
-    needs: ci
     runs-on: ubuntu-latest
     permissions:
       contents: read


Mahalaxmibejugam:
It would actually not improve the CI execution time. The flow for release pipeline is as follows:

Whenever a tag is created, it would start the cloudBuild pipeline in the background as we have setup a trigger to listen to tag pushes. In parallel release pipeline would trigger the emulator based CI pipeline which usually gets finished in under 3 mins. So the cloudBuild based tests and emulator tests are already running in parallel. Second step in the release pipeline is just polling for the cloudBuild pipeline to finish.

We would just be polling for more number of times by parallelizing the steps in release pipeline, if the unit tests CI fail then we don't have to poll for the cloudbuild status.

zhixiangli:
> It would actually not improve the CI execution time. The flow for release pipeline is as follows:
> 
> Whenever a tag is created, it would start the cloudBuild pipeline in the background as we have setup a trigger to listen to tag pushes. In parallel release pipeline would trigger the emulator based CI pipeline which usually gets finished in under 3 mins. So the cloudBuild based tests and emulator tests are already running in parallel. Second step in the release pipeline is just polling for the cloudBuild pipeline to finish.
> 
> We would just be polling for more number of times by parallelizing the steps in release pipeline, if the unit tests CI fail then we don't have to poll for the cloudbuild status.

Oh, yes, thanks for reminding me of this; I remember the logic now.

PR #906, opened 2026-06-18
Implement makedirs for HNS buckets

Mahalaxmibejugam:
This PR adds support for `makedirs` recursively creating paths and directories in HNS bucket. It also adds unit and integration tests.

- makedirs is restricted to creating folders within an existing HNS bucket and is a no-op for standard buckets.
- makedirs does not allow the creation of a folder that shares a name with an existing file at the same level to maintain filesystem compatibility, even though GCS supports the creation of files and folders with the same name at the same level.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -855,6 +855,18 @@ class ExtendedGcsFileSystem(HnsDirCacheUpdater, GCSFileSystem):
         self, path, bucket, key, create_parents, exist_ok=True
     ):
         logger.debug(f"Using HNS-aware mkdir for '{path}'.")
+
+        # Preemptively check if the path already exists to ensure fsspec compatibility
+        try:
+            info = await self._info(path)
+            if info["type"] != "directory":
+                raise FileExistsError(f"A file already exists at the path: {path}")
+            if not exist_ok:
+                raise FileExistsError(f"Directory already exists: {path}")
+            return
+        except FileNotFoundError:
+            pass
+
         parent = f"projects/_/buckets/{bucket}"
         folder_id = key.rstrip("/") + "/"
         request = storage_control_v2.CreateFolderRequest(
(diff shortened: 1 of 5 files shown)


yuxin00j (approved):

gcsfs/extended_gcsfs.py
```diff
@@ -885,6 +908,58 @@ async def _create_hns_folder(self, path, bucket, key, create_parents):
...
+            will still error if it exists as a file.
+
+        Raises
+        ------
+        FileNotFoundError
+            If the target bucket does not exist.
+        FileExistsError
+            If `exist_ok` is False and the directory already exists, or if a
+            file already exists at the target path regardless of `exist_ok`.
+        """
```
zhixiangli, line 933: The protocol needs to be stripped from the path to avoid polluting the cache.

gcsfs/tests/integration/test_extended_hns.py
```diff
@@ -449,6 +449,90 @@ def test_mkdir_enable_hierarchical_namespace(self, gcs_hns, buckets_to_delete):
...
 
+    def test_hns_mkdir_existing_file_error(self, gcs_hns):
+        """Test mkdir raises FileExistsError if a file exists at the path."""
+        gcsfs = gcs_hns
+        file_path = f"{TEST_HNS_BUCKET}/existing_file_mkdir_{uuid.uuid4().hex}.txt"
+        gcsfs.touch(file_path)
+        assert gcsfs.isfile(file_path)
+
+        with pytest.raises(FileExistsError, match="A file already exists"):
+            gcsfs.mkdir(file_path)
```
zhixiangli, line 460: nit: the folder isn't cleaned up.

Mahalaxmibejugam: cleanup would be taken care as part of the gcs_hns fixture

zhixiangli (approved):

PR #907, opened 2026-06-19
Implement zero-copy in write path of standard bucket

Yonghui-Lee:
During chunked uploads of standard buckets, GCSFS previously performed multiple memory copies of the data:
1. `self.buffer.getvalue()` created a complete copy of the internal buffer's bytes in memory.
2. `data[:chunk_length]` created another copy (slice) of the chunk to be uploaded.
3. `UnclosableBytesIO(data)` in the module-level `upload_chunk` and `simple_upload` wrapped the chunk in a stream, creating yet another copy of the chunk data.

For large write blocks, these copies caused substantial CPU overhead and transient RAM spikes.

### The Solution
1. **Zero-Copy Buffer Access**: Replaced `self.buffer.getvalue()` with `self.buffer.getbuffer()` to retrieve a zero-copy `memoryview` of the write buffer.
2. **Zero-Copy Slicing**: Slices are now created directly on `memoryview` objects (in `_upload_chunk` and `_pipe_file`), which is a zero-copy operation in Python.
3. **Bypassing Stream Wrappers**: Since `aiohttp` natively supports `bytes` and `memoryview` payloads without requiring stream-seeking or closing, we bypass `UnclosableBytesIO` wrapping entirely for these types, allowing the network socket to stream directly from the original buffer.
4. **Robust Testing**: Updated `test_upload_chunk_shortfall` to correctly assert on raw `bytes` payloads in addition to wrapped streams.

### Performance Impact

Chunk Size | Baseline (Before) | Optimized (After) | Change |
| :--- | :---: | :---: | :---: |
| **16 MiB** | 83.56(MiB/s) | 84.09(MiB/s) | **+0.6%** |
| **32 MiB** | 86.40(MiB/s) | 92.80(MiB/s) | **+7.4%** |

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1679,8 +1679,9 @@ class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
                 mode=mode,
             )
             try:
-                for offset in range(0, len(data), chunksize):
-                    bit = data[offset : offset + chunksize]
+                data_view = memoryview(data)
+                for offset in range(0, size, chunksize):
+                    bit = data_view[offset : offset + chunksize]
                     out = await upload_chunk(
                         self, location, bit, offset, size, content_type
                     )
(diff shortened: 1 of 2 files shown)


zhixiangli (approved):

ankitaluthra1:
Is performance impact consistent ? +0.6% is could very likely be variance in test results, is this mean or median over multiple run ?

Yonghui-Lee:
> Is performance impact consistent ? +0.6% is could very likely be variance in test results, is this mean or median over multiple run ?

It's mean of multiple runs. I run again with 10 rounds, the mean and median of optimized approach are slightly higher.

martindurant:
I would have expected a bigger performance impact - the bottleneck is actually in the encode/send? It's worth it nevertheless for the smaller memory use. This is another improvement that could be made across all HTTP-based backends.

PR #908, opened 2026-06-19
Add cache_type in user-agent fo http client 

ankitaluthra1:
- This PR updates the user-agent header to contain cache_type for Standard buckets using http client.
- This PR does not include same propagation for grpc client which would be done separately.


gcsfs/core.py
zhixiangli, line 1942: Need to pass cache type

gcsfs/core.py
```diff
@@ -469,12 +469,15 @@ def _get_params(self, kwargs):
...
 
-    def _get_headers(self, headers):
+    def _get_headers(self, headers, cache_type=None):
         out = {}
         if headers is not None:
             out.update(headers)
         if "User-Agent" not in out:
-            out["User-Agent"] = "python-gcsfs/" + version
+            ua = "python-gcsfs/" + version
+            if cache_type:
```
martindurant, line 478: Why not just `self.cache_type` rather than having to pass this at every call site?

ankitaluthra1: Are you saying to keep a global variable cache_type at GCSFileSystem level? 
cache_type is at GCSFile level and can be different for different objects, so it needs to be passed in every call from GCSFile and cant be FileSystem level variable.

gcsfs/core.py
ankitaluthra1: IIUC _get_file_request  is an internal method used to download a file from GCS to local disk sequentially. Its a straight stream-to-disk copy operation. It does not use any  fsspec  memory caching mechanism. Since no cache is involved, hence default  python-gcsfs/{version}  is the correct user agent here.

zhixiangli: Are we targeting the pass cache type for relevant usage or all usage?

ankitaluthra1: only for relevant paths where caching is actually used

gcsfs/core.py
martindurant: Got you.

Did we not have a way to pass general header information from a file instance on every upload already?

PR #918, opened 2026-06-22
Fix HNS _list_objects error swallowing

zhixiangli:
HNS _list_objects swallows all folder-metadata errors. This commit fixes it by catching only FileNotFoundError, logging unexpected exceptions, and re-raising them.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -747,9 +747,6 @@ class ExtendedGcsFileSystem(HnsDirCacheUpdater, GCSFileSystem):
                     return []
                 except FileNotFoundError:
                     pass
-                except Exception as e:
-                    logger.warning(f"Error fetching directory info for {path}: {e}")
-                    raise
             raise
 
     async def _mkdir(
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -747,6 +747,9 @@ class ExtendedGcsFileSystem(HnsDirCacheUpdater, GCSFileSystem):
                     return []
                 except FileNotFoundError:
                     pass
+                except Exception as e:
+                    logger.warning(f"Error fetching directory info for {path}: {e}")
+                    raise
             raise
 
     async def _mkdir(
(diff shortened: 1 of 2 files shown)


gcsfs/tests/test_extended_hns_gcsfs.py
```diff
@@ -2985,3 +2985,52 @@ async def test_get_bucket_type_general_exception(self, extended_gcsfs, caplog):
...
+
+
+@pytest.mark.asyncio
+async def test_list_objects_file_not_found_hns():
+    fs = ExtendedGcsFileSystem(token="anon")
+    with (
+        mock.patch(
+            "gcsfs.core.GCSFileSystem._list_objects", new_callable=mock.AsyncMock
+        ) as mock_super_list,
+        mock.patch.object(fs, "split_path", return_value=("bucket", "key", "gen")),
```
Mahalaxmibejugam: Can we remove the mock on split path as there is no API call involved in it?

gcsfs/extended_gcsfs.py
```diff
@@ -745,7 +745,7 @@ async def _list_objects(self, path, prefix="", versions=False, **kwargs):
                 try:
                     await self._get_directory_info(path, bucket, key, None)
                     return []
-                except (FileNotFoundError, Exception):
+                except FileNotFoundError:
```
Mahalaxmibejugam, line 748: Catching only FileNotFoundError in _list_objects means any permission or API-disabled errors in _get_directory_info will bubble up and crash standard ls/exists/info calls. To get the exact error details without breaking the FileNotFoundError API contract, can we use exception chaining? We can catch (FileNotFoundError, Exception) as e and do raise original_err from e—this keeps the traceback fully transparent while preventing crashes.

zhixiangli: Why should we suppress the error instead of failing loudly—such as crashing—and forcing the user to make the proper configurations?

Mahalaxmibejugam: The _list_objects method is used in ls(), which in turn supports methods like walk() that currently only handle FileNotFoundError.

list_objects() is overridden to addresses the scenario where we want to return [] for empty directories in HNS buckets. As we begin throwing new errors from these methods, let's verify that we aren't breaking compatibility with standard bucket behaviour.

PR #920, opened 2026-06-22
Fix silent truncation on short reads in zonal bucket downloads

zhixiangli:
Fixes a bug where a short read from GCS would silently truncate the target local file and return success. Now, if the amount of bytes written differs from the expected file size, a `ClientError` is raised, stopping the silent truncation and leaving the target local system clean by removing the corrupted partial file.

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -1655,7 +1655,7 @@ class ExtendedGcsFileSystem(HnsDirCacheUpdater, GCSFileSystem):
                         callback.relative_update(len(data))
 
                 if offset != size:
-                    raise aiohttp.client_exceptions.ClientError(
+                    raise aiohttp.ClientError(
                         f"Expected {size} bytes, but only received {offset} bytes"
                     )
         except Exception as e:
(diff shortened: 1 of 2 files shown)


gcsfs/extended_gcsfs.py
```diff
@@ -1652,6 +1653,11 @@ async def _get_file_request(
                         f2.write(data)
                         offset += len(data)
                         callback.relative_update(len(data))
+
+                if offset != size:
+                    raise aiohttp.ClientError(
```
Yonghui-Lee, line 1658: It might be better to move the validation to zb_hns_utils.download_range and zb_hns_utils.download_ranges. It protects all consumers of these lower-level helpers, including cat_file, memory-buffer reads, and any future calling paths.

zhixiangli: This check is better kept at the _get_file_request level for the following reasons:

- download_range is called with a fixed chunk size that over-requests past EOF, and mid-stream short reads are resumable. Therefore, raising an error per range would cause false positives on the final chunk and partial reads. This is why the helper only issues a warning.

- The offset != size check is a whole-object validation that requires the authoritative size. While _get_file_request resolves this (with a _info() fallback), the helper cannot do so reliably for unfinalized objects.

- cat_file and memory reads do not use these helpers—they call mrd.download_ranges with DirectMemmoveBuffer / PartialView, which already raise a BufferError on underfill.

- download_ranges has no production callers yet.

Yonghui-Lee (approved):

PR #921, opened 2026-06-22
Fix generation threading in gcsfs url and concurrent fetches

zhixiangli:
This branch updates `url` and `_cat_file_sequential` to properly process and pass `generation` query arguments to ensure concurrent fetches download a consistent object version.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -1173,10 +1173,11 @@ class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
         else:
             return [o["name"] for o in out]
 
-    def url(self, path):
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -1173,10 +1173,11 @@ async def _ls(
         else:
             return [o["name"] for o in out]
 
-    def url(self, path):
+    def url(self, path, generation=None):
```
Yonghui-Lee, line 1176: I think we should also change `GCSFile.url`.

Yonghui-Lee (changes requested):

Yonghui-Lee (approved):

PR #928, opened 2026-06-24
Use memoryview to enable zero-copy in _pipe_file

Yonghui-Lee:
This commit wraps the input `data` in a `memoryview` before slicing. Slicing a `memoryview` returns a zero-copy sub-view of the underlying buffer, which is natively and efficiently accepted by `AsyncAppendableObjectWriter.append` .

## Benchmark

* **File Size**: 2 GB
* **Chunk Size**: 50 MiB
* **Rounds**: 3
* **Concurrency**: 1 process, 1 thread

### Performance Comparison

| Metric | Before (Unoptimized) | After (memoryview Zero-Copy) | Impact / Improvement |
| :--- | :---: | :---: | :---: |
| **Mean Latency (s)** | `8.6438` s | `7.8415` s | **-9.28%** |
| **Mean Throughput (MiB/s)** | `236.93` MiB/s | `261.18` MiB/s | **+10.24%** |
| **Max CPU (%)** | `1.81` % | `0.88` % | **-51.38%** |
| **Max Memory (MiB)** | `2296.23` MiB | `2296.78` MiB | Negligible |

### DIFF
--- a/gcsfs/extended_gcsfs.py
+++ b/gcsfs/extended_gcsfs.py
@@ -1572,8 +1572,9 @@ class ExtendedGcsFileSystem(HnsDirCacheUpdater, GCSFileSystem):
         # Works for both 'overwrite' and 'create' modes
         writer = await zb_hns_utils.init_aaow(self.grpc_client, bucket, key)
         try:
-            for i in range(0, len(data), chunksize):
-                await writer.append(data[i : i + chunksize])
+            with memoryview(data) as data_view:
+                for i in range(0, len(data_view), chunksize):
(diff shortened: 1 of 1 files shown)


ankitaluthra1:
Can you please add before and after numbers for pipe ? including cpu and memory, I believe the difference would be more visible in large files, so that might be a better data to add in the PR

Yonghui-Lee:
> Can you please add before and after numbers for pipe ? including cpu and memory, I believe the difference would be more visible in large files, so that might be a better data to add in the PR

Added the benchmark result to PR's description.

zhixiangli (approved):

PR #929, opened 2026-06-24
Cache storage.Client per-thread in GCSFileSystem

Yonghui-Lee:
The `sign` method was instantiating a new synchronous `storage.Client` on every single call. Creating clients is highly expensive due to the overhead of credential resolution.

This commit introduces a thread-local cache (`_thread_local_storage_client`) on the `GCSFileSystem` instance using `threading.local()`. The client is lazily initialized on the first call to `sign()` within any given thread and cached as the `.instance` attribute.

Subsequent calls to `sign()` on the same thread reuse the cached client, completely eliminating the redundant instantiation overhead. Using a thread-local container guarantees full thread-safety, ensuring threads do not share the underlying connection pool or transport.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -11,6 +11,7 @@ import os
 import posixpath
 import re
 import sys
+import threading
 import uuid
 import warnings
 import weakref
@@ -351,6 +352,7 @@ class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
         self.session_kwargs = session_kwargs or {}
         self.default_location = default_location
         self.version_aware = version_aware
+        self._thread_local_storage_client = threading.local()
 
         if check_connection:
             warnings.warn(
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -11,6 +11,7 @@
 import posixpath
 import re
 import sys
+import threading
 import uuid
 import warnings
 import weakref
```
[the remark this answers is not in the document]
ankitaluthra1: lets make sure all testing scenarios including pickling are added for this change like integration tests which run on actual bucket for single process multi thread env and multi process with single thread env and multi thread with multi thread env, to make sure this is not breaking for any scenario

Also, please update PR desciprion with before and after numbers using real world scneario which is benefitted from this

gcsfs/core.py
```diff
@@ -2269,12 +2271,15 @@ def sign(self, path, expiration=100, **kwargs):
...
-        from google.cloud import storage
+        client = getattr(self._thread_local_storage_client, "instance", None)
+        if client is None:
+            from google.cloud import storage
 
-        client = storage.Client(
-            credentials=self.credentials.credentials,
-            project=self.project,
-        )
+            client = storage.Client(
```
martindurant, line 2278: Is this Client fundamentally different from the one used for IO elsewhere? Can some part of the IO client be used instead, since that's running on a known asyncio thread anyway?

Yonghui-Lee: It's different from IO client because it contains a complex cryptographic process. https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/_signing.py

PR #930, opened 2026-06-24
Implement zero-cost local backward seeks in PrefetchConsumer

Yonghui-Lee:
Currently, any backward seek in the `BackgroundPrefetcher` is treated as a "Hard Seek". This completely stops the background downloader, cancels all active tasks, flushes the queue, discards the buffer, and restarts a new network request at the new offset.

This PR implements a **zero-cost local seek optimization** for backward seeks. If the seek target offset falls within the currently active buffered block in `PrefetchConsumer` (between `block_offset` and `self.user_offset`), we perform a local seek by simply adjusting the buffer index (`_current_block_idx`) synchronously. This keeps the prefetch pipeline running without any network interruption or producer pipeline restarts.

### DIFF
--- a/gcsfs/prefetcher.py
+++ b/gcsfs/prefetcher.py
@@ -755,6 +755,7 @@ class BackgroundPrefetcher:
 
                 logger.debug("Executing _async_fetch for range %d - %d.", start, end)
 
+                block_offset = self.consumer.offset - self.consumer._current_block_idx
                 # If the prefetcher is in error state, let's do a hard seek to start offset.
                 if self._error:
                     logger.info(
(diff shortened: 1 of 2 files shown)


gcsfs/prefetcher.py
```diff
@@ -755,6 +755,7 @@ async def _async_fetch(self, start, end):
 
                 logger.debug("Executing _async_fetch for range %d - %d.", start, end)
 
+                block_offset = self.consumer.offset - self.consumer._current_block_idx
```
zhixiangli: nit: move this statement to the elif branch that uses it.

gcsfs/prefetcher.py
```diff
@@ -773,6 +774,19 @@ async def _async_fetch(self, start, end):
                         skip_amount = start - self.user_offset
                         await self.consumer.skip(skip_amount)
                         self.user_offset = start
+                    elif block_offset <= start < self.user_offset:
```
zhixiangli: nit: It's not a bug, but by definition, consumer.offset is probably preferred over user_offset, as we should use the consumer's actual position and block length as the bound.

gcsfs/prefetcher.py
Yonghui-Lee: Done

gcsfs/prefetcher.py
Yonghui-Lee: Done

zhixiangli (approved):

PR #935, opened 2026-06-24
macrobenchmarks: add checkpoint seeding and reduce emulator ranks to 4

zhixiangli:
- Avoid OOM: Reduce default ranksPerNode from 8 to 4 in the CPU emulator workload. Checkpoint restoration doubles peak RSS to ~92GB/rank, which OOMs the 720GB c4-standard-192 node when running 8 ranks.
- Automated Seeding: Add a seed-checkpoint step to Cloud Build. If enabled (_SEED_CHECKPOINT=true) and no external checkpoint is provided, it runs a quick 1-step workload to generate a format-exact seed checkpoint, which is then restored in the main run.
- Refactoring: Extract shared JobSet polling/diagnostic logic into a wait_for_jobset helper in lib.sh, used by both seed and run steps.
- Misc: Update RUN_ID to use SHORT_BUILD_ID; update cleanup and variable initialization to support the seeding step.

### DIFF
--- a/gcsfs/tests/perf/macrobenchmarks/workloads/hf-pytorch-lightning-cpu/helm_chart/values_base.yaml
+++ b/gcsfs/tests/perf/macrobenchmarks/workloads/hf-pytorch-lightning-cpu/helm_chart/values_base.yaml
@@ -1,5 +1,13 @@
 # Default values for the gcsfs-only CPU emulator.
-# 2 c4-standard-192 nodes x 8 procs per node = 16 ranks.
+# 2 c4-standard-192 nodes x 4 procs per node = 8 ranks.
+#
+# Why 4 ranks/node (not 8): when the run restores a checkpoint, every DDP rank
+# torch.load's the full ~48GB state (16GB model + 32GB AdamW) while still
+# holding the freshly-initialized model+optimizer, so per-rank peak RSS roughly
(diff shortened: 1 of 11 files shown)


gcsfs/tests/perf/macrobenchmarks/workloads/hf-pytorch-lightning-cpu/helm_chart/values_base.yaml
```diff
@@ -1,5 +1,13 @@
 # Default values for the gcsfs-only CPU emulator.
-# 2 c4-standard-192 nodes x 8 procs per node = 16 ranks.
+# 2 c4-standard-192 nodes x 4 procs per node = 8 ranks.
+#
+# Why 4 ranks/node (not 8): when the run restores a checkpoint, every DDP rank
```
yuxin00j, line 4: Maybe we don't need to expose these details in comments?

zhixiangli: Yes, this portion of change actually conflicts with https://github.com/fsspec/gcsfs/pull/936, so I will clean that up when resolving the conflicts.

yuxin00j (approved):

PR #939, opened 2026-06-26
test(perf): add microbenchmarks for put operations

zhixiangli:
Introduces performance microbenchmarks for GCS upload (put) operations. Includes:
- PutConfigurator and YAML configurations for single-threaded and multi-process put benchmarks.
- Single-threaded and multi-process put benchmark test cases.
- Unit tests for PutConfigurator.
- Fixtures and setup in conftest.py.
- Updated README documentation.
- Fixed ResourceMonitor import and multi-process argument mismatch.

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/put/test_put.py
+++ b/gcsfs/tests/perf/microbenchmarks/put/test_put.py
@@ -1,0 +1,117 @@
+import logging
+import time
+from concurrent.futures import ThreadPoolExecutor
+
+import pytest
+
+from gcsfs.tests.perf.microbenchmarks.put.configs import get_put_benchmark_cases
+from gcsfs.tests.perf.microbenchmarks.runner import (
+    filter_test_cases,
+    run_multi_process,
+    run_single_threaded,
+)
+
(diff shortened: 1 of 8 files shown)


gcsfs/tests/perf/microbenchmarks/put/test_put.py
```diff
@@ -0,0 +1,117 @@
...
+    run_single_threaded,
+)
+
+BENCHMARK_GROUP = "put"
+
+
+def _put_op(gcs, local_path, remote_path, chunk_size):
+    """Upload a local file to a single remote path."""
+    try:
+        gcs.put(local_path, remote_path, chunksize=chunk_size)
```
yuxin00j, line 20: Should we pass blocksize?

zhixiangli: put routes through _put_file, which takes chunksize (used directly as the resumable-upload part size). block_size is only a  parameter of the open()/GCSFile path, so passing it to put would land in **kwargs and be ignored.

yuxin00j (approved):

PR #949, opened 2026-07-02
fix(macrobenchmarks): use random optimizer moments, not zeros, in CPU sim

zhixiangli:
This pull request updates the _materialize_adamw_state function in the Llama 3.1 8B CPU simulation benchmark to initialize AdamW optimizer states with random values instead of zeros, preventing trivial data compression and ensuring realistic IO benchmark results. I have no feedback to provide as there are no review comments.


yuxin00j (approved):
nit: remove the last sentence in the PR description?

PR #955, opened 2026-07-07
ci: move build and release logic to reusable build-and-release workflow

Yonghui-Lee:
I separated the package building and GitHub release creation concerns from the PyPI publishing steps by refactoring the release workflow files to use a new reusable builder:

1.  **Isolated Build & GitHub Release:** Created a new reusable builder workflow (`build-and-release.yml`) that handles running CI tests, compiling artifacts, and publishing the GitHub Release. This workflow does not handle publishing to PyPI.
2.  **Top-Level PyPI Publishing:** Updated both entrypoint workflows (`release.yml` and `release-on-merge.yml`) to call the new reusable builder first, download the built artifacts, and then execute the PyPI publishing step (`pypa/gh-action-pypi-publish`) directly in their own top-level job context.

### DIFF
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -4,7 +4,8 @@ name: Release on Merge
 #   1. Triggers when a Pull Request modifying changelog.rst is closed and merged into main.
 #   2. If it is a release PR (has 'autorelease' label), it extracts the version.
 #   3. Creates and pushes the Git tag for that version.
-#   4. Calls the reusable release.yml workflow to build and publish to PyPI.
+#   4. Calls the reusable build-and-release.yml workflow to build and create GitHub Release.
+#   5. Publishes to PyPI.
 on:
   pull_request:
     types: [closed]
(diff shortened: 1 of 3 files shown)


.github/workflows/release-on-merge.yml
```diff
@@ -57,15 +58,30 @@ jobs:
...
       checks: read
       statuses: read
-    uses: ./.github/workflows/release.yml
+    uses: ./.github/workflows/build-and-release.yml
     with:
       ref: ${{ needs.tag.outputs.version }}
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  publish:
```
zhixiangli, line 73: Should we exclude the dev release here?

.github/workflows/release.yml
```diff
@@ -1,97 +1,24 @@
...
-          files: dist/*
+      checks: read
+      statuses: read
+    uses: ./.github/workflows/build-and-release.yml
+    with:
+      ref: ${{ github.ref_name }}
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   publish:
```
zhixiangli, line 23: Should we exclude the dev release here?

.github/workflows/release-on-merge.yml
Yonghui-Lee: Dev release can be shared. Only Pypi published can't be placed in the sub workflow.

zhixiangli (approved):

PR #960, opened 2026-07-09
Fix SQLite DB locking race condition in create_buckets.sh

ccbcqbz:
Running multiple parallel gcloud storage background tasks in an uninitialized ~/.config/gcloud directory occasionally triggers SQLite DB locking exceptions, resulting in failed Cloud Build steps. This PR adds a synchronous initialization command at the beginning of the bucket creation script to serialize DB creation and prevent the race condition.


ccbcqbz:
Thank you for the review! I have addressed the feedback by removing the stderr redirection and the `|| true` error suppression from `gcloud config list` so that failures during configuration initialization will correctly report stderr and fail the build fast.

Yonghui-Lee (approved):

PR #973, opened 2026-07-20
Fix VM and bucket deletion logic in Cloud Build cleanup

yuxin00j:
Removes invalid arguments in VM teardown, propagates exit codes for background bucket removals, and ensures proper project references to fix IAM permission failures during cleanup.

### DIFF
--- a/cloudbuild/cleanup.sh
+++ b/cloudbuild/cleanup.sh
@@ -7,6 +7,7 @@ WORKERS_STANDARD="${WORKERS_STANDARD:-1}"
 WORKERS_HNS="${WORKERS_HNS:-1}"
 WORKERS_ZONAL="${WORKERS_ZONAL:-1}"
 WORKERS_ZONAL_CORE="${WORKERS_ZONAL_CORE:-1}"
+BACKGROUND_PIDS=()
 
 worker_count() {
     local workers="$1"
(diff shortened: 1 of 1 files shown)


cloudbuild/cleanup.sh
```diff
@@ -27,25 +28,44 @@ worker_suffixes() {
...
+}
+
 delete_bucket_variants() {
     local workers="$1"
     local bucket_base="$2"
     shift 2
 
     while IFS= read -r suffix; do
-        gcloud storage rm --recursive "gs://${bucket_base}${suffix}" "$@" || true &
+        if gcloud storage buckets describe "gs://${bucket_base}${suffix}" --project="${PROJECT_ID}" "$@" &>/dev/null; then
```
zhixiangli, line 52: Why was this check added, and does its failure cause a leak?

cloudbuild/cleanup.sh
```diff
@@ -27,25 +28,44 @@ worker_suffixes() {
...
     done < <(worker_suffixes "${workers}")
 }
 
 echo "--- Deleting VM ---"
-# --async: the delete operation is durable server-side and completes regardless of
-# whether this build step is still running, so we don't block cleanup on it. (Bucket
-# deletes below must stay synchronous; killing those mid-delete would orphan objects.)
-gcloud compute instances delete "gcsfs-test-vm-${SHORT_BUILD_ID}" --zone="${ZONE}" --quiet --async || true
+# The delete operation is run in the background so we don't block bucket cleanup on it.
+if gcloud compute instances describe "gcsfs-test-vm-${SHORT_BUILD_ID}" --project="${PROJECT_ID}" --zone="${ZONE}" &>/dev/null; then
```
zhixiangli, line 61: Why was this check added, and does its failure cause a leak?

cloudbuild/cleanup.sh
yuxin00j: We removed the `|| true` to catch cleanup error. It the bucket has not been created, we do not want to report an error.

cloudbuild/cleanup.sh
yuxin00j: The reason is the same. here. It the vm has not been created, we do not want to report an error.

zhixiangli (approved):

PR #988, opened 2026-07-24
fix(benchmark): use fsspec.open instead of shared IPC instance to avoid multiprocessing timeout

yuxin00j:
Fix multiprocessing deadlock in `open` microbenchmark

**Problem**
The `test_open_multi_process` microbenchmark was timing out due to a deadlock. All 64 subprocesses were inadvertently sharing the same gRPC HTTP/2 connection instantiated in the parent process, which caused stream exhaustion and crashed the `asyncio` loop.

**Changes**
Swapped the shared `gcs.open(...)` call with `with fsspec.open(...)` in the `_open_op` worker process. This forces `fsspec` to dynamically resolve and cache a fresh, dedicated filesystem connection per worker process.

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/open/test_open.py
+++ b/gcsfs/tests/perf/microbenchmarks/open/test_open.py
@@ -14,12 +14,14 @@ from gcsfs.tests.perf.microbenchmarks.runner import (
 
 BENCHMARK_GROUP = "open"
 
+import fsspec
+
 
 def _open_op(gcs, path):
     start_time = time.perf_counter()
     try:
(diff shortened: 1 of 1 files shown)


Yonghui-Lee (approved):

Yonghui-Lee:
Other multi process micro benchmarks also have this problem. The reason they did not deadlock as frequently as test_open_multi_process is primarily due to lower concurrency and longer-running operations which are less likely to overwhelm the connection pool or trigger stream exhaustion.

Maybe we can discard connection sharing for all multi-process benchmarks by dynamically instantiating the GCS filesystem inside the child process.

PR #990, opened 2026-07-29
fix(hns): cache root bucket properly in _get_dirs_and_update_cache

yuxin00j:
## Summary
Fixes a performance bottleneck where `_get_dirs_and_update_cache` skipped updating `fsspec.dircache` when listing the root bucket (`dir_key == ""`) in Hierarchical Namespace (HNS) enabled buckets. This caused repeated metadata cache misses and redundant GCP network requests when loading multi-file datasets (e.g., Hugging Face `datasets` / Parquet).

### Root Cause & Technical Fix
* **Before:** In `gcsfs/core.py` (`_get_dirs_and_update_cache`), the parent directory traversal loop terminated early when encountering the root bucket (`not dir_key`). Because the root bucket listing was never written to `self.dircache`, subsequent `fs.info("bucket/file")` calls triggered cache misses and made synchronous HTTP network requests to GCS for every file.
* **Fix:** 
  * Updated `_get_dirs_and_update_cache` so that `cache_entries` is populated for root buckets (`"bucket"`), ensuring the directory contents are cached in `self.dircache["bucket"]`.
  * Guarded `dirs[parent]` assignment with `if dir_key:` so that root buckets are not incorrectly synthesized as subdirectories inside themselves (which previously caused `409 Conflict: The bucket you tried to delete is not empty` during test teardown).

### DIFF
--- a/gcsfs/tests/integration/test_extended_hns.py
+++ b/gcsfs/tests/integration/test_extended_hns.py
@@ -1328,6 +1328,21 @@ class TestExtendedGcsFileSystemFindIntegration:
         assert test_structure["nested_dir"] in dir_with_files_listing
 
         # Check content of the 'nested_dir' cache
+
+    @pytest.mark.parametrize("withdirs_param", [True, False])
+    def test_find_updates_dircache_for_root_bucket(
+        self, gcs_hns, test_structure, withdirs_param
+    ):
+        """Test that find() populates the dircache for the root bucket itself."""
+        root_bucket = TEST_HNS_BUCKET
+        gcs_hns.invalidate_cache()
+        assert not gcs_hns.dircache
+
+        # Run find on the root bucket to populate the cache
+        gcs_hns.find(root_bucket, withdirs=withdirs_param)
+
+        # Verify that the cache is now populated for the root bucket
+        assert root_bucket in gcs_hns.dircache
         nested_dir_listing = {
             d["name"] for d in gcs_hns.dircache[test_structure["nested_dir"]]
         }
--- a/gcsfs/tests/integration/test_extended_hns.py
+++ b/gcsfs/tests/integration/test_extended_hns.py
@@ -1328,6 +1328,16 @@ class TestExtendedGcsFileSystemFindIntegration:
         assert test_structure["nested_dir"] in dir_with_files_listing
 
         # Check content of the 'nested_dir' cache
+        nested_dir_listing = {
+            d["name"] for d in gcs_hns.dircache[test_structure["nested_dir"]]
+        }
+        assert test_structure["nested_file"] in nested_dir_listing
+
+        # Check content of the 'empty_dir' cache
+        empty_dir_listing = {
+            d["name"] for d in gcs_hns.dircache[test_structure["empty_dir"]]
+        }
+        assert not empty_dir_listing
 
     @pytest.mark.parametrize("withdirs_param", [True, False])
     def test_find_updates_dircache_for_root_bucket(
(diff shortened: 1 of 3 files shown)


gcsfs/tests/integration/test_extended_hns.py
```diff
@@ -1328,6 +1328,20 @@ def test_find_updates_dircache_without_prefix(
...
+        root_bucket = TEST_HNS_BUCKET
+        base_dir = test_structure["base_dir"]
+        gcs_hns.invalidate_cache()
+        assert not gcs_hns.dircache
+
+        # Run find on the root bucket to populate the cache
+        gcs_hns.find(root_bucket, withdirs=withdirs_param)
+
+        # Verify that the cache is now populated for the root bucket
+        assert root_bucket in gcs_hns.dircache
```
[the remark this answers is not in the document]
Yonghui-Lee: @yuxin00j may I know is it intended?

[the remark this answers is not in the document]
yuxin00j: Good catch! That was an accidental insertion. I've restored the original assertions to `test_find_updates_dircache_without_prefix` and placed `test_find_updates_dircache_for_root_bucket` cleanly after it in commit 0d7700a.

Yonghui-Lee (approved):

gcsfs/tests/test_core.py
Mahalaxmibejugam, line 1: Why are these changes necessary for this PR? Were the tests failing due to the updates in the caching logic? Also, if the GCS instance passed as a parameter is not being used, why is it being included?

gcsfs/tests/integration/test_extended_hns.py
```diff
@@ -1339,6 +1339,21 @@ def test_find_updates_dircache_without_prefix(
...
+        """Test that find() populates the dircache for the root bucket itself."""
+        root_bucket = TEST_HNS_BUCKET
+        gcs_hns.invalidate_cache()
+        assert not gcs_hns.dircache
+
+        # Run find on the root bucket to populate the cache
+        gcs_hns.find(root_bucket, withdirs=withdirs_param)
+
+        # Verify that the cache is now populated for the root bucket
+        assert root_bucket in gcs_hns.dircache
```
Mahalaxmibejugam, line 1355: Should we also add an assert on the value of the root_bucket key?

gcsfs/tests/test_core.py
yuxin00j: You are correct, these changes were orthogonal and unnecessary. I mistakenly added the fixture while running these specific tests in local isolation during debugging. I've completely reverted `test_core.py` in `4cc5ff0`.

gcsfs/tests/integration/test_extended_hns.py
yuxin00j: I've added an assertion in `c7069ab` to verify that `test_structure["base_dir"]`, etc. are successfully populated in the cached listing for `root_bucket`.

gcsfs/tests/integration/test_extended_hns.py
```diff
@@ -1339,6 +1339,25 @@ def test_find_updates_dircache_without_prefix(
...
+        root_bucket_listing = {
```
Mahalaxmibejugam, line 1356: let's assert the complete root_bucket_listing to make sure no extra items are getting added in the values.

yuxin00j: Done.

zhixiangli (approved):

PR #992, opened 2026-07-31
Makes prefetcher none cache default

ankitaluthra1:


docs/source/prefetcher.rst
```diff
@@ -71,17 +70,30 @@ Interaction with GCSFile
...
---------------------
+Feature Configuration & Disabling
+---------------------------------
 
-To use this architecture, set the following environment variables:
+Adaptive prefetching is enabled by default when ``cache_type`` is not explicitly set by the user, using ``DEFAULT_GCSFS_CONCURRENCY=4``.
+
+Prefetching can be disabled in three ways:
+
+1. Explicitly specify a ``cache_type`` when opening a file (e.g., ``cache_type="readahead"`` or any other cache_type):
```
googlyrahman: I like this idea, this ensures that we do not increase the egress for users with cache_type="none"

gcsfs/core.py
```diff
@@ -2305,6 +2305,34 @@ def sign(self, path, expiration=100, **kwargs):
 GoogleCredentials.load_tokens()
 
 
+def _get_prefetcher_and_cache_value(cache_type, kwargs):
+    """
+    Resolves effective cache_type and whether prefetch reader should be enabled.
+
+    Rules:
+    - Prefetcher is only used when cache_type is not explicitly set by the user (cache_type is None).
```
googlyrahman: People might confused with None, and "none" in future discussions, maybe a comment might help that cache_type="none" is not same as cache_type=None.

gcsfs/core.py
```diff
@@ -2305,6 +2305,34 @@ def sign(self, path, expiration=100, **kwargs):
...
+    - If user sets any cache_type (cache_type is not None), prefetcher is always disabled.
+    - By default when cache_type is not set, cache_type is "none".
+    """
+    if cache_type is not None:
+        use_prefetch_reader = False
```
googlyrahman, line 2318: Not important, because we do not have any major adoption of this feature, but just thinking out loud:

Any user who followed the previous documentation of prefetcher and mapped their codebase to use gcs.open(..., cache_type="none") while leaving USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING=true in their environment variables will ironically have their prefetcher silently disabled when upgrading.

gcsfs/core.py
```diff
@@ -2305,6 +2305,34 @@ def sign(self, path, expiration=100, **kwargs):
...
+    else:
+        cache_type = "none"
```
googlyrahman: Not important, because we're changing the cache_type to "none", and will explicitly document it in release notes, but just thinking out loud:

If a user disables the prefetcher by setting `USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING=false` or passing `use_experimental_adaptive_prefetching=False`, they will silently lose all caching. Before this PR, GCSFile fell back to fsspec's "readahead" by default, and ZonalFile fell back to "readahead_chunked". Users opting out of the experimental prefetcher may suffer read performance degradation because they are now routed to "none".

googlyrahman (approved):

gcsfs/core.py
ankitaluthra1: Yes, Noted. I will add a call out in release notes for this change

gcsfs/core.py
ankitaluthra1: Good point, I think it makes sense to fallback to simple readahead when prefetcher is false, so behaviour is reverted to current behaviour when prefetcher is disabled. Updated, PTAL

gcsfs/core.py
googlyrahman: This also need a change in our microbenchmarks, I remember they explicitly set this cache_type="none".

PR #993, opened 2026-08-03
feat: use fsspec adaptive-readahead cache and remove internal prefetcher

raj-prince:
With the release of fsspec 2026.9.0, the adaptive prefetching engine and AdaptiveReadaheadCache originally implemented in gcsfs have been officially upstreamed and merged into fsspec.

This PR removes gcsfs's internal prefetcher implementation and delegates background prefetching directly to fsspec's native cache layer via cache_type="adaptive".

**Changes:**
- Bumped fsspec version >= 2026.9.0.
- Removed prefetcher (and corresponding test files) implementation from gcsfs.
- GCSFile now defaults to cache_type="adaptive" when cache_type is not explicitly set by the user.
- Removed the experimental environment variable USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING and keyword argument use_experimental_adaptive_prefetching. Users configure or disable prefetching via standard fsspec arguments (e.g., cache_type="readahead" or cache_type="none").
- Automatically passes concurrency and max_prefetch_size to cache_options when cache_type == "adaptive".
- Added few important unit test to increase the project test coverage which came down because of removal of prefetcher changes.
- Added check in _fetch_range to return empty content in case of invalid range, earlier prefetcher check was used for this validation.

**Before and After perf changes (10 GiB seq read 16 MB IO) - no regression:** 

| Bucket / Storage Class | Configuration | Run 1 | Run 2 | Run 3 | Mean Throughput | Mean Time (10 GiB) | Delta |
| :--- | :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| **Rapid Storage**<br>(`cloudscope-demo-rapid`) | **Before** (`upstream/main`) | 1,981.2 MB/s | 2,009.0 MB/s | 2,035.3 MB/s | **2,008.5 MB/s** (1.96 GiB/s) | **5.10s** | — |
| | **After** (PR #993) | 1,949.4 MB/s | 2,104.7 MB/s | 2,104.2 MB/s | **2,052.8 MB/s** (2.00 GiB/s) | **4.99s** | **+2.2%** *(Parity)* |
| **Standard Regional**<br>(`cloudscope-demo-regional`) | **Before** (`upstream/main`) | 672.0 MB/s | 683.0 MB/s | 643.6 MB/s | **666.2 MB/s** (0.65 GiB/s) | **15.38s** | — |
| | **After** (PR #993) | 677.6 MB/s | 681.6 MB/s | 675.7 MB/s | **678.3 MB/s** (0.66 GiB/s) | **15.10s** | **+1.8%** *(Parity)* |


gcsfs/core.py
zhixiangli, line 2509: Integrate self._async_fetch_range into the prefetcher producer; otherwise, _async_fetch_range is bypassed, the split_factor is ignored, and an unnecessary asyncio.to_thread round-trip is introduced.

gcsfs/core.py
```diff
@@ -2730,9 +2706,11 @@ def _fetch_range(self, start=None, end=None):
         start, end : None or integers
             if not both None, fetch only given range
         """
+        if start is not None and self.size is not None and start >= self.size:
+            return b""
+        if start is not None and end is not None and start >= end:
```
zhixiangli: The check for `start >= end` in `_fetch_range` for both `GCSFile` and `ZonalFile` is causing issues with negative end offsets.

gcsfs/core.py
```diff
@@ -2505,6 +2486,15 @@ def __init__(
         cache_type, use_prefetch_reader, self.cache_source = (
             _get_prefetcher_and_cache_config(cache_type, kwargs)
         )
+        cache_options = dict(cache_options or {})
+        if cache_type == "adaptive":
+            if "concurrency" not in cache_options:
+                cache_options["concurrency"] = self.concurrency
+            if (
+                "max_prefetch_size" not in cache_options
+                and "max_prefetch_size" in kwargs
```
zhixiangli: Previously, GCSFile defaulted max_prefetch_size to MAX_PREFETCH_SIZE (256 MiB). Here, cache_options["max_prefetch_size"] is only set if "max_prefetch_size" was explicitly passed in kwargs, so AdaptiveReadaheadCache defaults max_prefetch_size to None

gcsfs/core.py
zhixiangli, line 2403: `AdaptiveReadaheadCache.__del__` executes `sync_teardown` synchronously on the GC thread, which undermines the `_defer_close` mechanism in `GCSFile.__del__`.

zhixiangli: self.cache_type assigned after super().__init__() causes AttributeError for cache_type="all"

raj-prince: Nice catch. Seems like an existing issue, we might need to patch this in previous release.

gcsfs/core.py
raj-prince: Created an utility and used it across different place.

gcsfs/utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+    Handles negative offsets following Python slice semantics when size is known.
+    """
+    if size is not None:
+        start = max(0, size + start) if (start is not None and start < 0) else start
+        end = max(0, size + end) if (end is not None and end < 0) else end
+        if start is not None and start >= size:
+            return True
+
+    if start is None or end is None or start < 0 or end < 0:
```
zhixiangli: nit: When start is None, is_empty_range returns False even when end == 0 (e.g. start=None, end=0 or start=None, end=-size) or size == 0. In _cat_file_sequential(path, start=None, end=0), this causes _process_limits to send an invalid Range: bytes=0--1 header, causing GCS to return the entire file instead of b"".

gcsfs/tests/test_core.py
```diff
@@ -3021,9 +3056,7 @@ def test_gcsfile_multithreaded_read_integrity(gcs):
     data = os.urandom(file_size)
     gcs.pipe(fn, data)
 
-    with gcs.open(
-        fn, "rb", use_experimental_adaptive_prefetching=True, block_size=2 * 1024 * 1024
-    ) as f:
+    with gcs.open(fn, "rb", block_size=2 * 1024 * 1024) as f:
```
zhixiangli: nit: this test bypasses AdaptiveReadaheadCache.

gcsfs/core.py
```diff
@@ -2339,38 +2337,19 @@ def sign(self, path, expiration=100, **kwargs):
 GoogleCredentials.load_tokens()
 
 
-def _get_prefetcher_and_cache_config(cache_type, kwargs):
+def _get_prefetcher_and_cache_config(cache_type, kwargs=None):
```
zhixiangli: nit: it still returns a 3-tuple (cache_type, False, cache_source) where the second element (use_prefetch_reader) and the kwargs argument are unused across all callers.

zhixiangli (approved):

gcsfs/tests/test_core.py
raj-prince: Removing this test as already validated via - test_gcsfile_prefetch_sequential_integrity (which will do concurrency underneath).

docs/source/changelog.rst
```diff
@@ -29,7 +29,7 @@ GCSFS prefetcher adapts to workload read IO patterns. It tracks the rolling aver
 * **Multi-worker dataloader scaling.** The prefetcher manufactures its own parallelism per worker instead of relying on process count alone.
 * **Accelerate the throughput even further with Rapid Buckets.** With Rapid Buckets single node throughput reaches 21 GiB/s with  16-process sequentially reading at 16 MiB I/O compared to standard buckets with 48processes.
 
-**Adaptive prefetcher is enabled by default** when cache_type is not explicitly set and concurrency value is set at 4(DEFAULT_GCSFS_CONCURRENCY=4) for both Standard and Rapid buckets. You can disable adaptive prefetcher by setting an explicit cache_type, or by setting  USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING='false', or by passing use_experimental_adaptive_prefetching=False to open() call.
```
ankitaluthra1: Need rewording to clear communicate that there is a new cache type which is not default, plus its incorrectly conveying not setting cache is only way, but cache_type="adaptive" can be explicitly set

gcsfs/core.py
```diff
@@ -2516,14 +2507,30 @@ def __init__(
...
         self.checker = get_consistency_checker(consistency)
 
+        cache = getattr(self, "cache", None)
+        prefetcher = getattr(cache, "_prefetcher", None)
+        if prefetcher is not None:
+            # Wire GCSFile/ZonalFile's native async range fetcher into the prefetcher producer.
+            # Otherwise, _async_fetch_range is bypassed, split_factor is ignored, and
+            # fsspec's default fetcher performs an unnecessary asyncio.to_thread round-trip.
+            # TODO: Remove this direct override once fsspec natively supports passing an async_fetcher.
+            prefetcher.fetcher = self._async_fetch_range
```
ankitaluthra1: Is prefetcher.fetcher used anywhere ? Corresponding tests are also validating without usage, should'nt it be prefetcher = _get_prefetcher(self) ?

docs/source/changelog.rst
```diff
@@ -29,7 +29,7 @@ GCSFS prefetcher adapts to workload read IO patterns. It tracks the rolling aver
...
+**Adaptive prefetcher is enabled by default** when cache_type is not explicitly set and concurrency value is set at 4 (DEFAULT_GCSFS_CONCURRENCY=4) for both Standard and Rapid buckets. You can disable adaptive prefetcher or choose an alternative cache strategy by setting an explicit cache_type (e.g., cache_type="readahead" or cache_type="none").
 
 **(Warning) Impact on memory:** Prefetching trades memory for throughput. Peak memory rises from ~170 MB to 600 MB on single-stream reads for 16 MB IO size and varies with requested IO sizes, and would be materially more under high process counts. Please ensure that   application memory  limits accordingly to use prefetcher without any Out of Memory(OOM) issues. To put hard limit, you can also use [user_max_prefetch_size](https://github.com/fsspec/gcsfs/blob/main/gcsfs/prefetcher.py#L154)
```
ankitaluthra1: old link

docs/source/prefetcher.rst
```diff
@@ -2,11 +2,9 @@
...
-asynchronous filesystems using the standard `cache_type=` argument. How it interacts with the
-existing cache types ("readahead", "first", etc.) remains to be decided, and in the meantime, use at your own risk.
-We intend to develop more sophisticated caching strategies, perhaps specialised to file types.
+**Note:** The adaptive prefetching engine, originally developed in ``gcsfs``, has been upstreamed directly into ``fsspec`` (starting in ``fsspec>=2026.9.0``) as ``fsspec.caching.AdaptiveReadaheadCache``. In ``gcsfs``, it is enabled by default via the ``"adaptive"`` cache type whenever ``cache_type`` is not explicitly set, using ``DEFAULT_GCSFS_CONCURRENCY=4``.
+
+To select an alternative cache or disable prefetching, pass the standard ``cache_type=`` argument (e.g., ``cache_type="readahead"``, ``cache_type="readahead_chunked"``, or ``cache_type="none"``). Standard caching options can be passed via ``cache_options`` (e.g., ``cache_options={"max_prefetch_size": 32 * 1024 * 1024}``).
 
 Additional caveats:
 - the bytes slicing/copying code uses low level (`ctypes`) calls and offloads to a dedicated thread for
```
ankitaluthra1: ctypes is not longer imported IIUC

gcsfs/core.py
```diff
@@ -2502,9 +2480,22 @@ def __init__(
             raise OSError("Attempt to open a bucket")
         self.generation = _coalesce_generation(generation, path_generation)
         self.concurrency = kwargs.get("concurrency", DEFAULT_CONCURRENCY)
-        cache_type, use_prefetch_reader, self.cache_source = (
-            _get_prefetcher_and_cache_config(cache_type, kwargs)
+        self.cache_type, self.cache_source = _get_prefetcher_and_cache_config(
+            cache_type
         )
+        cache_type = self.cache_type
```
ankitaluthra1: redundant assignment

gcsfs/core.py
```diff
@@ -2502,9 +2480,22 @@ def __init__(
...
+        self.bucket = bucket
+        self.key = key
```
ankitaluthra1: why are  self.bucket = bucket
        self.key = key above init(), as its not needed

gcsfs/utils.py
```diff
@@ -0,0 +1,27 @@
+"""General utility functions for gcsfs."""
+
+from __future__ import annotations
+
+
+def is_empty_range(start: int | None, end: int | None, size: int | None = None) -> bool:
+    """Returns True if the requested byte range represents zero bytes or falls beyond EOF."""
+    if size is not None:
```
ankitaluthra1: Having hard dependency on size can lead to incosistent reads, size can be stale for Rapid buckets as its eventually consistent, size can only be confirmed if size is captured after reading file is closed and finalised

gcsfs/core.py
```diff
@@ -2773,9 +2766,13 @@ def close(self):
         self._close_impl()
 
     def _close_impl(self):
+        # TODO: Remove explicit prefetcher cleanup once fsspec handles deferred cache teardown.
+        cache = getattr(self, "cache", None)
+        prefetcher = getattr(cache, "_prefetcher", None)
+        if prefetcher is not None:
+            prefetcher.close()
+            cache._prefetcher = None
```
ankitaluthra1: why is cache._prefetcher = None needed ?

ankitaluthra1 (changes requested):

ankitaluthra1:
Also, please add perf tests results for 5  MB and IO in addition to 16 MB, so case for small more frequent read size is covered

gcsfs/tests/test_core.py
```diff
@@ -3015,41 +3049,51 @@ def test_gcsfile_prefetch_random_seek_integrity(gcs):
             assert chunk == data[start : start + length]
 
 
-def test_gcsfile_multithreaded_read_integrity(gcs):
```
ankitaluthra1: why is this test deleted ?

gcsfs/tests/test_zonal_unit.py
```diff
@@ -382,12 +385,12 @@ def test_resolve_cache_config():
 
     # 3. Neither provided (defaults resolved via _get_prefetcher_and_cache_config)
     c_type, c_source = ExtendedGcsFileSystem._resolve_cache_config({})
-    assert c_type in ("none", "readahead")
+    assert c_type in ("adaptive", "readahead")
```
ankitaluthra1: why cant assertion be more strict as its only adaptive now ?

gcsfs/utils.py
```diff
@@ -0,0 +1,27 @@
+"""General utility functions for gcsfs."""
+
+from __future__ import annotations
+
+
+def is_empty_range(start: int | None, end: int | None, size: int | None = None) -> bool:
```
ankitaluthra1: how is this method logic different from existing method `ExtendedGcsFileSystem._process_limits_to_offset_and_length` ?

PR #998, opened 2026-08-09
Track whether cache_type is explicitly set or defaulting in User-Agent header

Mahalaxmibejugam:
### Summary of Changes

This PR adds tracking for whether `cache_type` was explicitly configured by the user or assigned as a default/fallback, differentiating them in the HTTP `User-Agent` header.

#### 1. Background & Problem
Previously, `gcsfs` emitted `cache_type/<type>` in the `User-Agent` header. With the introduction of the default adaptive prefetcher (which sets `cache_type="none"`), `cache_type/none` was emitted in two opposite scenarios:
- **Default (Prefetcher Enabled)**: User left `cache_type=None`.
- **Explicit Unbuffered (Prefetcher Disabled)**: User explicitly specified `cache_type="none"`.

Similarly, `cache_type/readahead` was emitted for both explicit `cache_type="readahead"" and when falling back because prefetching was disabled.

#### 2. Solution
- Track `cache_source` (`"explicit"` vs `"default"`) internally in `_get_prefetcher_and_cache_config` and `GCSFile`.
- In `_get_headers`, append `:e` for explicit cache configurations and `:d` for defaults:
  - Default open (`open(fn, "rb")`): `cache_type/none:d`
  - Explicit none (`open(fn, "rb", cache_type="none")`): `cache_type/none:e`
  - Explicit readahead (`open(fn, "rb", cache_type="readahead")`): `cache_type/readahead:e`
  - Fallback readahead (`open(fn, "rb", use_experimental_adaptive_prefetching=False)`): `cache_type/readahead:d`
  - Explicit custom (`open(fn, "rb", cache_type="mmap")`): `cache_type/mmap:e`
- Pass `cache_type` and `cache_source` through `_cat_file_sequential`, `_request`, and `_get_headers`.

#### 3. Tests
- Updated `test_gcsfile_prefetch_and_cache_type_rules` to verify `f.cache_source` across all 5 configurations.
- Updated `test_get_headers_includes_cache_type` to test unit header formatting.
- Updated `test_user_agent_includes_cache_type_and_source_in_read` for end-to-end live request header assertions.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -181,6 +181,18 @@ def _is_directory_marker(entry):
     return entry["size"] == 0 and entry["name"].endswith("/")
 
 
+def _get_cache_type_header_value(cache_type, cache_source=None):
+    """Format cache_type and cache_source into a User-Agent header value."""
+    if not cache_type:
+        return ""
+    suffix = ""
+    if cache_source == "explicit":
+        suffix = ":e"
+    elif cache_source == "default":
+        suffix = ":d"
+    return f"cache_type/{cache_type}{suffix}"
+
+
 class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
     r"""
     Connect to Google Cloud Storage.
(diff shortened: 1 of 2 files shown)


gcsfs/core.py
```diff
@@ -469,14 +469,19 @@ def _get_params(self, kwargs):
...
         if "User-Agent" not in out:
             ua = "python-gcsfs/" + version
             if cache_type:
-                ua += f" cache_type/{cache_type}"
+                suffix = ""
+                if cache_source == "explicit":
+                    suffix = ":e"
+                elif cache_source == "default":
+                    suffix = ":d"
+                ua += f" cache_type/{cache_type}{suffix}"
```
ankitaluthra1: consider extracting a method which should return complete cache_type string along with suffix and _get_headers appending this new method output to avoid long method and better readability.

ankitaluthra1 (changes requested):

PR #1003, opened 2026-08-11
subsystembenchmarks: support model id substitution in checkpointing save

Yonghui-Lee:
This PR adds support for specifying a custom model_id in the checkpointing save subsystem benchmark via Cloud Build substitutions. It allows it to use pre-downloaded model weights hosted on Google Cloud Storage instead of downloading them from the Hugging Face Hub on every run.

### DIFF
--- a/cloudbuild/subsystembenchmarks/subsystembenchmarks-cloudbuild.yaml
+++ b/cloudbuild/subsystembenchmarks/subsystembenchmarks-cloudbuild.yaml
@@ -9,6 +9,7 @@ substitutions:
   _SWEEP_AXES: ""
   _FILTER: ""
   _HF_TOKEN: ""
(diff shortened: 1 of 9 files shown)


cloudbuild/subsystembenchmarks/subsystembenchmarks-cloudbuild.yaml
```diff
@@ -9,6 +9,7 @@ substitutions:
   _SWEEP_AXES: ""
   _FILTER: ""
   _HF_TOKEN: ""
+  _MODEL_ID: "gs://huggingface-model-weights/Llama-3.1-8B"
```
zhixiangli, line 12: Would it cause every subsystem benchmark downloading the checkpoint model?

Yonghui-Lee: I add the check of group before downloading.

PR #1009, opened 2026-08-13
fix: prevent NotImplementedError when closing GCSFile from event loop thread

zhixiangli:
In the webdataset subsystem benchmark, the file is not closed properly, so file cleanup relies on the GC.

When a GCSFile or ZonalFile is garbage collected without an explicit close(), its __del__() finalizer calls close(). If finalization executes on the background fsspec I/O loop thread, synchronous asyn.sync() calls (such as ZonalFile MRD pool / AAOW teardown and file flushing) raise NotImplementedError ("Calling sync() from within a running loop").

- Add _on_loop_thread and _defer_close helpers in gcsfs/core.py to detect execution on the active event loop thread and offload _close_impl() to a daemon thread (gcsfs-deferred-close).
- Refactor GCSFile.close() to guard against re-entrancy via _close_deferred, bypass operations during interpreter shutdown (sys.is_finalizing()), and separate teardown logic into _close_impl().
- Update ZonalFile in gcsfs/zonal_file.py to override _close_impl() instead of close(), inheriting event loop thread detection and deferred execution.
- Add unit tests in gcsfs/tests/test_close_from_event_loop.py verifying read/write finalization on loop threads, idempotency, and synchronous off-loop error propagation.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -11,6 +11,7 @@ import os
 import posixpath
 import re
 import sys
+import threading
 import uuid
 import warnings
 import weakref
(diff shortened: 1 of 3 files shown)


yuxin00j (approved):

gcsfs/core.py
```diff
@@ -2684,8 +2725,23 @@ async def _async_fetch_range(self, start_offset, total_size, split_factor=1):
         )
 
     def close(self):
+        if self.closed or self._close_deferred:
```
Yonghui-Lee, line 2728: Instead of introducing a `ThreadPoolExecutor` and dispatching across threads, it might be better if we follow the pattern in `_finalize_mrd_pool_cache`:

* When `close()` is called **on the loop thread**: schedule the async teardown directly with `asyncio.run_coroutine_threadsafe(coro, loop)` without blocking.
* When `close()` is called **off the loop thread**: block synchronously using `asyn.sync(loop, coro)` as before.

zhixiangli: 1. _finalize_mrd_pool_cache works because MRDPoolCache.close() is a coroutine function, but GCSFile._close_impl() is not.
2. asyn.sync(loop, coro) is what this PR does already.

PR #1019, opened 2026-08-23
adds cache_type and cache_source for grpc client

ankitaluthra1:


gcsfs/extended_gcsfs.py
```diff
@@ -592,9 +602,19 @@ async def _cat_file(
                     path, start=start, end=end, concurrency=concurrency, **kwargs
                 )
 
+            cache_type = kwargs.get("cache_type", getattr(self, "cache_type", None))
+            cache_source = kwargs.get(
```
Mahalaxmibejugam: cache_source is internally used variable, so won't be available in kwargs and it would always default to None

gcsfs/extended_gcsfs.py
```diff
@@ -592,9 +602,19 @@ async def _cat_file(
...
+            )
+
             # Instantiate an MRDPool locally for this call
             mrd = await self._mrd_pool_cache.get(
-                bucket, object_name, generation, pool_size=concurrency
+                bucket,
+                object_name,
+                generation,
+                pool_size=concurrency,
+                cache_type=cache_type,
```
Mahalaxmibejugam, line 626: Should we resolve the cache_type before populating the cache and calling cache.get()? fs.open(path) and fs.open(path, cache_type="none") would both use different MRD instances for the same path, as the first case stores cache_type as None while the second one stores "none" in the cache key.

gcsfs/extended_gcsfs.py
```diff
@@ -442,10 +442,20 @@ async def _fetch_range_split(
         bucket, object_name, generation = self.split_path(path)
 
         if mrd is None:
+            cache_type = kwargs.get("cache_type", getattr(self, "cache_type", None))
```
Mahalaxmibejugam: cache_type and cache_source are not attributes of FileSystem class, hence getattr(self, "cache_type", None) would always evaluate to None.

gcsfs/zb_hns_utils.py
```diff
@@ -765,7 +798,7 @@ async def get(self, bucket_name, object_name, generation, pool_size):
         info = await fs._info(f"{bucket_name}/{object_name}", generation=generation)
         if generation is None:
             generation = info.get("generation")
-        key = (bucket_name, object_name, generation)
+        key = (bucket_name, object_name, generation, cache_type, cache_source)
```
Mahalaxmibejugam: Do we need cache_source in the cache key? The same MRD instance can be used even if the cache_type is explicitly set by the user or resolved to a default value.

Can we update code to make sure telemetry is recorded correctly(:d or :e) while making sure same mrd instance can be used for both type of requests(explicit or default for same cache_type)?

gcsfs/tests/test_extended_gcsfs.py
```diff
@@ -1479,9 +1481,16 @@ async def test_cat_file_delegates_resolved_range_to_mrd_fetch(
         )
         mock_concurrent_fetch.return_value = b"data"
 
-        await extended_gcsfs._cat_file("bucket/obj", concurrency=4)
+        await extended_gcsfs._cat_file(
+            "bucket/obj", concurrency=4, cache_type="readahead"
```
Mahalaxmibejugam, line 1485: any reason for passing hardcoded cache_type with this change?

Mahalaxmibejugam (approved):

gcsfs/zb_hns_utils.py
```diff
@@ -497,13 +523,21 @@ def __init__(
...
         self.generation = generation
         self._cache = cache
-        self._key = (bucket_name, object_name, generation)
+        self.cache_type = cache_type
+        # Note: MRDPool is shared across requests with different cache configs.
+        # self.cache_source reflects the originator of the pool. Dynamic scale-up
+        # operations (_create_mrd) will emit this initial cache_source telemetry,
+        # even if triggered by a request with a different cache_source.
+        self.cache_source = cache_source
+        self._key = (bucket_name, object_name, generation, cache_type)
```
zhixiangli, line 540: Is it possible for requests from different cache_sources to use the same MRD, and could that lead to corruption?

gcsfs/zb_hns_utils.py
```diff
@@ -42,15 +42,41 @@
...
+        metadata = [("x-goog-api-client", cache_val)]
+
+    kwargs = {}
+    if metadata:
+        kwargs["metadata"] = metadata
+
     try:
         return await AsyncMultiRangeDownloader.create_mrd(
-            grpc_client, bucket_name, object_name, generation
+            grpc_client, bucket_name, object_name, generation, **kwargs
```
zhixiangli, line 70: Could you confirm if the metadata is not lost during a retry or redirect?

gcsfs/tests/test_extended_gcsfs.py
ankitaluthra1: You mean why its hard coded in the test ? so it can be asserted...

gcsfs/extended_gcsfs.py
```diff
@@ -1693,7 +1723,16 @@ async def _get_file_request(
         generation = path_generation or kwargs.get("generation")
         callback = callback or NoOpCallback()
 
-        mrd_pool = await self._mrd_pool_cache.get(bucket, key, generation, pool_size=1)
+        cache_type, cache_source = self._resolve_cache_config(kwargs)
```
zhixiangli, line 1726: Are these disk download paths not using in-memory block caching?

PR #1020, opened 2026-08-24
feat(testing): Add dummy-io simulation support for zero-data-transfer benchmarking and testing

raj-prince:
### Summary
This PR introduces a **Dummy I/O Simulation Suite** (`DummyGcsFileSystem`, `DummyMRD`, `DummyMRDPool`, `DummyDelaySimulator`) within `gcsfs/tests/` to simulate read operations against Google Cloud Storage without performing real network data transfers.

**Motivation:** `Deterministic Offline Benchmarking` - Measure application CPU bottlenecks and I/O concurrency independently of network fluctuations.

---

### 1. Direct Instantiation in Python Scripts

Instantiate `DummyGcsFileSystem` directly and pass simulated latency/bandwidth limits:

```python
from gcsfs.tests.test_dummy_io import DummyGcsFileSystem

# Instantiate with 2ms TTFB latency and 1500 MB/s simulated bandwidth
fs = DummyGcsFileSystem(
    dummy_io_ttfb_ms=2.0,
    dummy_io_bandwidth_megabytes_per_sec=1500.0,
)

# Metadata is retrieved from real GCS, but reads return zeroed bytes with simulated timing
with fs.open("my-bucket/large_file.bin", "rb") as f:
    chunk = f.read(32 * 1024 * 1024)  # 32 MB
    print(f"Read {len(chunk)} simulated bytes")
```

---

### 2. Transparent Integration via `fsspec`

For downstream libraries (e.g., PyTorch `DataLoader`, Ray, Pandas) that use `fsspec.open("gs://...")`, register `DummyGcsFileSystem` globally without modifying application code:

```python
import fsspec
from fsspec import register_implementation
from gcsfs.tests.test_dummy_io import DummyGcsFileSystem

# Register DummyGcsFileSystem to handle 'gs://' and 'gcs://'
register_implementation("gs", DummyGcsFileSystem, clobber=True)
register_implementation("gcs", DummyGcsFileSystem, clobber=True)

# Standard fsspec calls will now use the zero-egress dummy read path
with fsspec.open("gs://my-bucket/dataset.tar", "rb") as f:
    data = f.read(64 * 1024 * 1024)
```

---

### 3. Environment Variable Configuration

Simulation parameters can be configured dynamically via environment variables:

| Environment Variable | Type | Default | Description |
|---|---|---|---|
| `GCSFS_DUMMY_IO_TTFB_MS` | `float` | `0.0` | Initial Time-to-First-Byte latency in milliseconds. |
| `GCSFS_DUMMY_IO_BANDWIDTH_MEGABYTES_PER_SEC` | `float` | `None` (unlimited) | Streaming bandwidth limit in Megabytes per second (MB/s). |

```bash
export GCSFS_DUMMY_IO_TTFB_MS=5.0
export GCSFS_DUMMY_IO_BANDWIDTH_MEGABYTES_PER_SEC=1000.0

python your_benchmark_script.py
```


gcsfs/dummy_gcsfs.py
```diff
@@ -0,0 +1,138 @@
...
+            bandwidth_mbps=dummy_io_bandwidth_mbps,
+        )
+        self._mrd_pool_cache.pool_class = DummyMRDPool
+
+    async def _cat_file_sequential(self, path, start=None, end=None, **kwargs):
+        """Simulate sequential cat file without network data transfer."""
+        if start is not None and end is not None and start >= end >= 0:
+            return b""
+
+        if start is not None and end is not None and start >= 0 and end >= start:
```
ankitaluthra1: why is this "if" needed, in actual code "_process_limits_to_offset_and_length" is always called, here also we can directly call this, end can also be more than file size so should be in-check using "_process_limits_to_offset_and_length"

gcsfs/dummy_gcsfs.py
```diff
@@ -0,0 +1,138 @@
...
+class DummyGcsFileSystem(ExtendedGcsFileSystem):
+    """
+    Subclass of ExtendedGcsFileSystem that bypasses network data transfer only on the read path.
+    """
+
+    def __init__(
+        self,
+        *args,
+        dummy_io_ttfb_ms: Optional[float] = None,
+        dummy_io_bandwidth_mbps: Optional[float] = None,
```
ankitaluthra1: is this mbps or MBps ?

gcsfs/dummy_gcsfs.py
```diff
@@ -0,0 +1,138 @@
...
+        if self.bandwidth_mbps and self.bandwidth_mbps > 0 and size > 0:
+            delay += size / (self.bandwidth_mbps * 1024 * 1024)
+        return delay
+
+    async def async_delay(self, size: int = 0, is_first_chunk: bool = False):
+        delay = self.calculate_delay(size, is_first_chunk)
+        if delay > 0:
+            await asyncio.sleep(delay)
+
+    def sync_delay(self, size: int = 0, is_first_chunk: bool = False):
```
ankitaluthra1: why is this method needed ?

ankitaluthra1 (changes requested):

gcsfs/dummy_gcsfs.py
raj-prince: not required, removing.

gcsfs/dummy_gcsfs.py
raj-prince: MBps, clarified and made the variable explicit (megabytes).

PR #1024, opened 2026-08-27
Implement native adjacent byte range coalescing in cat_ranges

yuxin00j:
## Summary
This PR implements native byte range coalescing for `GCSFileSystem.cat_ranges` and `ExtendedGcsFileSystem.cat_ranges` to optimize chunk-based workload downloads (such as PyTorch Distributed Checkpoint / DCP payload loading).

Adjacent and near-adjacent byte requests on the same object are coalesced into singular block reads natively, reducing request amplification and leveraging zero-copy `memoryview` slicing.

## Key Changes
- **Core Coalescing & Unpacking Helpers (`gcsfs/core.py`):**
  - `_coalesce_ranges`: Groups and merges overlapping, contiguous (`max_gap=0`), and near-contiguous (`gap <= max_gap`) range requests for a given file, generating relative slice offsets mapped to caller indices.
  - `_merge_file_ranges` & `_is_coalesce_enabled`: Unifies conditional coalescing vs 1-to-1 range mapping.
  - `_unpack_range_results`: Safely populates caller results with `bytes` when coalescing is inactive (`max_gap is None` or `< 0`) or `memoryview` slices when active (`max_gap >= 0`).
  - Safely handles bounded and unbounded ranges (`end=None`), empty inputs, scalar broadcasting, and separate non-overlapping clusters.

- **Auto / Dynamic max_gap Calculation (`gcsfs/core.py`):**
  - `_compute_adaptive_max_gap`: Dynamically derives an optimal coalescing gap per file from that file's requested chunk lengths (5% of median chunk size, capped at 1 MB). Prevents read amplification on small ranges while optimizing throughput for large megabyte-scale tensors, and avoids cross-file median skew when mixing small metadata files and large tensor shards.
  - Enables auto coalescing via `auto_max_gap=True` or `max_gap="auto"`.

- **`GCSFileSystem._cat_ranges` (`gcsfs/core.py`):**
  - Adds native support for coalesced byte range reads across multiple files.
  - Executes chunked reads using `asyn._run_coros_in_chunks` respecting `batch_size`.
  - Supports explicit integer `max_gap`, `auto_max_gap=True`, and `max_gap="auto"` (computed per file).
  - Supports both `on_error="return"` and `on_error="raise"` error modes.

- **`ExtendedGcsFileSystem._cat_ranges` (`gcsfs/extended_gcsfs.py`):**
  - **Bulk Zonal Downloads via MRD**: Connects range coalescing with Zonal (`RAPID`) buckets via `AsyncMultiRangeDownloader` (`MRD`).
  - **Dynamic Multi-MRD Pool Concurrency**: Sizes `pool_size = min(max_batches, max(1, concurrency))` to stream multiple batches in parallel across pooled MRDs.
  - **Per-File Adaptive max_gap for Zonal Downloads**: Computes effective coalescing gap per file and applies it to MRD range batches.
  - **Unified Coroutine Scheduling**: Combines Zonal MRD batch downloads and Non-Zonal HTTP GETs into a single coroutine pool scheduled via `asyn._run_coros_in_chunks`.
  - **Lazy Metadata Lookup**: Queries object metadata (`_info()`) only when unbounded ranges require file size resolution.
  - **Resource Cleanup**: Ensures all acquired `MRDPool` instances are cleanly closed in `finally` blocks upon completion or failure.

## Testing
- **Unit Tests in `gcsfs/tests/test_core.py`:**
  - Tests for `_compute_adaptive_max_gap` covering empty inputs, small chunks, 10MB chunks, 100MB chunks (capped at 1MB), and custom ratios.
  - Tests for `cat_ranges` with `auto_max_gap=True`, `max_gap="auto"`, per-file adaptive gap isolation across mixed file sizes, scalar broadcasting, contiguous coalescing (`max_gap=0`), multi-file coalescing, and error modes.
- **Unit Tests in `gcsfs/tests/test_extended_gcsfs.py`:**
  - Tests for non-zonal delegation (with `auto_max_gap=True` and `max_gap="auto"`).
  - Tests for zonal coalescing with `MockMRD` (including auto coalesced ranges, wide uncoalesced ranges, and per-file adaptive gap isolation).
  - Tests for dynamic MRD pool concurrency, chunk-based batch size limits via `_run_coros_in_chunks`, return types (`bytes` vs `memoryview`), and pool cleanup upon error.


gcsfs/extended_gcsfs.py
```diff
@@ -627,6 +634,186 @@ async def _cat_file(
...
+                    non_zonal_starts.append(s)
+                    non_zonal_ends.append(e)
+                    non_zonal_indices.append(idx)
+            else:
+                zonal_files.append((path, items, bucket, object_name, generation))
+
+        async def _fetch_zonal_file(path, items, bucket, object_name, generation):
+            # 2. Zonal Bucket: Acquire MRD from pool cache
+            try:
+                mrd_pool = await self._mrd_pool_cache.get(
```
zhixiangli: Every time get the pool, it leaks.

zhixiangli: To prevent leaks, should close the pool.

gcsfs/extended_gcsfs.py
```diff
@@ -627,6 +634,186 @@ async def _cat_file(
...
+                merged_ranges = [
+                    (
+                        s,
+                        e,
+                        [
+                            (
+                                idx,
+                                0,
+                                (
+                                    (e - s)
```
zhixiangli: How about normalizing them first?
- Normalize using _process_limits_to_offset_and_length.
- If the length is 0, assign "" directly.

gcsfs/core.py
```diff
@@ -1288,6 +1368,110 @@ async def _cat_file(
...
+            file_groups[p].append((s_val, e, orig_idx))
+
+        merged_paths = []
+        merged_starts = []
+        merged_ends = []
+        merged_slice_maps = []  # list of [(orig_idx, rel_start, rel_end), ...]
+
+        # 2. Coalesce adjacent/near-adjacent ranges per file
+        for p, items in file_groups.items():
+            for m_s, m_e, slice_list in _coalesce_ranges(items, max_gap):
```
zhixiangli: For the sake of completeness, is it possible for s < 0 or e < 0? Additionally, we should consider _process_limits_to_offset_and_length for normalization.

gcsfs/extended_gcsfs.py
```diff
@@ -627,6 +634,186 @@ async def _cat_file(
...
+                                        else None
+                                    )
+                                ),
+                            )
+                        ],
+                    )
+                    for s, e, idx in items
+                ]
+
+            effective_batch_size = batch_size or 64
```
zhixiangli: Can batch_size be -1 for no chunking?

zhixiangli: Do we need a cap of `effective_batch_size` such as 1000?

gcsfs/extended_gcsfs.py
```diff
@@ -627,6 +634,186 @@ async def _cat_file(
...
+            if not bucket_zonal_map[bucket]:
+                for s, e, idx in items:
+                    non_zonal_paths.append(path)
+                    non_zonal_starts.append(s)
+                    non_zonal_ends.append(e)
+                    non_zonal_indices.append(idx)
+            else:
+                zonal_files.append((path, items, bucket, object_name, generation))
+
+        async def _fetch_zonal_file(path, items, bucket, object_name, generation):
```
zhixiangli: Shall we bound total active download jobs across all files using an asyncio.Semaphore(max_concurrency)

zhixiangli (changes requested):

gcsfs/extended_gcsfs.py
yuxin00j: Done.

gcsfs/core.py
yuxin00j: Done. Added a _normalize_file_ranges function.

gcsfs/extended_gcsfs.py
yuxin00j: Done.

yuxin00j: Done.

gcsfs/extended_gcsfs.py
yuxin00j: Done. Added asyncio.Semaphore(file_concurrency) across all zonal files.

gcsfs/extended_gcsfs.py
yuxin00j: Done. Normalized ranges via _normalize_file_ranges and any 0-length slices are assigned b"" directly.

gcsfs/core.py
```diff
@@ -1288,6 +1370,251 @@ async def _cat_file(
...
+            results (list): Output results list to populate for 0-length or error slices.
+            on_error (str): "return" or "raise".
+            get_size_fn (callable, optional): Sync or async function returning file size
+                if known without calling self._info(path).
+
+        Returns:
+            list of (offset, end_offset, orig_idx): Normalized non-zero ranges to fetch,
+            or None if an error occurred and was captured in results.
+        """
+        needs_file_size = any(
```
zhixiangli, line 1473: 1. This will cause an issue when start=90 and end=110 on a 100-byte object; the length should be 10, not 20.

2, Could you also consolidate the duplicate logic?

yuxin00j: 1. In this case, backend will return byte 90-100. We want to save unnecessary _info call when start and end are positive.
2. Yes

gcsfs/core.py
```diff
@@ -1286,6 +1406,235 @@ async def _cat_file(self, path, start=None, end=None, **kwargs):
...
+            isinstance(max_gap, str) and max_gap.lower() == "auto"
+        )
+        if auto_enabled:
+            all_lengths = [
+                e - s
+                for items in valid_items_per_file.values()
+                for s, e, _ in items
+                if s is not None and e is not None and e > s
+            ]
+            effective_max_gap = _compute_adaptive_max_gap(all_lengths)
```
ankitaluthra1, line 1590: Keeping effective_max_gap can lead to regression in certain scenarios:

Example Scenario:
File A requests small 100-byte metadata slices.
File B requests 50 MB checkpoint shards.
If File B dominates, effective_max_gap scales up to 1 MB. This 1 MB gap is then applied to File A, causing up to 10,000× read amplification on small files.
If File A dominates, effective_max_gap collapses to 5 bytes, preventing legitimate coalescing on File B.

Consider making it per file

ankitaluthra1 (changes requested):

PR #1032, opened 2026-09-01
microbenchmarks: separate regional and zonal large file comparison configurations

Yonghui-Lee:
Separates the `download_large_file` and `upload_large_file` scenario configurations into distinct blocks for `regional` and `zonal` bucket types.
  - Regional large download/upload: 4 cases each (1 GB, 10 GB, 50 GB, 100 GB)
  - Zonal large download/upload: 5 cases each (1 GB, 10 GB, 50 GB, 100 GB, 200 GB)

### DIFF
--- a/gcsfs/tests/perf/microbenchmarks/comparison/configs.yaml
+++ b/gcsfs/tests/perf/microbenchmarks/comparison/configs.yaml
@@ -2,12 +2,6 @@ common:
   bucket_types:
     - "regional"
     - "zonal"
-  file_sizes_mb:
-    - 1024 # 1 GB
-    - 10240 # 10 GB
-    - 51200 # 50 GB
-    - 102400 # 100 GB
-    - 204800 # 200 GB
   chunk_sizes_mb:
     - 32
   threads:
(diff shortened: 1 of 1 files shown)


yuxin00j (approved):

gcsfs/tests/perf/microbenchmarks/comparison/configs.yaml
```diff
@@ -2,29 +2,56 @@ common:
...
   - name: "download_large_file"
+    bucket_types: ["regional"]
+    file_sizes_mb:
+      - 1024 # 1 GB
+      - 10240 # 10 GB
+      - 51200 # 50 GB
+      - 102400 # 100 GB
+
+  # 1b. Download Single Large File - Zonal
+  - name: "download_large_file"
```
zhixiangli, line 22: nit: is there a better name than "large/small file," as it is ambiguous what constitutes a large or small file?

zhixiangli (approved):

gcsfs/tests/perf/microbenchmarks/comparison/configs.yaml
Yonghui-Lee: Good point regarding ambiguity. The terms large_file and small_files are intended as high-level workload categories. Keeping the current naming can avoid breaking existing Cloud Build triggers and BigQuery.

PR #1040, opened 2026-09-04
fix: harden prefetcher/zonal teardown

### LINKED ISSUE #1037: Adaptive prefetcher's `close()` can hang forever when called from `__del__` during interpreter shutdown
### Environment

- gcsfs 2026.8.0
- fsspec 2026.7.0
- pyarrow reading Parquet from GCS
- Python 3.12.3, Linux

### Summary

`BackgroundPrefetcher.close()` calls `fsspec.asyn.sync()` with no timeout. It is reachable from
`AbstractBufferedFile.__del__`, which the interpreter may run at any time on any thread, including
during finalisation. When it runs after the fsspec IO loop thread has been torn down, `sync()` blocks
forever and the process never exits.

Since adaptive prefetching became the default read path, any consumer that lets the garbage collector
close its `GCSFile` handles is exposed. pyarrow is one such consumer: it opens the file handles
itself and does not close them explicitly.

### The chain

```
__del__ (fsspec/spec.py:2316)       if not self.closed: self.close()
close   (gcsfs/core.py:2689)        if ... self._prefetch_engine: self._prefetch_engine.close()
close   (gcsfs/prefetcher.py:875)   fsspec.asyn.sync(self.loop, self._async_close)   # no timeout
sync    (fsspec/asyn.py:91)         while True: if event.wait(1): break
```

`sync()` guards only on `loop.is_closed()`. A loop whose *thread* has died is not "closed", so the
guard passes, `run_coroutine_threadsafe()` happily queues the coroutine on a loop nobody is running,
and the event is never set. With `timeout=None` the `while True` has no exit path, so it polls once a
second indefinitely.

The IO loop thread is created as a daemon (`fsspec/asyn.py`, `th.daemon = True`), so the interpreter
tears it down early in finalisation. A destructor running after that point can only hang.

### Three outcomes from the same destructor

Which one you get depends on where the handle is reclaimed:

| Where `__del__` runs | Result |
|---|---|
| the fsspec IO loop thread | `NotImplementedError: Calling sync() from within a running loop`, swallowed as "Exception ignored in `__del__`"; the prefetcher never shuts down |
| another thread, loop alive | works correctly |
| another thread, during interpreter shutdown | **hangs forever** |

The first case is silent but constant. In our workload it fired on 47%–97% of file handles wherever
more than one file was read per dataset — roughly one swallowed exception per handle.

### Reproduction

```python
import gcsfs
import pyarrow.parquet as pq

fs = gcsfs.GCSFileSystem()
dataset = pq.ParquetDataset(paths, filesystem=fs)   # paths = ~20 small parquet objects
df = dataset.read(use_threads=True).to_pandas()
del dataset
# process may now hang at interpreter exit, or print "Exception ignored in __del__"
```

Needs more than one file to trigger with any reliability. With a single file the handle tends to be
released on the calling thread while the loop is still alive, and everything closes cleanly. We saw
this very sharply across sites reading different numbers of objects per fetch:

| objects per read | swallowed exceptions per object | ever hung |
|---|---|---|
| 1 | 0.00 | no, across 60+ runs |
| 14–19 | 0.47 | no |
| 17–22 | 0.70 | yes |
| 19–28 | 0.90 | no |
| 40–50 | 0.97 | yes |

A hung process shows as `S (sleeping)` in `futex_wait_queue_me` with almost no CPU accumulated —
ours used 9 seconds of CPU across 35 minutes of hanging — which is consistent with the one-second
poll loop rather than a busy spin.

### Suggested fix

`Prefetcher.close()` cannot safely assume it is on a normal thread with a live loop, because it is
reachable from a destructor. Two changes would each remove the failure:

1. **Pass a timeout** to `sync()` so the worst case is a bounded delay and an `FSTimeoutError`
   rather than a permanent hang.
2. **Detect the unusable cases and degrade** instead of blocking — if `asyncio.get_running_loop()`
   is the prefetcher's own loop, or the loop's thread is no longer alive, schedule the close
   fire-and-forget (or skip it) rather than waiting on it.

It may also be worth having `sync()` itself check that the loop is actually being serviced, not just
that it isn't closed. Right now "loop object exists but nothing is running it" is indistinguishable
from a healthy loop, and the failure it produces is an unkillable wait rather than an error.

### Workaround

Disabling the prefetcher avoids it entirely, since `_prefetch_engine` is then `None` and `close()`
never reaches `sync()`. Any of the three documented mechanisms works — an explicit `cache_type`, the
`USE_EXPERIMENTAL_ADAPTIVE_PREFETCHING` environment variable, or
`use_experimental_adaptive_prefetching=False` passed to `open()`.

For pyarrow specifically, none of those can be passed through, because pyarrow calls
`fs.open(path, "rb")` with no extra options. We subclass to inject it:

```python
class _PrefetchAwareGCSFileSystem(gcsfs.GCSFileSystem):
    def __init__(self, *args, use_prefetching=None, **kwargs):
        self._use_prefetching = use_prefetching
        super().__init__(*args, **kwargs)

    def _open(self, path, mode="rb", **kwargs):
        if self._use_prefetching is not None:
            kwargs.setdefault("use_experimental_adaptive_prefetching", self._use_prefetching)
        return super()._open(path, mode, **kwargs)
```

That works, but it does mean the documented per-open options are unreachable for the most common
Parquet-on-GCS path. A filesystem-level default would be a friendlier escape hatch.

Yonghui-Lee:
Fixes #1037

### Overview

When closing `BackgroundPrefetcher` or `ZonalFile` from synchronous finalizers (`__del__` or `weakref.finalize`), teardown could hang forever. This occurred in multi-threaded data pipelines (e.g. PyArrow Parquet datasets, Ray, Dask) where file handles are collected during interpreter shutdown or directly on the event loop thread.

This PR hardens the teardown paths across `BackgroundPrefetcher` and `ZonalFile`, eliminating reentrant deadlocks, preventing garbage collection of in-flight cleanup tasks, bounding synchronous teardowns, and isolating read/write failures.

### Root Causes & Fixes

1. **Reentrant Deadlock on Event Loop Thread**
   - **Problem:** If garbage collection drops the last reference to a file while executing on the `asyncio` event loop thread, calling `fsspec.asyn.sync()` or `future.result()` blocks the loop thread waiting on itself, causing an immediate deadlock or a swallowed `NotImplementedError`.
   - **Fix:** Uses `_on_loop_thread(loop)` to detect reentrant calls. Instead of blocking, the teardown coroutine is scheduled non-blockingly on the loop via `_defer_task()`.

2. **Garbage-Collected Background Tasks ("Task was destroyed but it is pending")**
   - **Problem:** Tasks created via `loop.create_task()` without an external strong reference can be garbage-collected mid-flight by Python's cyclic GC, aborting teardowns prematurely.
   - **Fix:** Introduced `_defer_task()` and `_deferred_close_tasks` in `gcsfs.zonal_file`. Strong references are maintained under a thread lock until the task completes, and unhandled exceptions are logged via a `done_callback`.

3. **In-Flight Reads & Bounded Timeouts in `BackgroundPrefetcher.close()`**
   - **Problem:** `PrefetchProducer.stop` intentionally awaits in-flight network chunk reads rather than cancelling them to prevent disrupting the shared `MultiRangeDownloader` connection pool. Furthermore, `run_coroutine_threadsafe` could leave unawaited coroutine warnings if the loop was concurrently closed.
   - **Fix:**
     - If `timeout` expires, teardown is left running in the background on the IO loop to allow buffers to drain safely without raising errors or truncating reads.
     - Explicitly closes unawaited coroutines if `run_coroutine_threadsafe` raises `RuntimeError`.

4. **Independent Teardown & Bounded Waits in `ZonalFile`**
   - **Problem:** In `ZonalFile._close_impl()`, if read pool (`mrd_pool.close()`) teardown raised an exception, write-side appendable object finalization (`close_aaow`) was skipped, causing potential data loss. Additionally, unbounded sync calls could stall the single serial `gcsfs-deferred-close` worker thread indefinitely.
   - **Fix:**
     - Read pool and write stream teardowns are isolated into independent `try...except` blocks in `_close_impl()`.
     - Bounded synchronous wait using `_DEFAULT_TEARDOWN_TIMEOUT_SECONDS = 60.0` (or file-level `timeout`).
     - Added guards for `sys.is_finalizing()`.

### DIFF
--- a/gcsfs/prefetcher.py
+++ b/gcsfs/prefetcher.py
@@ -1,18 +1,22 @@
 import asyncio
+import concurrent.futures
 import ctypes
 import logging
+import sys
 import weakref
 from collections import deque
 
 import fsspec.asyn
 
-logger = logging.getLogger(__name__)
-
+from gcsfs.core import _on_loop_thread
 from gcsfs.zb_hns_utils import (
     HAS_CPYTHON_API,
     PyBytes_AsString,
     PyBytes_FromStringAndSize,
 )
+from gcsfs.zonal_file import _defer_task
+
+logger = logging.getLogger(__name__)
 
 
 # Please refer to following discussion to understand why this is required at this point
@@ -821,15 +825,16 @@ class BackgroundPrefetcher:
                 raise
 
     async def _async_close(self):
-        """Asynchronous teardown logic protected by the async lock."""
+        """Asynchronous teardown logic protected by the async lock.
+
+        Held for the duration of teardown so a concurrent in-flight
+        ``_async_fetch`` cannot observe a stopped producer/empty queue as
+        EOF and silently return a truncated read.
+        """
         async with self._async_lock:
-            if self.is_stopped:
-                return
+            logger.debug("Signaling prefetcher stop and tearing down producer.")
 
-            self.is_stopped = True
-            logger.debug("Acquired async lock. Tearing down producer and buffers.")
-
-            if self.producer:
+            if self.producer and not self.producer.is_stopped:
                 await self.producer.stop()
 
             self.consumer.clear_buffer()
@@ -868,8 +873,70 @@ class BackgroundPrefetcher:
 
     async def aclose(self):
         """Safely shuts down the prefetcher from an asynchronous context."""
+        if self.is_stopped:
+            return
+        self.is_stopped = True
         await self._async_close()
 
-    def close(self):
-        """Safely shuts down the prefetcher from a synchronous context."""
-        fsspec.asyn.sync(self.loop, self._async_close)
+    def close(self, timeout: float | None = 10.0):
+        """Safely shuts down the prefetcher from a synchronous context.
+
+        Schedules teardown on the IO loop. If the wait times out, teardown
+        continues in the background to let in-flight reads finish cleanly.
+
+        Args:
+            timeout: Maximum seconds to wait for teardown to finish. Defaults
+                to 10.0 seconds. If 0 or negative, teardown runs fire-and-forget.
+                If None, waits indefinitely.
+        """
+        if self.is_stopped:
+            return
+        self.is_stopped = True
+
+        if sys.is_finalizing():
+            return
+
+        loop = self.loop
+        if loop is None or loop.is_closed() or not loop.is_running():
+            # If the event loop is absent, closed, or not currently serviced
+            # by a running thread, async scheduling is impossible.
+            return
+
+        if _on_loop_thread(loop):
+            # Avoid deadlock when closing reentrantly from the event loop thread.
+            _defer_task(
+                loop,
+                self._async_close(),
+                description="BackgroundPrefetcher teardown",
+                logger=logger,
+                log_level=logging.WARNING,
+            )
+            return
+
+        coro = self._async_close()
+        try:
+            future = asyncio.run_coroutine_threadsafe(coro, loop)
+        except RuntimeError:
+            # Loop was closed concurrently between the checks above and here.
+            # Explicitly close the unawaited coroutine to avoid RuntimeWarning.
+            coro.close()
+            return
+
+        if timeout is not None and timeout <= 0:
+            # Fire-and-forget requested: teardown is scheduled, return without waiting.
+            return
+
+        try:
+            future.result(timeout)
+        except concurrent.futures.TimeoutError:
+            # Teardown timed out on the caller thread. Leave coroutine running in
+            # background on the IO loop so pending network buffers can drain safely.
+            logger.warning(
+                "BackgroundPrefetcher teardown did not complete within %ss; "
+                "it will keep running in the background.",
+                timeout,
+            )
+        except Exception:
+            logger.warning(
+                "Exception while closing BackgroundPrefetcher", exc_info=True
+            )
--- a/gcsfs/prefetcher.py
+++ b/gcsfs/prefetcher.py
@@ -873,9 +873,9 @@ class BackgroundPrefetcher:
 
     async def aclose(self):
         """Safely shuts down the prefetcher from an asynchronous context."""
-        if self.is_stopped:
-            return
         self.is_stopped = True
+        if self.producer is None or self.producer.is_stopped:
+            return
         await self._async_close()
 
     def close(self, timeout: float | None = 10.0):
@@ -889,9 +889,9 @@ class BackgroundPrefetcher:
                 to 10.0 seconds. If 0 or negative, teardown runs fire-and-forget.
                 If None, waits indefinitely.
         """
-        if self.is_stopped:
-            return
         self.is_stopped = True
+        if self.producer is None or self.producer.is_stopped:
+            return
 
         if sys.is_finalizing():
             return
--- a/gcsfs/prefetcher.py
+++ b/gcsfs/prefetcher.py
@@ -874,8 +874,6 @@ class BackgroundPrefetcher:
     async def aclose(self):
         """Safely shuts down the prefetcher from an asynchronous context."""
         self.is_stopped = True
-        if self.producer is None or self.producer.is_stopped:
-            return
         await self._async_close()
 
     def close(self, timeout: float | None = 10.0):
@@ -890,8 +888,6 @@ class BackgroundPrefetcher:
                 If None, waits indefinitely.
         """
         self.is_stopped = True
-        if self.producer is None or self.producer.is_stopped:
-            return
 
         if sys.is_finalizing():
             return
--- a/gcsfs/prefetcher.py
+++ b/gcsfs/prefetcher.py
@@ -1,20 +1,17 @@
 import asyncio
-import concurrent.futures
 import ctypes
 import logging
-import sys
 import weakref
 from collections import deque
 
 import fsspec.asyn
 
-from gcsfs.core import _on_loop_thread
 from gcsfs.zb_hns_utils import (
     HAS_CPYTHON_API,
     PyBytes_AsString,
     PyBytes_FromStringAndSize,
+    sync_teardown,
 )
-from gcsfs.zonal_file import _defer_task
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


gcsfs/zonal_file.py
```diff
@@ -19,6 +22,53 @@
...
+            exc = t.exception()
+            if exc:
+                log = logger or logging.getLogger("gcsfs")
+                log.log(
+                    log_level,
+                    "%s failed during asynchronous execution: %s",
+                    description,
+                    exc,
+                    exc_info=exc,
+                )
```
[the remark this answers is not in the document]
Yonghui-Lee: Done

gcsfs/prefetcher.py
```diff
@@ -1,18 +1,22 @@
...
-
+from gcsfs.core import _on_loop_thread
 from gcsfs.zb_hns_utils import (
     HAS_CPYTHON_API,
     PyBytes_AsString,
     PyBytes_FromStringAndSize,
 )
+from gcsfs.zonal_file import _defer_task
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
Yonghui-Lee: The task garbage collection issue applies to `loop.create_task()`, not `asyncio.run_coroutine_threadsafe()`. When calling `run_coroutine_threadsafe()`, the event loop itself holds strong references to the callback and the scheduled task in its internal queues until completion. Discarding the returned `concurrent.futures.Future` does not cause the underlying task to be garbage-collected or aborted mid-flight.

yuxin00j (approved):

gcsfs/prefetcher.py
```diff
@@ -868,8 +873,70 @@ def fetch(self, start: int | None, end: int | None) -> bytes:
 
     async def aclose(self):
         """Safely shuts down the prefetcher from an asynchronous context."""
+        if self.is_stopped:
+            return
+        self.is_stopped = True
```
zhixiangli, line 876: Setting self.is_stopped = True before _async_close() finishes means if the caller is cancelled while producer.stop() runs, is_stopped stays True while producer tasks remain alive. Subsequent  aclose() / close() calls exit immediately due to the early if self.is_stopped: return check, permanently leaking network tasks and buffers.

gcsfs/zonal_file.py
```diff
@@ -397,13 +447,81 @@ def _upload_chunk(self, final=False):
...
+        Unlike purely read-side caching, write-side teardown finalizes appendable
+        objects server-side. Branches that cannot safely run the teardown coroutine
+        log warnings or errors to provide visibility into potential unfinalized resources.
+        """
+        path = getattr(self, "path", "<unknown>")
+
+        if sys.is_finalizing():
+            return
+
+        if loop is None or not loop.is_running():
```
zhixiangli, line 491: If the event loop is missing or not running, _sync_teardown() logs and returns None instead of raising. zonal_file.py:447 sees no error, marking the file closed and returning success even though close_aaow was skipped with finalize_on_close=True. This regresses previous asyn.sync() behavior and risks silent upload finalization failure.

gcsfs/prefetcher.py
Yonghui-Lee: Early return is removed.

gcsfs/zonal_file.py
Yonghui-Lee: An error is added.

gcsfs/zonal_file.py
```diff
@@ -397,13 +447,80 @@ def _upload_chunk(self, final=False):
...
+                    finalize_on_close=self.finalize_on_close,
+                    description="finalizing AsyncAppendableObjectWriter",
+                )
+            except Exception as e:
+                errors.append(e)
+
+        if errors:
+            raise errors[0]
+
+    def _sync_teardown(self, loop, func, *args, description, **kwargs):
```
raj-prince, line 479: Currently, prefetcher.py and zonal_file.py implement separate, slightly divergent teardown methods (asyncio.run_coroutine_threadsafe vs fsspec.asyn.sync), both re-checking sys.is_finalizing(), loop.is_running(), and _on_loop_thread(). Defining _defer_task in zonal_file.py also introduces circular coupling between prefetcher and zonal_file.

Can we consolidate into a single helper in gcsfs/core.py?

gcsfs/tests/test_prefetcher.py
```diff
@@ -724,3 +729,304 @@ def test_fast_slice_pypy_fallback():
 
     # 3. Verify exact bounds
     assert _fast_slice(src, 0, len(src)) == src
+
+
+def test_close_when_sys_finalizing(prefetcher_factory, monkeypatch):
```
raj-prince, line 734: Overall, I feel we can make the test simple and readable  by:
1. minimizing the mock instance, if possible please use the real light weight object if possible.
2. following AAA (Arrange-Act-Assert) - a clear segregation b/w these three if possible.
3. focusing on contract instead of internal implementation, for example, why are we asserting on mock scheduled, we should only care about close() returns immediately without hanging or raising any error.

E.g.,
```python
def test_close_during_interpreter_finalization(monkeypatch):
    # Arrange
    bp = BackgroundPrefetcher(fetcher=MockFetcher(b"X"), size=100, loop=fsspec.asyn.get_loop())
    monkeypatch.setattr(sys, "is_finalizing", lambda: True)

    # Act
    bp.close()

    # Assert
    assert bp.is_stopped is True
```

gcsfs/zonal_file.py
ankitaluthra1: Since these are common utilss and not GCSFile or GCSFilesystem related core features, gcsfs/core.py is not the right place, either use zb_hns_utils or create another prefetcher_utils to define common utility

gcsfs/tests/test_prefetcher.py
ankitaluthra1: I don't see much diff here apart from new lines, m I missing something ?

gcsfs/zonal_file.py
Yonghui-Lee: I move the helper to zb_hns_utils

gcsfs/tests/test_prefetcher.py
Yonghui-Lee: Thanks for the feedback. I refactor the tests.

ankitaluthra1 (approved):

PR #1044, opened 2026-09-07
perf(micro): optimize and fix bug in processing objects path

raj-prince:
### Summary
This PR optimizes the object metadata formatting path in `GCSFileSystem._process_object` by replacing `posixpath.join(bucket, object_metadata["name"])` with fast f-string interpolation `f"{bucket}/{object_metadata['name']}"`.

**Motivation:**
`_process_object` is on the critical hot path for all discovery, listing, and traversal operations (`ls(detail=True)`, `find()`, `glob()`, and `walk()`). For every object returned by the GCS JSON API (up to thousands or millions in large datasets), `_process_object` is invoked in a tight loop.

* **Overhead Removed:** `posixpath.join` introduces extra Python function calls, type checks, and normalization logic per object. Replacing it with direct C-level string interpolation speeds up object dictionary formatting by **~18%**.
* **Bug Fix for Leading Slash Object Names:** When an object key literally starts with a `/` (e.g. `"/file.txt"`), `posixpath.join("my-bucket", "/file.txt")` treated the key as an absolute root path and silently dropped the bucket name (producing `"/file.txt"`), which caused downstream `fs.open()` / `fs.cat()` calls to fail with `FileNotFoundError` or permission errors. Using `f"{bucket}/{object_metadata['name']}"` ensures the bucket prefix is always preserved (producing `"my-bucket//file.txt"`), which correctly round-trips through `_split_path`.

---

### Benchmark Results

Testing batch processing of 20,000 object metadata items (via `timeit`):

| Metric | Baseline (`posixpath.join`) | Optimized (f-string) | Improvement |
| :--- | :--- | :--- | :--- |
| **Batch Latency (20k items)** | 22.59 ms | 18.58 ms | **~18% faster** |
| **Per-Object Latency** | 1.13 µs | 0.93 µs | **-0.20 µs / object** |
| **Throughput** | ~885,000 ops/sec | ~1,076,000 ops/sec | **+21.5% ops/sec** |

---

### Test

Added unit and regression tests in `gcsfs/tests/test_process_object.py`:
- `test_process_object_structure`: Validates that all dictionary fields (`name`, `size`, `type`, `ctime`, `mtime`, `generation`, `metageneration`) are preserved with identical types and values.
- `test_process_object_leading_slash`: Validates that objects with leading slashes preserve the bucket prefix and correctly roundtrip through `_split_path`.

### DIFF
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -8,7 +8,6 @@ import json
 import logging
 import mimetypes
 import os
-import posixpath
 import queue
 import re
 import sys
@@ -574,7 +573,7 @@ class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
         """
         result = dict(object_metadata)
         result["size"] = int(object_metadata.get("size", 0))
-        result["name"] = posixpath.join(bucket, object_metadata["name"])
+        result["name"] = f"{bucket}/{object_metadata['name']}"
         result["type"] = "file"
         # Translate time metadata from GCS names to fsspec standard names.
         # TODO(issues/559): Remove legacy names `updated` and `timeCreated`?
--- a/gcsfs/tests/test_core.py
+++ b/gcsfs/tests/test_core.py
@@ -3712,8 +3712,7 @@ def test_user_agent_includes_cache_type_and_source_in_read(gcs):
         assert any("cache_type/readahead:d" in ua for ua in user_agents)
 
 
-def test_process_object_structure():
-    fs = GCSFileSystem(token="anon")
+def test_process_object_structure(gcs):
     bucket = "my-bucket"
     metadata = {
         "kind": "storage#object",
@@ -3730,7 +3729,7 @@ def test_process_object_structure():
         "md5Hash": "dummyHash==",
     }
 
-    processed = fs._process_object(bucket, metadata)
+    processed = gcs._process_object(bucket, metadata)
 
     assert processed["name"] == "my-bucket/nested/file.txt"
     assert processed["size"] == 4096
@@ -3745,16 +3744,15 @@ def test_process_object_structure():
     assert processed["metageneration"] == "2"
 
 
-def test_process_object_leading_slash():
-    fs = GCSFileSystem(token="anon")
+def test_process_object_leading_slash(gcs):
     bucket = "my-bucket"
     metadata = {
         "name": "/leading_slash_file.txt",
         "size": "100",
     }
 
-    processed = fs._process_object(bucket, metadata)
-    parsed_bucket, parsed_key, _ = fs.split_path(processed["name"])
+    processed = gcs._process_object(bucket, metadata)
+    parsed_bucket, parsed_key, _ = gcs.split_path(processed["name"])
 
     assert processed["name"] == "my-bucket//leading_slash_file.txt"
     assert parsed_bucket == "my-bucket"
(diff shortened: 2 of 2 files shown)


Yonghui-Lee (approved):

gcsfs/tests/test_core.py
```diff
@@ -3710,3 +3710,52 @@ def test_user_agent_includes_cache_type_and_source_in_read(gcs):
...
+    )
+    assert processed["mtime"] == datetime(
+        2024, 1, 15, 11, 45, 0, 123456, tzinfo=timezone.utc
+    )
+    assert processed["generation"] == "12345"
+    assert processed["metageneration"] == "2"
+
+
+def test_process_object_leading_slash():
+    fs = GCSFileSystem(token="anon")
```
ankitaluthra1: Consider using  the setup by using the existing gcs fixture instead of hardcoding instances.

gcsfs/core.py
```diff
@@ -574,7 +573,7 @@ def _process_object(self, bucket, object_metadata):
         """
         result = dict(object_metadata)
         result["size"] = int(object_metadata.get("size", 0))
-        result["name"] = posixpath.join(bucket, object_metadata["name"])
+        result["name"] = f"{bucket}/{object_metadata['name']}"
```
ankitaluthra1, line 576: posixpath.join natively normalizes duplicate slashes. Ensure this change is backward compatible

ankitaluthra1 (changes requested):

gcsfs/core.py
raj-prince: I feel, the change is fully backward compatible for the following reasons:

**1. Bucket names never have trailing slashes**
* In Google Cloud Storage, bucket names are strictly alphanumeric (plus `-`, `_`, `.`) and cannot contain `/`.
* Internally across all callers in `gcsfs` (`split_path`, `_do_list_objects`, `_get_object`, `_dircache`), `bucket` is extracted via `path.split("/", 1)[0]`, which never ends with a trailing `/`.

**2. `posixpath.join` does not normalize duplicate slashes**
In Python's standard library ([`posixpath.py:72-92`](https://github.com/python/cpython/blob/main/Lib/posixpath.py#L72-L92)), `join` is implemented as:

```python
def join(a, *p):
    a = os.fspath(a)
    sep = _get_sep(a)  # '/'
    path = a
    for b in p:
        b = os.fspath(b)
        if b.startswith(sep) or not path:
            path = b
        elif path.endswith(sep):
            path += b
        else:
            path += sep + b
    return path
```

* For a nested key with consecutive slashes like `b = "dir//file.txt"`, `posixpath.join("my-bucket", "dir//file.txt")` executes the `else` branch: `"my-bucket" + "/" + "dir//file.txt"` $\rightarrow$ `"my-bucket/dir//file.txt"`. It **never** collapsed `//` to `/` (collapsing duplicate slashes is done by `posixpath.normpath()`, not `join()`).
* In fact, normalizing or collapsing slashes in object keys would be undesirable because GCS treats `dir/file.txt` and `dir//file.txt` as distinct object keys. We should preserve the exact key returned by the GCS API.

**3. Edge-case bug fix**
The only difference with `posixpath.join` is the `if b.startswith(sep)` branch: when an object key in GCS literally starts with a `/` (e.g. `"/file.txt"`), `posixpath.join("my-bucket", "/file.txt")` discarded the bucket name entirely and returned `"/file.txt"`, which broke subsequent `fs.open()` calls. Using `f"{bucket}/{object_metadata['name']}"` preserves the bucket qualification (`"my-bucket//file.txt"`).

Therefore, `f"{bucket}/{object_metadata['name']}"` is 100% backward compatible with `posixpath.join` across all standard and nested paths while avoiding the overhead of `posixpath.join`.

Please let me know, if you think any such case.

PR #1054, opened 2026-09-09
Feat: Automatic upstream framework telemetry detection

Mahalaxmibejugam:
### Summary

Adds automatic upstream framework detection and telemetry tagging in `gcsfs` outgoing `User-Agent` HTTP request headers (e.g., `fw/pandas`, `fw/torch`, `fw/lightning`, `fw/ray`, `fw/dask`, `fw/polars`, `fw/duckdb`, `fw/huggingface`, etc.).

### Key Design & Architecture

1. **Zero Upstream Changes**:
   Automatically detects top-level calling frameworks at runtime without requiring upstream libraries or end users to pass custom arguments or wrap filesystems.

2. **Top-Down Call Stack Inspection (`FrameworkDetector`)**:
   Inspects the Python call stack from the root/outermost frame down to the innermost frame using `sys._getframe()`, correctly attributing nested framework hierarchies (e.g., PyTorch Lightning -> PyTorch DataLoader -> gcsfs is tagged as `fw/lightning`).

3. **GCS-Scoped Sync Method Mirroring (`mirror_gcs_sync_methods`)**:
   Attaches `_gcs_sync_wrapper` directly to `GCSFileSystem` instances to capture caller thread telemetry and bridge it into the background asyncio loop thread via `self._sync`.
   Keeps `AbstractFileSystem` and other multi-cloud filesystems (`s3fs`, `adlfs`, `LocalFileSystem`) 100% untouched and isolated.

4. **Multi-Thread, Async, and Process Safety**:
   - `ContextVar` ensures complete thread and task isolation with zero cross-thread context leakage.
   - `os.register_at_fork` cleanly resets telemetry state in child worker processes.
   - Preserves caller framework context across background prefetchers and deferred file close routines.

5. **RFC 9110 Token Sanitization**:
   Sanitizes and truncates framework tokens (`fw/<name>`) to comply strictly with RFC 9110 / HTTP User-Agent grammar and prevent header injection.

6. **Opt-Out Control**:
   Respects `GCSFS_NO_TELEMETRY=1` or `GCSFS_NO_TELEMETRY=true` to completely disable framework detection if desired.

### Tests Added
- Comprehensive unit test suite in `gcsfs/tests/test_telemetry.py` (sanitization, top-to-bottom frame traversal, opt-out, ContextVar isolation, multi-threaded concurrency barriers, fork resets, full method coverage).
- Real upstream framework integration tests in `gcsfs/tests/test_upstream_frameworks_real.py` (PyTorch Lightning trainer, PyTorch Lightning checkpoint save, PyTorch native DataLoader, Pandas `read_csv` and `to_csv`).

### DIFF
--- a/docs/source/telemetry.rst
+++ b/docs/source/telemetry.rst
@@ -1,0 +1,80 @@
+Telemetry & Usage Attribution
+=============================
+
+``gcsfs`` includes lightweight, automatic caller framework detection to identify the upstream data science and machine learning libraries that interact with Google Cloud Storage (GCS).
+
+Purpose
+-------
+
+The Python data ecosystem comprises a rich variety of frameworks—including Pandas, PyTorch, Ray, Dask, Hugging Face, and PyArrow. Telemetry attribution helps the maintainers and Google Cloud Storage understand real-world usage patterns, identify performance bottlenecks for specific workloads, and prioritize compatibility improvements across popular open-source libraries.
+
+What is Collected
+-----------------
+
+When an operation is executed, ``gcsfs`` identifies the name of the top-level calling framework and appends an anonymous brand identifier token to the HTTP ``User-Agent`` header of outgoing requests to Google Cloud Storage.
+
+Example Header
+~~~~~~~~~~~~~~
+
+.. code-block:: text
+
+   User-Agent: gcsfs/2026.3.0 fsspec/2026.3.0 fw/pandas
+
+Recognized Frameworks
+~~~~~~~~~~~~~~~~~~~~~
+
+The detector recognizes major ecosystem libraries, including:
+
+- **Data Processing & Analytics:** ``pandas`` (``fw/pandas``), ``dask`` (``fw/dask``), ``ray`` (``fw/ray``), ``pyspark`` (``fw/spark``), ``pyarrow`` (``fw/pyarrow``), ``duckdb`` (``fw/duckdb``), ``xarray`` (``fw/xarray``), ``fastparquet`` (``fw/fastparquet``).
+- **Machine Learning & Deep Learning:** ``torch`` (``fw/torch``), ``torchdata`` (``fw/torchdata``), ``lightning`` / ``pytorch_lightning`` (``fw/lightning``), ``datasets`` (``fw/datasets``), ``transformers`` (``fw/transformers``), ``orbax`` (``fw/orbax``), ``tensorstore`` (``fw/tensorstore``), ``tensorflow`` (``fw/tensorflow``), ``jax`` (``fw/jax``).
+
+If no recognized framework is present on the call stack (e.g. pure Python scripts or unrecognized utilities), no framework token is added to the header.
+
+Privacy & Security Guarantees
+-----------------------------
+
+Telemetry collection is strictly limited to open-source library names and respects user privacy and security:
+
+- **No User Data or Payloads:** Object contents read or written are never inspected or recorded.
+- **No File Paths or Bucket Names:** Bucket names, object keys, folder hierarchies, and URIs are never collected.
+- **No User Code or Logic:** Variable names, function arguments, and script logic are never accessed.
+- **No Credentials or Identifiers:** API keys, service account credentials, IAM identities, project IDs, and IP addresses are never collected.
+- **Header Injection Safety:** All tokens are sanitized strictly against RFC 9110 HTTP token specifications (alphanumeric characters, hyphens, underscores, and dots only).
+
+How It Works
+------------
+
+1. **Zero Configuration:** Upstream libraries do not need any code changes or special configuration. Detection occurs transparently when filesystem methods (e.g. ``fs.ls``, ``fs.cat_file``, ``fs.open``, ``df.to_parquet``, ``torch.load``) are called.
+2. **Thread Boundary Crossing:** In ``fsspec``, synchronous operations dispatch coroutines to a background asyncio event loop thread. ``gcsfs`` captures caller identity on the user's thread and safely propagates it to the background event loop via Python ``contextvars``.
+3. **Performance & Caching:** Detection uses lightweight stack frame inspection (via ``sys._getframe()``) capped at shallow depths. Active ``GCSFile`` instances cache their framework attribution so subsequent chunk downloads and background prefetching operations execute with zero stack-walking overhead (:math:`O(1)`).
+
+Disabling Telemetry
+-------------------
+
+Telemetry collection is enabled by default, but you can opt out at any time by setting an environment variable.
+
+Using Environment Variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Set ``GCSFS_NO_TELEMETRY=true`` in your shell environment:
+
+.. code-block:: bash
+
+   export GCSFS_NO_TELEMETRY=true
+
+In Python Code
+~~~~~~~~~~~~~~
+
+Set the environment variable before importing ``gcsfs`` or executing storage calls:
+
+.. code-block:: python
+
+   import os
+   os.environ["GCSFS_NO_TELEMETRY"] = "true"
+
+   import gcsfs
+
+When disabled:
+
+- All stack frame inspections and detector evaluations are skipped immediately.
+- Outgoing requests contain only standard ``gcsfs/<version>`` and ``fsspec/<version>`` headers.
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -528,15 +528,10 @@ class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
         async def _coro_with_context():
             token = set_telemetry_context(tokens_map)
             try:
-                if inspect.iscoroutinefunction(func):
-                    return await func(*args, **kwargs)
-                elif inspect.iscoroutine(func) or inspect.isawaitable(func):
-                    return await func
-                else:
-                    res = func(*args, **kwargs)
-                    if inspect.iscoroutine(res) or inspect.isawaitable(res):
-                        return await res
-                    return res
+                res = func(*args, **kwargs) if callable(func) else func
+                if inspect.isawaitable(res):
+                    return await res
+                return res
             finally:
                 reset_telemetry_context(token)
 
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -37,7 +37,12 @@ from .concurrency import parallel_tasks_first_completed, split_range
 from .credentials import GoogleCredentials
 from .inventory_report import InventoryReport
 from .retry import errs, retry_request, validate_response
-from .telemetry.context import Dimension, reset_telemetry_context, set_telemetry_context
+from .telemetry.context import (
+    Dimension,
+    get_telemetry_context,
+    reset_telemetry_context,
+    set_telemetry_context,
+)
 from .telemetry.manager import default_usage_tracker, mirror_gcs_sync_methods
 from .zb_hns_utils import DEFAULT_CONCURRENCY, MAX_PREFETCH_SIZE
 
@@ -2441,9 +2446,10 @@ def _start_deferred_close_worker():
 
 
 def _defer_close(file):
-    tokens_map = default_usage_tracker.collect_tokens_map()
-    if getattr(file, "caller_framework", None):
-        tokens_map[Dimension.FRAMEWORK.value] = file.caller_framework
+    tokens_map = get_telemetry_context()
+    fw = getattr(file, "caller_framework", None)
+    if fw:
+        tokens_map[Dimension.FRAMEWORK.value] = fw
 
     def _run():
         token = set_telemetry_context(tokens_map)
@@ -2801,7 +2807,7 @@ class GCSFile(fsspec.spec.AbstractBufferedFile):
 
     async def _async_fetch_range(self, start_offset, total_size, split_factor=1):
         """Async fetcher mapped to the Prefetcher engine for regional buckets."""
-        tokens_map = default_usage_tracker.collect_tokens_map()
+        tokens_map = get_telemetry_context()
         if self.caller_framework:
             tokens_map[Dimension.FRAMEWORK.value] = self.caller_framework
 
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -398,7 +398,6 @@ class GCSFileSystem(DirCacheUpdater, asyn.AsyncFileSystem):
         self.credentials = GoogleCredentials(
             project, access, token, on_google=self.on_google
         )
-        mirror_gcs_sync_methods(self)
 
     @property
     def _location(self):
@@ -2410,6 +2409,9 @@ def _on_loop_thread(loop):
         return asyncio.get_running_loop() is loop
     except RuntimeError:
         return False
+
+
+mirror_gcs_sync_methods(GCSFileSystem)
 
 
 _DEFERRED_CLOSE_THREAD_NAME = "gcsfs-deferred-close"
--- a/gcsfs/core.py
+++ b/gcsfs/core.py
@@ -3,6 +3,7 @@ Google Cloud Storage pythonic interface
 """
 
 import asyncio
+import inspect
 import io
 import json
 import logging
@@ -36,6 +37,12 @@ from .concurrency import parallel_tasks_first_completed, split_range
 from .credentials import GoogleCredentials
 from .inventory_report import InventoryReport
 from .retry import errs, retry_request, validate_response
+from .telemetry.context import (
+    Dimension,
+    reset_telemetry_context,
+    set_telemetry_context,
+)
+from .telemetry.manager import default_usage_tracker, mirror_gcs_sync_methods
 from .zb_hns_utils import DEFAULT_CONCURRENCY, MAX_PREFETCH_SIZE
 
 logger = logging.getLogger("gcsfs")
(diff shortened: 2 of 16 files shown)


gcsfs/extended_gcsfs.py
```diff
@@ -2113,3 +2114,6 @@ async def simple_upload(
         default_finalize = getattr(fs, "finalize_on_close", False)
         finalize_on_close = kwargs.get("finalize_on_close", default_finalize)
         await zb_hns_utils.close_aaow(writer, finalize_on_close=finalize_on_close)
+
+
+mirror_gcs_methods(ExtendedGcsFileSystem)
```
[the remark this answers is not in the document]
Mahalaxmibejugam: Telemetry is implemented only for JSON path in this PR

gcsfs/extended_gcsfs.py
```diff
@@ -31,6 +31,7 @@
 from gcsfs.concurrency import split_range
 from gcsfs.core import GCSFile, GCSFileSystem, _get_prefetcher_and_cache_config
 from gcsfs.retry import DEFAULT_RETRY_CONFIG, get_storage_control_retry_config
+from gcsfs.telemetry.manager import _gcs_sync_wrapper, mirror_gcs_methods
```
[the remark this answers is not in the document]
Mahalaxmibejugam: Telemetry is implemented only for JSON path in this PR

gcsfs/telemetry/manager.py
```diff
@@ -0,0 +1,290 @@
...
+
+def wrap_file_methods():
+    from gcsfs.core import GCSFile
+
+    for name, attr in list(GCSFile.__dict__.items()):
+        if not name.startswith("__") and callable(attr):
+            if not getattr(attr, "_is_telemetry_wrapped", False):
+                wrapped = _file_telemetry_wrapper(attr)
+                wrapped._is_telemetry_wrapped = True
+                setattr(GCSFile, name, wrapped)
```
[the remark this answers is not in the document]
Mahalaxmibejugam: Telemetry is implemented only for JSON path in this PR

gcsfs/telemetry/detectors/framework.py
```diff
@@ -0,0 +1,97 @@
...
+
+import sys
+from typing import Dict, Optional
+
+from gcsfs.telemetry.context import Dimension
+from gcsfs.telemetry.detectors.base import BaseDetector
+from gcsfs.telemetry.sanitizer import sanitize_framework
+
+
+class FrameworkDetector(BaseDetector):
```
Yonghui-Lee, line 13: Maybe we can introduce a framework cache in FrameworkDetector. The key can be caller's Code Object (`frame.f_code`).

gcsfs/telemetry/manager.py
```diff
@@ -0,0 +1,336 @@
...
+                    fsspec.asyn.sync(self.loop, gen.aclose)
+                except Exception:
+                    pass
+
+    wrapper._is_gcs_gen_wrapped = True
+    return wrapper
+
+
+def _setup_file_telemetry(file_obj) -> Optional[Any]:
+    fw = getattr(file_obj, "caller_framework", None)
```
Yonghui-Lee, line 105: If a file handle is opened in vanilla user code and then passed into a recognized framework (e.g. `pd.read_parquet(file_handle)`, we will return "" instead of the framework.

docs/source/telemetry.rst
```diff
@@ -0,0 +1,80 @@
...
+-----------------
+
+When an operation is executed, ``gcsfs`` identifies the name of the top-level calling framework and appends an anonymous brand identifier token to the HTTP ``User-Agent`` header of outgoing requests to Google Cloud Storage.
+
+Example Header
+~~~~~~~~~~~~~~
+
+.. code-block:: text
+
+   User-Agent: gcsfs/2026.3.0 fsspec/2026.3.0 fw/pandas
```
Yonghui-Lee: _build_user_agent generates
```
User-Agent: python-gcsfs/2026.3.0 [cache_type/...] fw/pandas
```

gcsfs/telemetry/detectors/base.py
```diff
@@ -0,0 +1,35 @@
...
+
+        Returns
+        -------
+        Optional[str]: Formatted token string (e.g. 'fw/pandas') or None if not detected.
+        """
+        raise NotImplementedError
+
+    def is_enabled(self) -> bool:
+        """Check if telemetry opt-out environment variable is active."""
+        return os.environ.get("GCSFS_NO_TELEMETRY", "").lower() not in (
```
Yonghui-Lee: We can also support `DO_NOT_TRACK` env standard which is used in many libraries like huggingface_hub, vLLM.

gcsfs/core.py
```diff
@@ -2358,6 +2391,7 @@ def _get_prefetcher_and_cache_config(cache_type, kwargs):
     return cache_type, use_prefetch_reader, cache_source
 
 
+mirror_gcs_methods(GCSFileSystem)
```
Yonghui-Lee, line 2394: I think it's better to bind sync wrappers per-instance like fsspec instead of at class definition time.

- Broken Subclass Polymorphism: Class-level wrappers capture the parent's _ls in a closure. Subclasses overriding async def _ls are silently ignored when calling fs.ls().
- Broken Test Mocking: monkeypatch.setattr(GCSFileSystem, "_cat_file", ...) no longer intercepts fs.cat_file().
- The ~90 µs saving is negligible: GCSFileSystem initialization time is not on a hot path.

docs/source/telemetry.rst
Mahalaxmibejugam: Thanks for catching this, done.

gcsfs/telemetry/manager.py
Mahalaxmibejugam: Yes this is deliberately not included in this PR. This PR is getting more complicated and big by adding this change,  it will be taken up in future.

gcsfs/telemetry/detectors/base.py
Mahalaxmibejugam: Done

gcsfs/extended_gcsfs.py
```diff
@@ -284,7 +285,7 @@ async def _lookup_bucket_type(self, bucket):
         self._storage_layout_cache[bucket] = bucket_type
         return self._storage_layout_cache[bucket]
 
-    _sync_lookup_bucket_type = asyn.sync_wrapper(_lookup_bucket_type)
+    _sync_lookup_bucket_type = _gcs_sync_wrapper(_lookup_bucket_type)
```
zhixiangli, line 288: Do we need to apply the `_gcs_sync_wrapper` for `make_bucket_requester_pays`, `mkdir`, `rmdir`, `getxattr`, `setxattrs`, `merge`, `mv`, `mv_file`, `rm`, `sync_process_limits_to_offset_and_length`, `makedirs`, etc., to avoid bypassing telemetry?

gcsfs/telemetry/manager.py
```diff
@@ -0,0 +1,336 @@
...
+        return _file_async_gen_wrapper(func)
+    elif inspect.isgeneratorfunction(func):
+        return _file_sync_gen_wrapper(func)
+    elif inspect.iscoroutinefunction(func):
+        return _file_async_wrapper(func)
+    else:
+        return _file_sync_wrapper(func)
+
+
+def wrap_file_methods():
```
zhixiangli, line 196: Should we make it subclass-aware? Otherwise, the Zonal/gRPC path will have no telemetry.

gcsfs/telemetry/manager.py
```diff
@@ -0,0 +1,336 @@
...
+        if token is not None:
+            from gcsfs.telemetry.context import reset_telemetry_context
+
+            reset_telemetry_context(token)
+
+
+def _file_async_gen_wrapper(func):
+    @functools.wraps(func)
+    async def wrapper(self, *args, **kwargs):
+        gen = func(self, *args, **kwargs)
```
zhixiangli, line 139: Ensure `await gen.aclose()`?

gcsfs/telemetry/manager.py
Mahalaxmibejugam: We are only targeting the JSON path for GCSFile in this telemetry, so the zonal file is deliberately not included.

gcsfs/telemetry/manager.py
Mahalaxmibejugam: Done

gcsfs/telemetry/detectors/framework.py
Mahalaxmibejugam: I have investigated f_code caching but it breaks when frameworks compose (e.g., if Dask calls Pandas, caching the innermost Pandas f_code would later incorrectly attribute direct Pandas calls to Dask).

Because we already use ContextVar to propagate the telemetry token, the stack is only walked exactly once per high-level filesystem operation anyway, making the overhead negligible.

gcsfs/telemetry/sanitizer.py
```diff
@@ -0,0 +1,63 @@
+"""RFC 9110 compliant token sanitization helper."""
```
raj-prince, line 1: a file for sanitization with long doc seems overkill to me? Can we use a common utility or helper file? Also, I feel, we can implement a single method which can work for both framework and token usecase.

gcsfs/telemetry/detectors/base.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        Execute detection logic and return an RFC 9110 compliant telemetry token.
+
+        Returns
+        -------
+        Optional[str]: Formatted token string (e.g. 'fw/pandas') or None if not detected.
+        """
+        raise NotImplementedError
+
+    def is_enabled(self) -> bool:
```
raj-prince, line 29: Why do we need this as part of every detector?

gcsfs/telemetry/detectors/base.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import Optional
+
+from gcsfs.telemetry.context import Dimension
+
+
+class BaseDetector:
+    """Base interface for extracting telemetry tokens."""
+
+    @property
+    def name(self) -> Dimension | str:
```
raj-prince, line 15: Not fully understand, why name is required at instance level? Why not a class level attribute e.g., name = "fw" for FrameworkDetector? I would want to avoid that also by ensuring detect() method should return the type like, fw/pytorch.

gcsfs/telemetry/detectors/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""Telemetry detectors package."""
```
raj-prince, line 1: IMO, creating a subpackage with couple of detectors doesn't add much value, should we club the detector subpackage as a single file?

gcsfs/telemetry/context.py
```diff
@@ -0,0 +1,57 @@
...
+
+if hasattr(os, "register_at_fork"):
+
+    def _reset_telemetry_in_child():
+        _current_telemetry.set(None)
+
+    os.register_at_fork(after_in_child=_reset_telemetry_in_child)
+
+
+def get_telemetry_context() -> Dict[str, str]:
```
raj-prince, line 31: Not sure, if 5 methods are required, I feel 3 methods should be sufficient - get, set and reset.

gcsfs/telemetry/manager.py
```diff
@@ -0,0 +1,344 @@
...
+    get_telemetry_context,
+    reset_telemetry_context,
+    set_telemetry_context,
+)
+from gcsfs.telemetry.detectors.base import BaseDetector
+from gcsfs.telemetry.detectors.framework import FrameworkDetector
+from gcsfs.telemetry.sanitizer import sanitize_token
+
+
+def _gcs_async_wrapper(func: Callable, obj: Any = None) -> Callable:
```
raj-prince, line 21: Do we really need 8 different wrapper? Can we minimize that - Gemini tells yes, could you please check once?

gcsfs/telemetry/manager.py
```diff
@@ -0,0 +1,344 @@
...
+            or hasattr(target_type, smethod)
+        ):
+            if not getattr(mth, "__doc__", None):
+                mth.__doc__ = getattr(
+                    getattr(AbstractFileSystem, smethod, None), "__doc__", ""
+                )
+            setattr(obj, smethod, mth)
+
+
+class UsageMetricsTracker:
```
raj-prince, line 284: Why do we need dynamic registration and two methods of getting the header token? I feel, we should keep the interface minimal for UsageMetricsTracker - init(), get_tokens().

raj-prince (commented):
"Overall, approach looks good to me. I've left some comments on how we can simplify the implementation, trim down unused abstractions, and fix a couple of edge cases.

Something to think about: reviewing ~2K lines is hard to review thoroughly in one go. Breaking this kind of feature into two parts (e.g., core telemetry components first, then wiring it into GCSFileSystem/GCSFile) usually makes feedback and iteration a lot smoother and faster."

gcsfs/core.py
Mahalaxmibejugam: We fixed this in commit 0c2c0901 by updating _gcs_sync_wrapper and _gcs_async_gen_wrapper to dynamically resolve getattr(self, method_name) at call time instead of capturing the static function reference. Keeping the descriptors at the class level with runtime lookup fixes subclass polymorphism (_ls, _walk) and monkeypatch.setattr(GCSFileSystem, "_cat_file", ...) even on already-instantiated/cached GCSFileSystem fixtures while preserving class-level attributes for MagicMock(spec=GCSFileSystem) and super().

gcsfs/extended_gcsfs.py
Mahalaxmibejugam: All these methods are already covered, mirror_gcs_methods() at the bottom of core.py and extended_gcsfs.py overwrites their asyn.sync_wrapper(...) definitions with _gcs_sync_wrapper(...) automatically because their names match method[1:] (e.g., "_mkdir"[1:] == "mkdir") 

Only sync_process_limits_to_offset_and_length was not wrapped but this method is not used in production code.

gcsfs/telemetry/detectors/__init__.py
Mahalaxmibejugam: We kept gcsfs/telemetry/detectors/ as a subpackage because FrameworkDetector (fw/) is only the first telemetry dimension—upcoming PRs will add other detectors under detectors/ folder, keeping each detector in its own module avoids growing a single monolithic file.

gcsfs/telemetry/detectors/base.py
Mahalaxmibejugam: Updated name from property to a class-level attribute (name = Dimension.FRAMEWORK). We still keep name as the dimension key on the class because UsageMetricsTracker stores tokens in a ContextVar dictionary keyed by dimension ({"fw": "fw/pytorch", "env": "..."}). Keying by detector.name lets collect_tokens_map() check if dim_key not in tokens: in O(1) before calling detect(), avoiding redundant stack walks on inner HTTP calls (including when detect() returned None and cached "", or when GCSFile.caller_framework already populated "fw").

gcsfs/telemetry/detectors/base.py
Mahalaxmibejugam: is_enabled() is defined once on BaseDetector so all current and future detectors inherit the DO_NOT_TRACK / GCSFS_NO_TELEMETRY opt-out check automatically, while allowing individual detectors to override is_enabled() if a detector-specific toggle is needed.

gcsfs/telemetry/sanitizer.py
Mahalaxmibejugam: removed gcsfs/telemetry/sanitizer.py and consolidated sanitize_framework + sanitize_token into a single sanitize_token() helper in gcsfs/telemetry/context.py

gcsfs/telemetry/context.py
Mahalaxmibejugam: Other two methods were added for supporting get and set for single dimension, but consolidated them to 3 functions (get_telemetry_context, set_telemetry_context, and reset_telemetry_context) with optional dimension support.

gcsfs/telemetry/manager.py
Mahalaxmibejugam: Removed register_detector() since detectors are passed to __init__(). Kept get_dimension(dimension) alongside get_tokens() because get_tokens() returns all active dimensions (e.g. ['fw/pandas', 'env/gke']) for the HTTP User-Agent header, whereas GCSFile.__init__ uses get_dimension(Dimension.FRAMEWORK) to extract and cache only the framework token (self.caller_framework) without mixing in other dimensions.

gcsfs/telemetry/manager.py
Mahalaxmibejugam: Originally included separate wrappers across all callable types (sync, async, and generators) to ensure full forward-compatible coverage as methods are wrapped dynamically. Since GCSFile currently only uses sync and coroutine methods for I/O, streamlined its wrappers into a single _file_telemetry_wrapper function. For GCSFileSystem, _gcs_sync_wrapper, _gcs_async_wrapper, and _gcs_async_gen_wrapper remain separate because each serves a distinct execution path (_run_sync cross-thread bridging, native coroutine scoping, and per-iteration yield bridging for _walk).

Issue #1048, opened 2026-09-08
`cat_file` costs three HTTP round-trips per object since 2026.8.0, causing real slowdown on small files


dhrp:
This issue was detected because our production system had a 50% slowdown in fetching zarr files after upgrading to 2026.8

Analysis found that the adaptive prefetch feature, which was introduced in 2026.8 by itself was not the cause; but an increase of the default concurrency. 

## Impact

Every `cat_file()` call that does not pass an explicit `end` now costs **three HTTP round-trips instead of one**: an object GET, an objects-list call, and only then the download. This is exactly the call zarr (and therefore xarray) makes for every chunk it reads, so workloads made of many small objects pay the full price.

With debug logging on, reading a single 8-byte zarr object looks like this in 2026.7.0:

```
GET: https://storage.googleapis.com/download/storage/v1/b/<bucket>/o/<zarr>%2Freference_time%2F0?alt=media, (), {}
```

and like this in 2026.8.0:

```
GET: b/{}/o/{}, ('<bucket>', '<zarr>/reference_time/0'), None
GET: b/{}/o, ('<bucket>',), None
GET: https://storage.googleapis.com/download/storage/v1/b/<bucket>/o/<zarr>%2Freference_time%2F0?alt=media, (), {'Range': 'bytes=0-7'}
```

Measured on a production zarr store (ECMWF ensemble output, chunks of 24 B to 5.4 MB), fresh `GCSFileSystem` per run, 5 runs, medians, `zarr==3.1.5`, `xarray==2025.12.0`:

| workload | 2026.7.0 | 2026.8.0 | requests 2026.7.0 → 2026.8.0 |
|---|---:|---:|---:|
| `cat_file` on 8 B object, ×20 | 1.00 s | 1.80 s (1.8×) | 20 → 60 |
| `open_zarr(consolidated=True)` (49 coordinate objects) | 1.67 s | 2.48 s (1.5×) | 49 → 157 |
| `isel(...).load()` touching 35 chunks of ~450 KB | 0.45 s | 0.81 s (1.8×) | 35 → 112 |
| `isel(...).load()` touching 1 chunk of 5.4 MB + 1 scalar | 0.39 s | 0.60 s (1.6×) | 2 → 7 |
| `cat_file` on a single 5.4 MB chunk, ×5 | 1.78 s | 1.91 s (1.1×) | 5 → 15 |

[...]

PR #1055, opened 2026-09-10
fix: reverting default cat_file concurrency back to 1 (#1051)

raj-prince:
* Cherry picked the - fix: default cat_file concurrency to 1 (#1048)
* Updated change log.


cloudbuild/run_tests.sh
ankitaluthra1, line 131: These tests are not needed for zonal ?

raj-prince: Not sure. Added as part of this PR - https://github.com/fsspec/gcsfs/pull/766. I don't see similar test for zonal. 

But how is this related to this PR?

ankitaluthra1: Placement issue :((

I mean I see new deselect statements added , do we really need this anymore, the comment says tests are deselected because prefetcher is not enabled for zonal file but thats not true anymore, correct ?
 "--deselect=gcsfs/tests/test_core.py::test_cat_file_default_concurrency"
      "--deselect=gcsfs/tests/test_core.py::test_cat_file_explicit_concurrency"
      "--deselect=gcsfs/tests/test_core.py::test_prefetcher_default_concurrency"

raj-prince: Agree, the comment is outdated. I think, we should fix this as a separate PR in the master. 

The issue is not only with the newly added test, but ideally should have cleanup while adding prefetcher support for zonal bucket.

ankitaluthra1 (approved):

PR #1056, opened 2026-09-10
Always show the e2e test summary (counts and failure logs) in github UI

ankitaluthra1:
**Problem**
The e2e Cloud Build check shows only the **first ~65k characters** of the build log on GitHub. With four test suites streaming `pytest -vv -s` output in parallel, the log reaches that limit partway through the test run. The failure tracebacks, pytest's passed/failed/skipped counts and the `check-failure` step never show up.

External PR #1049's [cat file failure](https://github.com/fsspec/gcsfs/pull/1049/checks?check_run_id=102369502178) is an example. The check text was cut off , while the tests were still printing `PASSED` lines. The steps table shows *which* step failed but not *why*. The "Build Log" link needs GCP project access, so external PR authors can't use it.

**Solution**
Keep the test steps nearly silent, and print a size-capped summary as soon as the tests finish, before any verbose output.  The summary is capped at 30,000 characters. Failures beyond that are listed by test id only. The full logs are printed **after** the summary, so they're still in Cloud Logging. The step always exits 0, so `cleanup` still runs.

### DIFF
--- a/cloudbuild/run_tests.sh
+++ b/cloudbuild/run_tests.sh
@@ -31,8 +31,42 @@ RESULTS_LOG="${RESULTS_DIR}/${TEST_SUITE}.log"
 mkdir -p "${RESULTS_DIR}"
 ARGS+=("--junitxml=${RESULTS_DIR}/${TEST_SUITE}.xml")
 
+# While pytest runs, print the tests still in progress every HEARTBEAT_SECS. A hung test is failed by
+# pytest-timeout and a killed pytest leaves its log tail for test-report, but if the whole build times out
+# (or the VM is lost) test-report never runs, so these lines keep a stuck test visible in the step log.
+HEARTBEAT_SECS="${HEARTBEAT_SECS:-300}"
+if ! [[ "${HEARTBEAT_SECS}" =~ ^[0-9]+$ ]] || (( HEARTBEAT_SECS <= 0 )); then
+  HEARTBEAT_SECS=300
+fi
+
+in_progress_tests() {
+  # With -vv, xdist logs "<nodeid>" when a test starts and "[gwN] <OUTCOME> <nodeid>" when it finishes.
+  awk '
+    { sub(/ <- .*/, ""); sub(/[ \t]+$/, "") }
+    /^\[gw[0-9]+\] [A-Z]+ / { sub(/^\[gw[0-9]+\] [A-Z]+ /, ""); delete running[$0]; next }
+    /^[^ ]+\.py::/ { running[$0] = 1 }
+    END { for (t in running) printf "%s%s", (n++ ? ", " : ""), t; if (!n) printf "none" }
+  ' "${RESULTS_LOG}" 2>/dev/null | cut -c1-500
+}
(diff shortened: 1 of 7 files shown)


cloudbuild/run_tests.sh
```diff
@@ -24,14 +24,26 @@ if ! [[ "${PYTEST_XDIST_WORKERS}" =~ ^[0-9]+$ ]] || (( PYTEST_XDIST_WORKERS <= 0
...
 
+# Full pytest output goes to files instead of the build log. GitHub shows only the first ~65k chars
+# of the Cloud Build log, so the test-report step prints a summary from the JUnit XML before these logs.
+RESULTS_DIR="${HOME}/test-results"
+RESULTS_LOG="${RESULTS_DIR}/${TEST_SUITE}.log"
+mkdir -p "${RESULTS_DIR}"
+ARGS+=("--junitxml=${RESULTS_DIR}/${TEST_SUITE}.xml")
+
+run_pytest() {
+  pytest "$@" > "${RESULTS_LOG}" 2>&1
```
zhixiangli: Previously, output was streamed directly into the step log, and now it is written to a file in the VM. If the process hangs or is killed, it may be hard for us to check for errors?

cloudbuild/summarize_test_results.py
```diff
@@ -0,0 +1,155 @@
...
+
+    if not problems:
+        head.append("")
+        head.append("No failures or errors.")
+
+    parts = ["\n".join(head)]
+    used = len(parts[0]) + len(FOOTER) + NOTE_RESERVE
+    listed_only = []
+    for detail, one_line in problems:
+        if not listed_only and used + len(detail) + 2 <= max_chars:
```
zhixiangli: Could we print everything without the max_chars limitation, or at least ensure that 1 detail line appears for each suite?

zhixiangli (approved):

cloudbuild/summarize_test_results.py
ankitaluthra1: The root cause is on GitHub's side: check-run output is capped (Cloud Build cuts it at ~63k characters and keeps the start), so anything printed late in the build never shows up. max_chars is our budget to keep the summary inside that window after ~15k of setup output. The summary puts the most important parts first: failed build steps and per-suite passed/failed/skipped counts always print, then failure details fill the remaining space. The complete logs always follow the summary, and GitHub shows as much of them as fits.

I agree on the per-suite part: details used to fill in suite order, so one suite with many failures could crowd out the others. Space is now shared round-robin, so every suite with a failure gets at least one detail. Added a test for this.

cloudbuild/run_tests.sh
ankitaluthra1: If pytest is killed, the summary still shows the last 60 lines of its log, and a single hung test is failed by pytest-timeout with its stack trace. If the whole build times out, the step now prints the tests still running every 5 minutes, so a stuck test stays visible; normal runs finish in under 3.5 minutes and print nothing extra.

PR #1065, opened 2026-09-14
fix(perf): use N/A for missing benchmark fields to fix BigQuery schema validation

Mahalaxmibejugam:
This PR fixes a bug where missing fields in benchmarks (such as `batch_size` added by the cat_ranges microbenchmarks) were serialized as empty strings instead of `"N/A"`. This caused BigQuery schema validation to fail for `INT64` columns. BigQuery expects missing numerical fields to be marked as `N/A` per the table configuration `nullMarker: "N/A"`.


Mahalaxmibejugam:
>  Does "nullMarkers": ["N/A", ""] help?

We could use a "" NullMarker, but if any fields in future contain a legitimate "", they would be parsed as NULL, making it impossible to distinguish between the two. So handling it at python layer might be better? WDYT?

zhixiangli (approved):

